PR #65, opened 2026-01-13
fix(installer): handle byte array response in Windows PowerShell 5.1

maliktafheem:
## Summary

Fixes the Windows installer failing on Windows PowerShell 5.1 with the error:

## Problem

In Windows PowerShell 5.1 (the default on Windows 10/11),  returns  as a **byte array** () for non-HTML responses, not a string. The current code assumes  is a string and calls  directly on it.

Additionally, without , PowerShell shows a security warning prompt that can confuse users.

## Solution

- Add  flag to  calls to:
  - Avoid security warning prompts
  - Ensure more consistent behavior across PowerShell versions
- Detect if  is a byte array and convert to string using UTF-8 encoding
- Maintain full compatibility with PowerShell 7+ where  is already a string

## Testing

Tested on:
- Windows 11 with Windows PowerShell 5.1 - ✅ Works
- Fresh install via  pattern

## Changes

- : Added type detection and UTF-8 conversion for checksum response

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -42,10 +42,20 @@ $checksumUrl = "$binaryUrl.sha256"
 New-Item -ItemType Directory -Force -Path $installDir | Out-Null
 
 $tmpFile = [System.IO.Path]::GetTempFileName()
-Invoke-WebRequest -Uri $binaryUrl -OutFile $tmpFile
+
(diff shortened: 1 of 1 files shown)


backnotprop:
Thanks @maliktafheem

backnotprop:
@maliktafheem

Thanks for [#65](https://github.com/backnotprop/plannotator/pull/65).

PR #74, opened 2026-01-14
bugfix: Add WSL detection for browser opening utility

0rdep:
Changes packages/server/browser.ts to detect if running on WSL and open the browser using cmd.

### DIFF
--- a/packages/server/browser.ts
+++ b/packages/server/browser.ts
@@ -53,7 +53,7 @@ export async function openBrowser(url: string): Promise<boolean> {
       if (platform === "darwin") {
         await $`open -a ${browser} ${url}`.quiet();
       } else if (platform === "win32" || wsl) {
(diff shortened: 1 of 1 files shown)


backnotprop:
Thanks for adding WSL support!

I pushed a small fix to add an empty title parameter to the Windows `start` command:

```diff
- await $`cmd.exe /c start ${browser} ${url}`.quiet();
+ await $`cmd.exe /c start "" ${browser} ${url}`.quiet();
```

The Windows `start` command interprets the first quoted argument as a window title. Without the empty `""`, browser paths containing spaces (like `C:\Program Files\...`) would be misinterpreted, causing custom browser settings to silently fail.

This is a well-documented Windows quirk - the empty quotes tell `start` that the title is empty, so it correctly treats the next argument as the program to run.

backnotprop:
@0rdep

Thanks for [#74](https://github.com/backnotprop/plannotator/pull/74).

PR #96, opened 2026-01-19
fix: smart detection for agent switching in OpenCode

### LINKED ISSUE #94: Agent switch default of "build" causes OpenCode crash when build agent is disabled
### Problem

When a user has the default `build` agent disabled in OpenCode and approves a plan without explicitly changing the agent switch setting, OpenCode crashes with:

```
TypeError: undefined is not an object (evaluating 'agent.name')
  at createUserMessage (src/session/prompt.ts:831:14)
```

This happens because:
1. Plannotator defaults to `switchTo: 'build'` ([agentSwitch.ts#L30](https://github.com/backnotprop/plannotator/blob/main/packages/ui/utils/agentSwitch.ts#L30))
2. The plugin sends this agent name to OpenCode without validation ([index.ts#L119](https://github.com/backnotprop/plannotator/blob/main/apps/opencode-plugin/index.ts#L119))
3. OpenCode's `Agent.get()` returns `undefined` for non-existent agents
4. OpenCode crashes accessing `.name` on `undefined`

### Why this is problematic

- **The error appears to be an OpenCode bug** - nothing indicates plannotator is involved
- **The setting is buried** in the settings cog modal - easy to miss
- **Users with custom agents** (who disable defaults) hit this immediately
- **Related issue #47** identified this scenario but the implemented fix (custom input) doesn't prevent the crash for users unaware of the setting

### Suggested solutions (in order of preference)

**Option 1: Smart detection (mentioned as future enhancement in #47)**
- Query OpenCode's `/agent` endpoint on UI load
- Only show available agents in dropdown
- Validate saved setting on startup; warn if invalid

**Option 2: Change default to "disabled"**

One-line fix in `agentSwitch.ts`:
```typescript
const DEFAULT_SETTINGS: AgentSwitchSettings = {
  switchTo: 'disabled',  // was 'build'
};
```

Agent switching is a workflow decision that requires user intent - users unaware of the feature gain nothing from it defaulting to "build", while users with custom agent configurations risk crashes. Defaulting to "disabled" makes agent switching opt-in, which is appropriate for a feature that modifies post-approval behavior.

**Option 3: Validate before sending**
- Before calling `ctx.client.session.prompt({ agent: targetAgent })`, verify agent exists via `/agent` endpoint
- Show user-friendly error if invalid

### Environment
- Plannotator: v0.6.6
- OpenCode: v1.1.25

### Related
- Issue #47 (custom agent support) - mentioned auto-detection as planned
- Issue #37 (configurable agent switch) - added "disabled" option

---

*Note: OpenCode should also add a null check in `Agent.get()` or `createUserMessage`, but that's a separate issue for sst/opencode.*

backnotprop:
## Summary

- Queries OpenCode API for available agents instead of using hardcoded list
- Shows warning banner when saved agent (default: "build") doesn't exist in OpenCode
- Dynamic dropdown populated with actual agents from OpenCode

Fixes #94

## Changes

| File | Change |
|------|--------|
| `packages/server/index.ts` | Add `/api/agents` endpoint, `opencodeClient` option |
| `packages/server/review.ts` | Add `/api/agents` endpoint, `opencodeClient` option |
| `apps/opencode-plugin/index.ts` | Pass `ctx.client` to both servers |
| `packages/ui/components/Settings.tsx` | Fetch agents on mount, dynamic dropdown, warning banner |
| `packages/ui/utils/agentSwitch.ts` | Relax type to `string` for dynamic agents |

## Test plan

- [ ] Test with default "build" agent available → no warning
- [ ] Test with "build" agent disabled in OpenCode → warning shown
- [ ] Test dropdown shows actual available agents from OpenCode
- [ ] Test custom agent input still works
- [ ] Test fallback when API fails (hardcoded options shown)

### DIFF
--- a/packages/ui/components/Settings.tsx
+++ b/packages/ui/components/Settings.tsx
@@ -271,6 +271,14 @@ export const Settings: React.FC<SettingsProps> = ({ taterMode, onTaterModeChange
                       {/* Dynamic agents from OpenCode API */}
                       {availableAgents.length > 0 ? (
                         <>
+                          {/* Show invalid saved value as disabled option if not in list */}
+                          {agent.switchTo !== 'custom' &&
+                           agent.switchTo !== 'disabled' &&
+                           !availableAgents.some(a => a.id.toLowerCase() === agent.switchTo.toLowerCase()) && (
+                            <option value={agent.switchTo} disabled>
+                              {agent.switchTo} (not found)
+                            </option>
+                          )}
                           {availableAgents.map((a) => (
                             <option key={a.id} value={a.id}>
                               {a.name}
--- a/packages/ui/hooks/useAgents.ts
+++ b/packages/ui/hooks/useAgents.ts
@@ -3,7 +3,7 @@
  */
 
 import { useState, useEffect, useCallback } from 'react';
-import { getAgentSwitchSettings, getEffectiveAgentName } from '../utils/agentSwitch';
+import { getAgentSwitchSettings } from '../utils/agentSwitch';
 
 export interface Agent {
   id: string;


backnotprop:
### Code review

PR #163, opened 2026-02-18
fix(browser): use open -a only for PLANNOTATOR_BROWSER on macOS

rerorero:
open -a expects a macOS application name (e.g. "Google Chrome"), not a shell command. When BROWSER=open, the previous code invoked open -a open <url>, which exits with Unable to find application named 'open'.

  This also breaks VS Code devcontainers, which set BROWSER to a script path (/path/to/helpers/browser.sh) — as seen in VS Code's source. This was the primary motivation behind PR #153, but the implementation incorrectly routed BROWSER through open -a on macOS.

https://github.com/microsoft/vscode/blob/790026d0e1c167f9b09a977ddedcbb0d193e3658/src/vs/server/node/extensionHostConnection.ts#L63

  PLANNOTATOR_BROWSER retains its macOS-specific behavior (open -a <app-name>). BROWSER is now executed directly as $BROWSER <url> on all platforms.

### DIFF
--- a/packages/server/browser.ts
+++ b/packages/server/browser.ts
@@ -52,8 +52,8 @@ export async function openBrowser(url: string): Promise<boolean> {
       const plannotatorBrowser = process.env.PLANNOTATOR_BROWSER;
       if (plannotatorBrowser && platform === "darwin") {
         await $`open -a ${plannotatorBrowser} ${url}`.quiet();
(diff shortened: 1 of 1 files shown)


backnotprop:
### Code review

backnotprop:
@rerorero

Thanks for [#163](https://github.com/backnotprop/plannotator/pull/163).

PR #165, opened 2026-02-18
fix(server): prevent permission mode override from unauthenticated request

dor-tzur-lmnd:
## Summary

The `/api/approve` endpoint in the plan server accepts a `permissionMode` field from the browser POST body. This value **takes priority** over the trusted `permissionMode` received from the Claude Code hook event via stdin:

```typescript
// Before: browser value overrides trusted hook event value
const effectivePermissionMode = requestedPermissionMode || permissionMode;
```

The resulting value flows directly into Claude Code's `updatedPermissions` with `setMode`, which changes the session's permission level:

```json
{
  "hookSpecificOutput": {
    "decision": {
      "behavior": "allow",
      "updatedPermissions": [{ "type": "setMode", "mode": "<attacker-controlled>", "destination": "session" }]
    }
  }
}
```

Since the local HTTP server has **no authentication**, any process on localhost can POST to `/api/approve` with an arbitrary `permissionMode` value to escalate the Claude Code session's permission level.

## Fix

Remove the `permissionMode` field from the `/api/approve` request body type. The server now exclusively uses the `permissionMode` from the original hook event (stdin), which is the trusted source provided by Claude Code itself.

```typescript
// After: only use the trusted value from the hook event
resolveDecision({ approved: true, feedback, savedPath, agentSwitch, permissionMode });
```

### Files changed
- `packages/server/index.ts` — Remove browser `permissionMode` override in `/api/approve` handler

## Test plan

- [ ] Verify plan approval still works correctly
- [ ] Verify the session permission mode is preserved as-is after plan approval
- [ ] Verify `/api/approve` POST body no longer accepts `permissionMode`

### DIFF
--- a/packages/server/index.ts
+++ b/packages/server/index.ts
@@ -244,7 +244,6 @@ export async function startPlannotatorServer(
             // Check for note integrations and optional feedback
             let feedback: string | undefined;
             let agentSwitch: string | undefined;
-            let requestedPermissionMode: string | undefined;
             let planSaveEnabled = true; // default to enabled for backwards compat
             let planSaveCustomPath: string | undefined;
             try {
@@ -254,7 +253,6 @@ export async function startPlannotatorServer(
                 feedback?: string;
                 agentSwitch?: string;
                 planSave?: { enabled: boolean; customPath?: string };
-                permissionMode?: string;
               };
 
               // Capture feedback if provided (for "approve with notes")
@@ -265,11 +263,6 @@ export async function startPlannotatorServer(
               // Capture agent switch setting for OpenCode
               if (body.agentSwitch) {
                 agentSwitch = body.agentSwitch;
-              }
-
-              // Capture permission mode from client request (Claude Code)
-              if (body.permissionMode) {
-                requestedPermissionMode = body.permissionMode;
               }
 
               // Capture plan save settings
@@ -312,9 +305,8 @@ export async function startPlannotatorServer(
               savedPath = saveFinalSnapshot(slug, "approved", plan, diff, planSaveCustomPath);
             }
 
-            // Use permission mode from client request if provided, otherwise fall back to hook input
-            const effectivePermissionMode = requestedPermissionMode || permissionMode;
-            resolveDecision({ approved: true, feedback, savedPath, agentSwitch, permissionMode: effectivePermissionMode });
+            // Always use the permission mode from the hook event (stdin), never from the browser
+            resolveDecision({ approved: true, feedback, savedPath, agentSwitch, permissionMode });
             return Response.json({ ok: true, savedPath });
           }
 


backnotprop:
### Code review

Found 2 issues:

1. **Permission mode preference feature is silently broken.** The UI (`packages/editor/App.tsx`) reads the user's stored permission mode preference from cookies via `getPermissionModeSettings().mode`, and sends it as `body.permissionMode` on every approve request. The `PermissionModeSetup` onboarding dialog and Settings dropdown both let users configure their preferred post-approval mode (e.g. `acceptEdits`, `bypassPermissions`). After this PR, the server always echoes back the current session mode from stdin -- the user's preference is silently ignored. The entire permission mode preference subsystem still exists and presents working UI, but has no effect. A safer fix would be to validate/allowlist the client-supplied value (e.g. only accept known `PermissionMode` enum values) rather than removing the field entirely.

https://github.com/backnotprop/plannotator/blob/91b4e2edf9734b15221e0e23cbfb33608030263d/packages/editor/App.tsx#L551-L557

https://github.com/backnotprop/plannotator/blob/91b4e2edf9734b15221e0e23cbfb33608030263d/packages/server/index.ts#L305-L308

2. **Dead code left in UI.** The PR only modifies server-side code but does not clean up the client. `App.tsx` still imports `permissionMode` utils (line 34), manages `permissionMode` state (line 362), includes `permissionMode` in the POST body type (line 552), and sends it on approval (line 556). `PermissionModeSetup.tsx` and `packages/ui/utils/permissionMode.ts` still exist. Users will see UI for configuring a preference that is silently discarded.

https://github.com/backnotprop/plannotator/blob/91b4e2edf9734b15221e0e23cbfb33608030263d/packages/editor/App.tsx#L33-L35

https://github.com/backnotprop/plannotator/blob/91b4e2edf9734b15221e0e23cbfb33608030263d/packages/editor/App.tsx#L361-L363

backnotprop:
@dor-tzur-lmnd Quick question on the approach -- the `permissionMode` field sent from the browser isn't just an echo of the current session mode. It's the user's *preferred* post-approval mode, configured via the Settings UI / onboarding dialog and stored in a cookie. The feature lets users say "after I approve this plan, switch my session to `acceptEdits`" (so Claude can execute the plan without prompting for every file edit).

By always using the stdin value, the hook response just echoes back the mode that was already active -- the user's preference is silently ignored.

Would you be open to an allowlist approach instead? Something like:

```ts
const VALID_MODES = new Set(["bypassPermissions", "acceptEdits", "default"]);
const effectiveMode = (body.permissionMode && VALID_MODES.has(body.permissionMode))
  ? body.permissionMode
  : permissionMode;
```

That would close the injection vector (arbitrary strings can't get through) while preserving the permission mode preference feature. Thoughts?

PR #162, opened 2026-02-18
feat: Add auto-close countdown delay to completion overlay

flex-yj-kim:
## Motivation

When the auto-close setting is off, users see a static completion screen after submitting feedback or approving a plan. They have to manually close the tab every time — but many don't even know the auto-close setting exists in the first place.

By surfacing an opt-in checkbox directly on the completion overlay, we guide users to discover the setting at the exact moment it's most relevant. Once enabled, the countdown timer provides clear feedback before the tab closes, and users can fine-tune the delay (immediate / 3s / 5s) in Settings whenever they want.

## Summary

- Extend auto-close from a simple on/off toggle to configurable delay options: Off / Immediately / After 3 seconds / After 5 seconds
- Show a countdown timer on the completion screen when a delay is set
- Add an opt-in checkbox on the completion overlay to enable auto-close (defaults to 3s) when the setting is off, with a hint that the delay is configurable in Settings
- Handle window.close() browser restriction gracefully with a fallback message when the tab wasn't opened by script
- Extract duplicated completion overlay from both plan editor and review editor into a shared CompletionOverlay component with useAutoClose hook
- Fix build:opencode failure caused by missing @plannotator/server workspace dependency

## Screenshots

<table>
<tr>
<td>image</td>
<td>image</td>
</tr>
</table>

<table>
<tr>
<td>image</td>
<td>image</td>
</tr>

<tr>
<td>image</td>
<td>image</td>
</tr>
</table>

PR #169, opened 2026-02-21
fix(opencode-plugin): move bundled workspace dep to devDependencies

rcdailey:
## Problem

`@plannotator/opencode@0.8.3` is broken for all consumers. Installing it via npm/bun fails with `BunInstallFailedError` because the published package.json contains a `workspace:*` dependency on `@plannotator/server`, which is unresolvable outside the monorepo.

The regression was introduced in #162 (commit 3a9a38e), which added `"@plannotator/server": "workspace:*"` to `dependencies`. Prior versions (0.8.2 and earlier) did not have this dependency and installed correctly.

## Fix

Move `@plannotator/server` from `dependencies` to `devDependencies`. This is safe because the build script already bundles it into `dist/index.js` (only `@opencode-ai/plugin` is externalized via `--external`), so consumers never need to resolve it at install time.

Verified the full build (`bun run build`) still succeeds with this change.

### DIFF
--- a/apps/opencode-plugin/package.json
+++ b/apps/opencode-plugin/package.json
@@ -35,7 +35,9 @@
     "prepublishOnly": "bun run build"
   },
   "dependencies": {
-    "@opencode-ai/plugin": "^1.1.10",
+    "@opencode-ai/plugin": "^1.1.10"
+  },
+  "devDependencies": {
     "@plannotator/server": "workspace:*"
   },
   "peerDependencies": {


backnotprop:
Thanks and sorry about that @rcdailey 

Patch is in `0.8.5`

rcdailey:
No problem, glad I could help. And thank you for the quick turnaround on the merge.

Unrelated: I noticed your release workflow only builds binaries and creates GitHub releases but doesn't publish to npm. Is that intentional? Happy to contribute a workflow for automated npm publishing if that'd be useful.

backnotprop:
@rcdailey it's not intentional... just lazy with npm auth stuff. Lmk your thoughts and what you would need.

PR #184, opened 2026-02-26
feat: open linked .md files in read-only tabs with TOC navigation

dgrissen2:
## Summary

When a plan references local `.md` files (e.g., `[see architecture](./docs/ARCHITECTURE.md)`), clicking the link now opens the file in a new browser tab using the full Plannotator viewer in **read-only mode** with the existing TOC sidebar for navigation.

Previously, `.md` links were treated as plain external links — the browser would download or 404 on them.

### `packages/server/index.ts` — new `/api/doc` endpoint
- Accepts `?path=<filename>` query parameter
- **4-strategy path resolution** (in order):
  1. Absolute path
  2. Relative to `process.cwd()` (project root)
  3. Bare filename search — globs `**/<filename>` across the project
- Returns **400 with match list** if a bare filename matches multiple files (ambiguous)
- Returns **404** if not found, **403** if path escapes project root (path traversal protection)

### `packages/ui/components/Viewer.tsx` — `.md` link detection + `isReadOnly` prop
- `InlineMarkdown` now detects local `.md`/`.mdx` links and rewrites their `href` to `/?doc=<path>&readonly=true` with `target="_blank"`, adding a small external-link icon
- New `isReadOnly` prop: when `true`, skips initializing `web-highlighter` (prevents phantom highlights on text selection) and suppresses both annotation toolbars

### `packages/editor/App.tsx` — read-only mode
- Parses `?doc=` URL param on load and fetches from `/api/doc` instead of `/api/plan`
- On error (404/400/403/500): shows a clear error view with the message — **no silent fallback** to the main plan
- When `isReadOnly`:
  - Shows a read-only banner with the resolved filepath
  - Hides: Approve/Deny/Feedback buttons, ModeSwitcher, Settings, AnnotationPanel, panel toggle button
  - Opens the sidebar to the TOC tab automatically
  - Passes `isReadOnly` through to `Viewer`

## Test plan

- [ ] Create a plan with a link like `[see docs](./README.md)` — verify it renders with an external-link icon and `/?doc=README.md&readonly=true` href
- [ ] Click the link — new tab opens with the README rendered in full Plannotator viewer
- [ ] Verify read-only banner shows the resolved filepath
- [ ] Verify no annotation toolbars appear, no Approve/Deny buttons, no ModeSwitcher
- [ ] Verify TOC sidebar opens automatically and heading navigation works
- [ ] Link to a non-existent file (`missing.md`) — verify error view with "File not found" message, not a fallback to the main plan
- [ ] Create two files with the same name in different directories, link by bare name — verify error view lists both matches
- [ ] `bun run build:hook` passes

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -40,7 +40,7 @@ plannotator/
 │   │   │   ├── plan-diff/        # PlanDiffBadge, PlanDiffViewer, clean/raw diff views
 │   │   │   └── sidebar/          # SidebarContainer, SidebarTabs, VersionBrowser
 │   │   ├── utils/                # parser.ts, sharing.ts, storage.ts, planSave.ts, agentSwitch.ts, planDiffEngine.ts
-│   │   ├── hooks/                # useSharing.ts, usePlanDiff.ts, useSidebar.ts
+│   │   ├── hooks/                # useSharing.ts, usePlanDiff.ts, useSidebar.ts, useLinkedDoc.ts
 │   │   └── types.ts
 │   ├── editor/                   # Plan review App.tsx
 │   └── review-editor/            # Code review UI
@@ -151,6 +151,7 @@ Send Annotations → feedback sent to agent session
 | `/api/upload`         | POST   | Upload image, returns `{ path, originalName }` |
 | `/api/obsidian/vaults`| GET    | Detect available Obsidian vaults           |
 | `/api/plan/vscode-diff` | POST   | Open diff in VS Code (body: baseVersion)   |
+| `/api/doc`              | GET    | Serve linked .md/.mdx file (`?path=<path>`) |
 
 ### Review Server (`packages/server/review.ts`)
 
--- a/packages/editor/App.tsx
+++ b/packages/editor/App.tsx
@@ -361,6 +361,9 @@ const App: React.FC = () => {
   const [origin, setOrigin] = useState<'claude-code' | 'opencode' | 'pi' | null>(null);
   const [globalAttachments, setGlobalAttachments] = useState<ImageAttachment[]>([]);
   const [annotateMode, setAnnotateMode] = useState(false);
+  const [isReadOnly, setIsReadOnly] = useState(false);
+  const [docFilepath, setDocFilepath] = useState<string | null>(null);
+  const [docError, setDocError] = useState<string | null>(null);
   const [isLoading, setIsLoading] = useState(true);
   const [isSubmitting, setIsSubmitting] = useState(false);
   const [submitted, setSubmitted] = useState<'approved' | 'denied' | null>(null);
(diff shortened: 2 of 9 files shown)


backnotprop:
Sounds awesome will check it out today

backnotprop:
Hey — thanks for this. The `/api/doc` endpoint and the link detection in `InlineMarkdown` were clean and well thought out.

It was such a good idea that I wanted to empower it a bit more. We ended up building on top of your work to let users annotate linked docs the same way they annotate the plan, with all that feedback aggregated into the deny/approve payload sent to Claude. Instead of opening a new tab, linked docs now swap in-place within the same viewer so annotations persist across navigation.

Pushed a commit to your branch with those changes. The main additions:
- `useLinkedDoc` hook for managing the view swap and annotation caching
- `exportLinkedDocAnnotations()` for including linked doc feedback in the payload
- Some UI indicators (border, badges, TOC context) so it's clear when you're viewing a linked file
- A couple of fixes (glob excluding `node_modules`, Tailwind cascade issue with `border-color`)

Your endpoint and link detection carried over intact — just added `node_modules`/`.git` exclusion to the glob. Thanks again for getting this started.

backnotprop:
### Code review

Found 3 issues:

1. `annotationsOutput` useMemo depends on `linkedDocHook` (the whole hook return object), which is a new reference every render — the memo recomputes on every render, defeating memoization entirely. Before this PR, `annotationsOutput` was correctly memoized against `[blocks, annotations, globalAttachments]`. Fix by extracting only needed stable values (e.g., `linkedDocHook.getDocAnnotations`) or memoizing the hook's return object. (performance regression — `useMemo` effectively a no-op)

https://github.com/backnotprop/plannotator/blob/d4e36fb0e26d3add8d12a9c8c0f41cf4e29d9a08/packages/editor/App.tsx#L800-L802

2. `getDocAnnotations()` returns only what is in `docCache`, which is written when navigating away via `back()` or opening another doc. If the currently-active linked doc has annotations and the user triggers submit (e.g., via the Cmd+Enter shortcut, which is not guarded by `linkedDocHook.isActive`), those annotations are silently dropped from the feedback payload. (data loss — active linked doc annotations missing from output)

https://github.com/backnotprop/plannotator/blob/d4e36fb0e26d3add8d12a9c8c0f41cf4e29d9a08/packages/ui/hooks/useLinkedDoc.ts#L205-L208

3. `exportLinkedDocAnnotations` sorts annotations by `a.blockId.localeCompare(b.blockId)`, but block IDs have the format `block-N`. Lexicographic sort puts `block-10` before `block-9`, producing wrong annotation order for documents with 10+ blocks. The existing `exportAnnotations` correctly sorts by `blocks.findIndex()`. (bug — wrong output order for docs with 10+ blocks)

https://github.com/backnotprop/plannotator/blob/d4e36fb0e26d3add8d12a9c8c0f41cf4e29d9a08/packages/ui/utils/parser.ts#L344-L348

backnotprop:
### Code review

Found 1 issue:

1. `handleApprove` gate only checks plan-level annotations, silently dropping linked-doc-only feedback. When a user annotates only linked documents (zero plan annotations) and approves, `annotations.length > 0 || globalAttachments.length > 0` is false so `body.feedback` is never set — even though `annotationsOutput` includes the linked doc annotations via `getDocAnnotations()`.

https://github.com/backnotprop/plannotator/blob/e5ca4566bff1abdde7459e233553887323d95355/packages/editor/App.tsx#L631-L634

backnotprop:
@dgrissen2

Thanks for [#184](https://github.com/backnotprop/plannotator/pull/184).

PR #195, opened 2026-02-28
fix: respect XDG Base Directory Specification of ~/.local/bin

idletea:
The installer is currently broken for people with a defined `XDG_DATA_HOME` because that env var isn't the place where user-specific executables are meant to go. They go (unconditionally, there is no env var to override it) in `~/.local/bin`

So hardcoding `~/.local/bin` is respecting the XDG spec.

Spec reference for that [here](https://specifications.freedesktop.org/basedir/latest/#variables)

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -2,7 +2,7 @@
 set -e
 
 REPO="backnotprop/plannotator"
-INSTALL_DIR="${XDG_DATA_HOME:-$HOME/.local}/bin"
+INSTALL_DIR="$HOME/.local/bin"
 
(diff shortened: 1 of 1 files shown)


backnotprop:
Merged and deployed — the updated install.sh is now live at https://plannotator.ai/install.sh (S3 bucket updated manually). Verified via curl that `INSTALL_DIR` is `$HOME/.local/bin` as expected.

backnotprop:
@idletea

Thanks for [#195](https://github.com/backnotprop/plannotator/pull/195).

PR #197, opened 2026-03-01
fix(opencode): notify URL in remote mode and add 10-min decision timeout

### LINKED ISSUE #192: Remote Claude Code sessions: no way to open plan review UI without a local browser
## Problem

When using Plannotator via a **remote Claude Code instance** (SSH, cloud VM, headless server), the hook starts a local server and either:
- Tries to open a browser that doesn't exist, or
- Correctly detects the remote session (`SSH_TTY`/`SSH_CONNECTION`/`PLANNOTATOR_REMOTE=1`) and skips the browser — but then **there's no way to actually view the plan review UI** from your local machine.

The server binds to `localhost:<port>` on the remote host, which isn't accessible from the user's local browser. The user is stuck — they can see the URL printed in the terminal, but can't open it.

This is a real blocker for anyone using Claude Code over SSH (which is a very common setup — remote dev servers, cloud instances, codespaces, etc.).

## Expected Behavior

When running on a remote instance, Plannotator should give the user a **shareable URL** they can open in their local browser — no tunneling, no port forwarding required.

### Option A: Auto-generate share URL in remote mode (Recommended)

When a remote session is detected, automatically generate the compressed share URL (or short paste URL from #188) and print it to the terminal. The user copies the link, opens it on their local machine via `share.plannotator.ai`, done.

```
🔗 Remote session detected — open this link on your local machine:
   https://share.plannotator.ai/p/aBcDeFgH

   (or full URL: https://share.plannotator.ai/#eJy0Vk1v4z...)

   Waiting for your decision...
```

This is the cleanest UX — zero setup, works everywhere. The share portal is already a static SPA that works independently of the server.

### Option B: `--url-only` / `--no-open` CLI flag

Add a flag that skips the browser open and just prints the share URL:

```bash
# In .claude/hooks/ExitPlanMode.sh
plannotator --url-only
```

This gives users explicit control regardless of remote detection.

### Option C: Clipboard-friendly output

When remote, copy the share URL to the remote clipboard (if available via `xclip`/`xsel`/OSC 52 terminal escape) and print it. Many modern terminals (iTerm2, WezTerm, kitty) support OSC 52 for remote-to-local clipboard sync.

### Option D: QR code in terminal

Print a QR code in the terminal that the user can scan with their phone to open the review. Libraries like `qrcode-terminal` make this trivial. Great for quick mobile review.

## Implementation Notes

- Remote detection already works well in `packages/server/remote.ts` (`PLANNOTATOR_REMOTE`, `SSH_TTY`, `SSH_CONNECTION`)
- The `onReady` callback in all three modes (plan/review/annotate) already receives `isRemote` — just needs to generate and print a share URL when `isRemote === true`
- With PR #188 (short URLs), this becomes even better — a 50-char URL instead of a 40KB hash URL
- The `handleServerReady`/`handleReviewServerReady`/`handleAnnotateServerReady` functions are the exact place to add this logic
- Should work for all three modes: plan review, code review (`plannotator review`), and annotate (`plannotator annotate`)

## Workaround (current)

Users can manually set up SSH port forwarding:
```bash
ssh -L 19432:localhost:19432 remote-host
```
But this is fragile (port changes), requires planning ahead, and is a terrible UX.

---

I use Claude Code on remote instances daily and this is the #1 thing preventing me from using Plannotator in that workflow. Happy to create a PR implementing any of these approaches — Option A seems like the best default behavior.

codeg-old:
## Problem

In `PLANNOTATOR_REMOTE=1` environments (SSH, devcontainer, OpenCode web UI accessed via Tailscale or a reverse proxy), two compounding bugs make `submit_plan` effectively unusable:

### Bug 1 — No URL notification (closes #192)

When `isRemote=true`, `handleServerReady()` correctly skips opening a browser — but does nothing else. The user has **no way to know** where to navigate to reach the Plannotator UI. The server silently waits forever.

For OpenCode users accessing the web UI remotely, `ctx.client.app.log()` provides the right channel to surface this information directly in the chat UI.

### Bug 2 — No timeout on `waitForDecision()`

If the user never visits the URL (because they don't know it), `waitForDecision()` blocks indefinitely and **holds the port permanently**. Any subsequent `submit_plan` call fails immediately with `EADDRINUSE` (after 5 retries × 500ms). From the user's perspective, `submit_plan` appears to silently not execute.

## Fix

**`apps/opencode-plugin/index.ts`**

1. **URL notification**: When `isRemote=true`, emit a log message via `ctx.client.app.log()` showing the Plannotator URL (both `localhost:<port>` and `hostname:<port>` for remote access).

2. **10-minute timeout**: Wrap `waitForDecision()` in `Promise.race()` with a 10-minute timeout that auto-denies and releases the port, breaking the permanent-occupation cycle.

```diff
+import os from "node:os";

  onReady: (url, isRemote, port) => {
    handleServerReady(url, isRemote, port);
+   if (isRemote) {
+     const hostname = os.hostname();
+     ctx.client.app.log({
+       level: "info",
+       message: `[Plannotator] UI ready — open in browser: http://localhost:${port} (or http://${hostname}:${port} if remote)`,
+     }).catch(() => {});
+   }
  },

- const result = await server.waitForDecision();
+ const PLANNOTATOR_TIMEOUT_MS = 10 * 60 * 1000; // 10min timeout
+ const result = await Promise.race([
+   server.waitForDecision(),
+   new Promise<{ approved: boolean; feedback?: string }>((resolve) =>
+     setTimeout(
+       () => resolve({ approved: false, feedback: "[Plannotator] No response within 10 minutes. Port released automatically. Please call submit_plan again." }),
+       PLANNOTATOR_TIMEOUT_MS
+     )
+   ),
+ ]);
```

## Notes

- The timeout auto-deny returns a clear message so the agent knows to call `submit_plan` again, rather than treating it as a real rejection.
- `os.hostname()` uses the Node.js built-in — no new dependencies.
- Only the `submit_plan` execute path is changed. The `/plannotator-review` and `/plannotator-annotate` command handlers are not affected.

---

> **AI authorship disclosure**: This fix was authored by Claude (Anthropic) and reviewed by [@codeg-dev](https://github.com/codeg-dev) before submission.

### DIFF
--- a/apps/opencode-plugin/index.ts
+++ b/apps/opencode-plugin/index.ts
@@ -26,6 +26,7 @@ import {
   handleAnnotateServerReady,
 } from "@plannotator/server/annotate";
 import { getGitContext, runGitDiff } from "@plannotator/server/git";
+import os from "node:os";
 
 // @ts-ignore - Bun import attribute for text
 import indexHtml from "./plannotator.html" with { type: "text" };
@@ -313,10 +314,26 @@ Do NOT proceed with implementation until your plan is approved.
             opencodeClient: ctx.client,
             onReady: (url, isRemote, port) => {
               handleServerReady(url, isRemote, port);
+              if (isRemote) {
+                const hostname = os.hostname();
+                ctx.client.app.log({
+                  level: "info",
+                  message: `[Plannotator] UI ready — open in browser: http://localhost:${port} (or http://${hostname}:${port} if remote)`,
+                }).catch(() => {});
+              }
             },
           });
 
-          const result = await server.waitForDecision();
+          const PLANNOTATOR_TIMEOUT_MS = 10 * 60 * 1000; // 10min timeout
+          const result = await Promise.race([
+            server.waitForDecision(),
+            new Promise<{ approved: boolean; feedback?: string }>((resolve) =>
+              setTimeout(
+                () => resolve({ approved: false, feedback: "[Plannotator] No response within 10 minutes. Port released automatically. Please call submit_plan again." }),
+                PLANNOTATOR_TIMEOUT_MS
+              )
+            ),
+          ]);
           await Bun.sleep(1500);
           server.stop();
 
--- a/apps/opencode-plugin/index.ts
+++ b/apps/opencode-plugin/index.ts
@@ -26,7 +26,7 @@ import {
   handleAnnotateServerReady,
 } from "@plannotator/server/annotate";
 import { getGitContext, runGitDiff } from "@plannotator/server/git";
-import os from "node:os";
+import { writeRemoteShareLink } from "@plannotator/server/share-url";
 
 // @ts-ignore - Bun import attribute for text
 import indexHtml from "./plannotator.html" with { type: "text" };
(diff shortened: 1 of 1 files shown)


backnotprop:
Awesome, taking a look

backnotprop:
@codeg-dev

Thanks for [#197](https://github.com/backnotprop/plannotator/pull/197).

PR #177, opened 2026-02-23
feat: Plan Diff marketing dialog, Pi origin, and docs

backnotprop:
## Summary

- **Plan Diff Marketing dialog**: First-run informational dialog announcing Plan Diff Mode — banner image placeholder, two view modes described, version history, coming-soon cross-plan comparison, per-origin video demo links, community feedback callout
- **Pi as first-class origin**: Added `'pi'` to the origin type union across `App.tsx`, `Settings.tsx`, `UpdateBanner.tsx`, and `useAgents.ts` with display name ("Pi") and violet badge color
- **Per-origin video URLs**: `PlanDiffMarketing` resolves a different demo video URL depending on whether the user is on Claude Code, OpenCode, or Pi
- **Blog post**: `plan-diff-see-what-changed.md` — short, SEO-friendly post with video placeholder, community issue references (#138, #111)
- **Docs updates**: Brief Plan Diff mentions added to root README, all plugin READMEs, quickstart, plan-review command doc, and UI settings reference

## Test plan

- [ ] Load from Claude Code — orange badge, PlanDiffMarketing dialog shows on first run with Claude Code video URL
- [ ] Load from OpenCode — zinc badge, video URL points to OpenCode demo
- [ ] Load from Pi — violet badge, "Pi" display name, Pi-specific video URL
- [ ] Dismiss dialog → cookie set → dialog does not reappear on reload
- [ ] Dialog chains correctly: PermissionModeSetup → UIFeaturesSetup → PlanDiffMarketing
- [ ] Blog post renders correctly on marketing site
- [ ] README feature table includes Plan Diff row

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #199, opened 2026-03-01
fix: Fix diff view not clearing on sidebar tab switch

flex-yj-kim:
## Summary
- Clear diff view state when switching from Versions tab to Contents tab in the sidebar
- Dismiss diff view overlay on Escape key press
- Add dev mock API for version history to support local development

### DIFF
--- a/apps/hook/dev-mock-api.ts
+++ b/apps/hook/dev-mock-api.ts
@@ -1,0 +1,257 @@
+/**
+ * Vite plugin that mocks plannotator API endpoints for local development.
+ * Provides plan data with version history so the Versions tab works in dev mode.
+ */
+import type { Plugin } from 'vite';
+
+// Version 1: earlier draft (shorter, missing sections)
+const PLAN_V1 = `# Implementation Plan: Real-time Collaboration
+
+## Overview
+Add real-time collaboration features to the editor using WebSocket connections.
+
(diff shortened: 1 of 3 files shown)


backnotprop:
### Code review

Found 1 issue:

1. `isPlanDiffActive` is read inside the tab-switch `useEffect` but missing from its dependency array, creating a stale closure. The Escape key effect directly below correctly includes `[isPlanDiffActive]` in its deps. This is the same class of bug that was caught in PR #177.

https://github.com/backnotprop/plannotator/blob/69a672868e0bfc2ce1a6f97fe490fd21f3be19a9/packages/editor/App.tsx#L409-L414

backnotprop:
**Correction to my earlier review — disregard the previous comment.**

After pulling the branch locally and tracing the data flow more carefully, the missing `isPlanDiffActive` dependency is **intentional, not a bug**.

The effect's purpose is to clear the diff view when `activeTab` *transitions* to `toc`. The `isPlanDiffActive` check inside the body is just a guard to skip a redundant state update. If `isPlanDiffActive` were added to the deps array, it would create a worse bug: activating diff view while already on the TOC tab would immediately get cleared by the effect re-firing.

The closure captures `isPlanDiffActive` from the render where `activeTab` changed, which is always fresh at that point. No issue here.

### Code review

PR #200, opened 2026-03-02
fix: Fix review showing empty files + propagate diff errors to UI

flex-yj-kim:
## Summary
- Force standard `a/`/`b/` diff prefixes to fix empty file list when `diff.mnemonicPrefix` is enabled in git config
- Propagate git diff errors to the review UI instead of silently showing "No changes"

### Mnemonic prefix fix

When `diff.mnemonicPrefix = true` is set in git config, `git diff` replaces the standard `a/`/`b/` path prefixes with context-dependent letters (`c/` for commit, `w/` for worktree, `i/` for index). Both the internal `parseDiffToFiles` regex and the `@pierre/diffs` library assume `a/`/`b/`, so non-standard prefixes cause silent parse failures — the UI shows "No changes" even though uncommitted changes exist.

Fixed by adding `--src-prefix=a/ --dst-prefix=b/` to all `git diff` invocations in `packages/server/git.ts`.

### Error propagation

Previously, git diff errors (e.g., no commits yet, invalid ref) were silently caught and returned as empty patches. The UI could not distinguish between "no changes" and "git command failed".

Added `error` field to `DiffResult` → `ReviewServerOptions` → `/api/diff` response → `diffError` state in the review UI. When an error occurs, the UI now shows a red warning icon with the error message instead of the misleading "No changes" text.

image

### DIFF
--- a/apps/pi-extension/server.ts
+++ b/apps/pi-extension/server.ts
@@ -389,15 +389,15 @@ export function getGitContext(): GitContext {
 export function runGitDiff(diffType: DiffType, defaultBranch = "main"): { patch: string; label: string } {
   switch (diffType) {
     case "uncommitted":
-      return { patch: git("diff HEAD"), label: "Uncommitted changes" };
+      return { patch: git("diff HEAD --src-prefix=a/ --dst-prefix=b/"), label: "Uncommitted changes" };
     case "staged":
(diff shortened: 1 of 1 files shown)


backnotprop:
### Code review

PR #188, opened 2026-02-26
Add short URL sharing via paste service to fix URL truncation

GunitBindal:
## Summary

Fixes #187 — Share URLs for large plans are 10-40KB+ and get truncated by Slack, WhatsApp, email clients, etc., making the team sharing feature unusable for real-world plans. Also addresses the root cause behind #142 (silent fallback to demo plan on truncated URLs).

This adds an **optional paste-service backend** that stores compressed plan payloads and returns short ~50-char URLs like `share.plannotator.ai/p/aBcDeFgH`.

### What changed

**New: `apps/paste-service/`** — Pluggable paste service with multiple deployment targets
- `POST /api/paste` stores compressed plan data, returns short ID
- `GET /api/paste/:id` retrieves stored data
- Pluggable storage backends: filesystem (self-hosted), Cloudflare KV (hosted), S3
- Configurable TTL, 512KB limit, CORS configured
- Works in both Bun and Cloudflare Worker runtimes

**New: `packages/shared/`** — Shared utilities (compression)
- Extracted `compress`/`decompress` from `packages/ui` to avoid cross-package duplication
- Used by both `@plannotator/server` and `@plannotator/ui`

**Modified: `packages/ui/utils/sharing.ts`**
- Added `createShortShareUrl()` — POSTs compressed data to paste service (5s timeout)
- Added `loadFromPasteId()` — fetches plan by paste ID (10s timeout)
- Exported `toShareableImages` and `parseShareableImages` (previously module-private)
- Imports `compress`/`decompress` from `@plannotator/shared`

**Modified: `packages/ui/hooks/useSharing.ts`**
- Added `shortShareUrl`, `isGeneratingShortUrl`, `shortUrlError` to hook state
- `loadFromHash()` now checks for `/p/<id>` path pattern first, then falls back to hash
- Short URL generation requires explicit user action (click "Create short link")
- All existing behavior preserved

**Modified: `packages/ui/components/ExportModal.tsx`**
- Share tab shows short URL as primary copy target when available
- Full hash URL always visible as backup
- Loading spinner while short URL generates
- Amber hint when paste service is unavailable
- Copy button tracks which URL was copied (short/full/annotations independently)

**Modified: `packages/server/share-url.ts`**
- Uses `@plannotator/shared/compress` instead of hand-rolled pipeline
- Added `writeRemoteShareLink()` helper for remote session URL display

**Modified: `apps/hook/server/index.ts`**
- Passes `pasteApiUrl` from env to server for short URL support
[...]

PR #203, opened 2026-03-02
Deploy paste service with E2E encryption

backnotprop:
## Summary

- Adds **end-to-end encryption** (AES-256-GCM) to the short URL sharing flow — plans are encrypted in the browser before upload, the paste service stores only ciphertext it cannot read
- The decryption key is generated via Web Crypto API, never leaves the browser, and lives only in the URL fragment (`#key=...`)
- Deploys the Cloudflare Worker to `plannotator-paste.plannotator.workers.dev` with KV namespaces
- Adds CI/CD auto-deploy job for paste service changes
- Updates all docs, README, and UI strings with encryption messaging

## Changes

| File | What |
|------|------|
| `packages/shared/crypto.ts` | New — AES-256-GCM encrypt/decrypt via Web Crypto API |
| `packages/shared/package.json` | Add `./crypto` export |
| `packages/ui/utils/sharing.ts` | Encrypt before POST, decrypt on load, key in fragment |
| `packages/ui/hooks/useSharing.ts` | Parse `#key=` from URL fragment |
| `packages/ui/components/ExportModal.tsx` | Updated 3 user-facing strings |
| `apps/paste-service/core/handler.ts` | Cache-Control: private, no-store |
| `apps/paste-service/wrangler.toml` | Real KV namespace IDs |
| `.github/workflows/deploy.yml` | deploy-paste job |
| README + 4 marketing docs | Encryption/privacy messaging |

## Data flow

```
Create:  Plan → compress → encrypt(AES-256-GCM) → POST ciphertext → /p/{id}#key={key}
Load:    /p/{id}#key={key} → GET ciphertext → decrypt(key) → decompress → Plan

Server sees: opaque ciphertext only. Zero knowledge.
Key: generated in browser, lives only in URL fragment, never sent in HTTP requests.
```

## Remaining manual steps

- [ ] Add `CLOUDFLARE_API_TOKEN` and `CLOUDFLARE_ACCOUNT_ID` as GitHub repo secrets

## Test plan

- [ ] Create a short URL from a plan — URL contains `#key=` fragment
- [ ] Open the short URL in a new tab — plan loads correctly
- [ ] Import a short URL via the import dialog — imports correctly
- [ ] Open a short URL without the `#key=` fragment — fails gracefully
- [ ] Full hash-based URLs still work (no regression)
- [ ] Verify Worker responds: `curl https://plannotator-paste.plannotator.workers.dev/api/paste/nonexistent` → 404

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,9 +48,6 @@ Plannotator lets you privately share plans, annotations, and feedback with colle
 - Fully open source and self-hostable ([see docs](https://plannotator.ai/docs/guides/sharing-and-collaboration/))
 
 > [!NOTE]
-> [share.plannotator.ai](https://share.plannotator.ai) uses a default fallback (demo) plan that is hard-coded into the site. This isn't a leaked plan — the site has no storage layer.
-
-> [!NOTE]
 > Short links are end-to-end encrypted. A single-use AES-256-GCM key is generated in your browser via the [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/generateKey), used to encrypt the plan, and embedded in the URL fragment (`#key=...`). The key never leaves the browser — it is never sent to the paste service or any server. Only someone with the exact URL can decrypt the plan.
 
 ## Install
--- a/packages/ui/hooks/useSharing.ts
+++ b/packages/ui/hooks/useSharing.ts
@@ -133,7 +133,9 @@ export function useSharing(
 
           return true;
         }
-        // Paste fetch failed — fall through to try the hash fragment instead.
+        // Paste fetch failed — short URL path can't fall back to hash parsing
+        // (the hash contains #key=, not plan data).
+        return false;
       }
 
       const payload = await parseShareHash();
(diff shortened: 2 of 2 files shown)


backnotprop:
@GunitBindal I've added e2e encryption for more privacy guarentees. Even I wont be able to see other's plans in the stored KV.

backnotprop:
### Code review

Found 2 issues:

1. `loadFromHash` fallthrough silently feeds `#key=...` to `parseShareHash()` on paste failure — when an encrypted short URL's paste expires or fails to load, the code falls through (line 136) to `parseShareHash()` which reads `window.location.hash.slice(1)` and gets `key=<base64url>` instead of compressed plan data. Decompression always fails silently, so the user sees a blank page with no error message instead of an actionable "link expired" notice. The fallthrough comment "try the hash fragment instead" can never succeed for encrypted short URLs since the hash slot is consumed by the encryption key.

https://github.com/backnotprop/plannotator/blob/0309c54a55a4b532f22ad2be0552cc2db6d0833d/packages/ui/hooks/useSharing.ts#L135-L139

2. `importFromShareUrl` regex lost its end-anchor, reintroducing the false-match bug fixed in PR #188 — the regex changed from `/\/p\/([A-Za-z0-9]{6,16})(?:\?|$)/` to `/\/p\/([A-Za-z0-9]{6,16})(?:#key=([A-Za-z0-9_-]+))?/`, dropping the `(?:\?|$)` anchor. A URL with extra path segments after the paste ID (e.g. `/p/aBcDeFgH/extra`) would now incorrectly match. PR #188 explicitly added the `loadFromHash` pathname regex with `^...$` anchoring to prevent this class of bug, but the `importFromShareUrl` regex was not given the same treatment.

https://github.com/backnotprop/plannotator/blob/0309c54a55a4b532f22ad2be0552cc2db6d0833d/packages/ui/hooks/useSharing.ts#L262-L264

Issue #216, opened 2026-03-04
Cross-block annotation highlights not restored on shared URL or draft restore


backnotprop:
## Problem

When an annotation spans multiple blocks (e.g., text selected across two paragraphs), the highlight mark is not visually restored — though the annotation itself appears in the sidebar panel. This affects both shared URL restore and draft restore.

## Root Cause

`findTextInDOM()` in `Viewer.tsx` searches for the annotation's `originalText` as a literal string in the DOM. For cross-block selections, `originalText` contains `\n\n` characters representing block boundaries. But in the rendered DOM, those boundaries are structural (`<p>` elements), not text node newlines — so the search fails silently.

The web-highlighter library already captures `startMeta`/`endMeta` (DOM path metadata) when annotations are created, and provides `highlighter.fromStore(startMeta, endMeta, text, id)` for exact restoration. However:

1. `toShareable()` strips `startMeta`/`endMeta` from the compact tuple format
2. `fromShareable()` leaves them `undefined`
3. `applySharedAnnotations()` falls back to text search, which fails for cross-block text

## Affected Paths

- Shared URL restore (`useSharing.ts` → `applySharedAnnotations`)
- Draft restore (`useAnnotationDraft.ts` → `applySharedAnnotations`)

## Possible Fixes

1. **Include `startMeta`/`endMeta` in shareable format** — extend the tuple to carry DOM metadata, then use `highlighter.fromStore()` on restore
2. **Fix `findTextInDOM()` to handle cross-block text** — normalize both search text and DOM text to handle structural block boundaries

Option 2 is more robust since it fixes all restore paths without changing the share payload format.

## Key Files

- `packages/ui/components/Viewer.tsx` — `findTextInDOM()`, `applySharedAnnotations()`
- `packages/ui/utils/sharing.ts` — `toShareable()`, `fromShareable()`
- `packages/ui/types.ts` — `Annotation.startMeta` / `endMeta`

backnotprop:
Thanks for the report. I am consolidating the older backlog so related work has one current tracking issue.

This specific request is now tracked in #906 as part of the annotation robustness and review modes work. During the June 13, 2026 older-than-two-month triage, I copied the relevant requirement from this issue into #906 and explained why it belongs with the related work.

I am closing this older issue so the backlog points at the current tracking issue. If there is a newer reproduction or a detail that is missing from #906, please comment there.

PR #217, opened 2026-03-04
Auto-save annotation drafts to survive server crashes

### LINKED ISSUE #212: Support auto save annotation
I got `The original annotation output file (b7xne11zr.output) no longer exists ( continue from error)` in claude .

claude analysis : 

The annotation session was interrupted before commit (the status was killed), so the annotation data was not persisted.

Plannotator works by temporarily storing the annotation data in the browser memory, and only when you click Submit (Submit) will it be sent to the server via /api/feedback and saved. The annotation is missing because the previous service process was terminated.

-------

This is very frustrate as I wrote massive comments and it lost all.

backnotprop:
## Summary

- Adds server-side draft persistence at `~/.plannotator/drafts/{content-hash}.json` so annotations survive server crashes
- Debounced auto-save (500ms) on annotation changes — silent, no UI indicators
- On reload with existing draft, a dialog prompts to Restore or Dismiss
- Drafts auto-deleted on approve/deny/feedback submit
- Works across all three servers (plan, review, annotate)

Closes #212

**Known limitation:** Cross-block annotation highlights don't visually restore (pre-existing issue in `findTextInDOM()`, tracked in #216).

## Test plan

- [ ] Add annotations to a plan, kill the server, restart with same plan content — dialog should appear
- [ ] Click Restore — annotations appear in sidebar panel
- [ ] Click Dismiss — draft deleted, no dialog on next reload
- [ ] Approve/deny a plan — draft file cleaned up
- [ ] Same flow works for code review (`/plannotator-review`)

### DIFF
--- a/packages/editor/App.tsx
+++ b/packages/editor/App.tsx
@@ -535,7 +535,7 @@ const App: React.FC = () => {
     submitted: !!submitted,
   });
 
-  const handleRestoreDraft = () => {
+  const handleRestoreDraft = React.useCallback(() => {
     const { annotations: restored, globalAttachments: restoredGlobal } = restoreDraft();
     if (restored.length > 0) {
       setAnnotations(restored);
@@ -545,7 +545,7 @@ const App: React.FC = () => {
         viewerRef.current?.applySharedAnnotations(restored);
       }, 100);
     }
-  };
+  }, [restoreDraft]);
 
   // Fetch available agents for OpenCode (for validation on approve)
   const { agents: availableAgents, validateAgent, getAgentWarning } = useAgents(origin);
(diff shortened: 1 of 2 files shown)


backnotprop:
### Code review

Found 1 issue:

1. `draftKey` is computed once at server startup from `options.rawPatch` but never updated when the user switches diff types via `/api/diff/switch`. After a switch, `currentPatch` is updated (line 158) but all draft operations continue using the original key. If the server crashes after a diff switch, restored annotations will be applied against the wrong patch content.

https://github.com/backnotprop/plannotator/blob/3df3e5de68df61bd48e7cca8a411945c801799c2/packages/server/review.ts#L85-L92

The `/api/diff/switch` handler updates `currentPatch` but not `draftKey`:

https://github.com/backnotprop/plannotator/blob/3df3e5de68df61bd48e7cca8a411945c801799c2/packages/server/review.ts#L156-L161

PR #227, opened 2026-03-05
fix: include untracked new files in uncommitted/unstaged code review diffs

### LINKED ISSUE #189: `plannotator-review` UI fails to display newly created files
I noticed that the review interface only renders files that have been modified. Newly added files are missing from the view.

zerone0x:
## Summary

Fixes #189

Newly created files (untracked by git) were missing from the `plannotator-review` UI because `git diff HEAD` and `git diff` only output changes for files that are already tracked. Any file created by the agent but not yet staged with `git add` was invisible to the code review interface.

## Root Cause

In `packages/server/git.ts`, the `runGitDiff()` function uses:

- `git diff HEAD` for the **uncommitted** view
- `git diff` for the **unstaged** view

Both commands silently omit untracked files.

## Fix

Add a `getUntrackedFileDiffs()` helper that:

1. Discovers untracked files via `git ls-files --others --exclude-standard`
   (respects `.gitignore`, so `node_modules`, `dist`, etc. are excluded automatically)
2. Generates a standard unified diff for each file via `git diff --no-index --src-prefix=a/ --dst-prefix=b/ /dev/null <file>`

The resulting diffs are appended to the tracked-file diff in both the **uncommitted** and **unstaged** modes. The **staged** / **last-commit** / **branch** modes are not affected (they work correctly as-is).

## Changes

- `packages/server/git.ts` — add `getUntrackedFileDiffs()` helper; update `uncommitted` and `unstaged` cases in `runGitDiff()` to include its output (+47/-4 lines)

---
🤖 Generated with Claude Code

### DIFF
--- a/packages/server/git.ts
+++ b/packages/server/git.ts
@@ -97,6 +97,41 @@ export async function getGitContext(): Promise<GitContext> {
 }
 
 /**
+ * Get diffs for untracked (new) files not yet added to git.
+ *
+ * `git diff HEAD` and `git diff` only show tracked files. Newly created files
+ * that haven't been staged with `git add` are invisible to those commands.
+ * This helper discovers them via `git ls-files --others --exclude-standard` and
+ * generates a proper unified diff for each using `git diff --no-index`.
+ *
+ * Note: `git diff --no-index` exits with code 1 when files differ (standard git
+ * behaviour), so we use `.nothrow()` to avoid treating that as an error.
+ */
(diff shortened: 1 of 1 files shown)


backnotprop:
Nice!

backnotprop:
@zerone0x

Thanks for [#227](https://github.com/backnotprop/plannotator/pull/227).

PR #233, opened 2026-03-06
fix(annotate): resolve file paths correctly on Windows

theflysurfer:
## Problem

The `plannotator annotate <file.md>` command fails to resolve file paths correctly on Windows when:

1. The file path uses `@` prefix (Claude Code file reference syntax)
2. The working directory changes between the caller and the Bun script (e.g., Pi extension sets `cwd` but the script runs from a different directory)
3. Relative paths are resolved against the wrong base directory

`path.resolve(filePath)` uses `process.cwd()` which may differ from the original caller's working directory.

## Fix

- Strip `@` prefix if present (common in Claude Code `@/path/to/file` references)
- Support `PLANNOTATOR_CWD` env var to pass the original working directory
- Use `path.isAbsolute()` + `path.join(originalCwd, filePath)` instead of bare `path.resolve()`
- Add optional debug logging via `PLANNOTATOR_DEBUG` env var

## Testing

Tested on Windows 11 with Pi coding agent. Both relative and absolute paths resolve correctly now.

### DIFF
--- a/apps/hook/server/index.ts
+++ b/apps/hook/server/index.ts
@@ -37,6 +37,7 @@ import {
 } from "@plannotator/server/annotate";
 import { getGitContext, runGitDiff } from "@plannotator/server/git";
 import { writeRemoteShareLink } from "@plannotator/server/share-url";
+import path from "path";
 
 // Embed the built HTML at compile time
 // @ts-ignore - Bun import attribute for text
--- a/apps/hook/server/index.ts
+++ b/apps/hook/server/index.ts
@@ -111,15 +111,33 @@ if (args[0] === "review") {
   // ANNOTATE MODE
   // ============================================
 
-  const filePath = args[1];
+  let filePath = args[1];
   if (!filePath) {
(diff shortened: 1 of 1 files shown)


backnotprop:
### Code review

Found 1 issue:

1. **`path` module used but never imported -- will crash at runtime.** The PR removes the dynamic `const path = await import("path")` but continues to call `path.isAbsolute()` and `path.join()`. There is no top-level `import path from "path"` in this file, so the annotate subcommand will throw `ReferenceError: path is not defined` on every invocation. Fix: add `import path from "path"` to the top-level imports, or keep a local dynamic import before the new usage.

https://github.com/backnotprop/plannotator/blob/55f9577b54fc35f3998f96f43448627347c947a8/apps/hook/server/index.ts#L120-L131

backnotprop:
### Code review

backnotprop:
@theflysurfer

Thanks for [#233](https://github.com/backnotprop/plannotator/pull/233).

PR #256, opened 2026-03-09
fix(vscode): reuse IPC server port across restarts

### LINKED ISSUE #252: VS Code extension uses stale port when terminal is restored on reopen
## Description

When VS Code is closed and reopened, restored terminals retain the old `PLANNOTATOR_VSCODE_PORT` environment variable. The extension starts a new IPC server on a different random port (via `server.listen(0, "127.0.0.1")` in `ipc-server.ts`), but the restored terminal still has the old port value baked into its environment.

When `bin/open-in-vscode` runs `curl` against the stale port, the connection fails silently and Plannotator doesn't open in a webview panel.

## Steps to Reproduce

1. Open VS Code with the Plannotator extension installed
2. Open an integrated terminal — note `PLANNOTATOR_VSCODE_PORT` is set (e.g., `echo $PLANNOTATOR_VSCODE_PORT`)
3. Close VS Code completely
4. Reopen VS Code — the terminal is restored with the old env vars
5. Trigger Plannotator (e.g., run a Claude session that exits plan mode)
6. The router script (`bin/open-in-vscode`) tries to connect to the old port, which no longer exists

## Expected Behavior

Plannotator opens in a VS Code webview panel, regardless of whether the terminal was restored.

## Actual Behavior

Nothing happens. The `curl` call in `open-in-vscode` fails silently (`>/dev/null 2>&1`) because the IPC server is listening on a new port that the terminal doesn't know about.

## Root Cause

In `extension.ts` (lines 67-74), the port is injected into the terminal environment once during activation:

```typescript
context.environmentVariableCollection.replace(
  "PLANNOTATOR_VSCODE_PORT",
  String(port),
);
```

The `environmentVariableCollection` API *does* update newly created terminals, but restored terminals from a previous session may retain the stale value from the old `environmentVariableCollection` state until VS Code propagates the update — which doesn't always happen reliably for already-running shell processes.

Meanwhile, `ipc-server.ts` binds to port `0` (random) every time:

```typescript
server.listen(0, "127.0.0.1", () => { ... });
```

So after a restart, the extension gets a new random port, but restored terminals still reference the old one.

## Relevant Files

- `apps/vscode-extension/src/extension.ts` — sets `PLANNOTATOR_VSCODE_PORT` via `environmentVariableCollection`
- `apps/vscode-extension/src/ipc-server.ts` — creates IPC server on random port
- `apps/vscode-extension/bin/open-in-vscode` — router script that reads `PLANNOTATOR_VSCODE_PORT` and curls the IPC server

backnotprop:
## Summary
- Persist the IPC server port in `workspaceState` and try to rebind to it on extension activation
- Falls back to a random port if the preferred one is taken (e.g., another VS Code window claimed it)
- Fixes restored terminals having a stale `PLANNOTATOR_VSCODE_PORT` after VS Code restart

Closes #252

## Test plan
- [ ] Install extension, open terminal, note `echo $PLANNOTATOR_VSCODE_PORT`
- [ ] Close and reopen VS Code — verify the same port is reused
- [ ] Trigger plannotator in a restored terminal — verify it opens in a webview panel
- [ ] Open two VS Code windows on different workspaces — verify no port collision

### DIFF
--- a/apps/vscode-extension/src/extension.ts
+++ b/apps/vscode-extension/src/extension.ts
@@ -48,13 +48,16 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
     vscode.window.showInformationMessage("Plannotator panel opened");
   };
 
-  // Start local IPC server to receive URLs from the router script
+  // Start local IPC server to receive URLs from the router script.
+  // Reuse the last port so restored terminals still have a valid PLANNOTATOR_VSCODE_PORT.
+  const lastPort = context.workspaceState.get<number>("ipcPort");
   const { server, port } = await createIpcServer((url) => {
     openInPanel(url).catch((err) => {
       log.error(`[open] failed: ${err}`);
       vscode.window.showErrorMessage(`Plannotator: ${err}`);
     });
-  });
+  }, lastPort);
+  context.workspaceState.update("ipcPort", port);
   context.subscriptions.push({ dispose: () => server.close() });
 
   // Inject env vars into integrated terminals
--- a/apps/vscode-extension/src/ipc-server.ts
+++ b/apps/vscode-extension/src/ipc-server.ts
@@ -6,31 +6,39 @@ import * as http from "http";
  */
 export function createIpcServer(
   onUrl: (url: string) => void,
+  preferredPort?: number,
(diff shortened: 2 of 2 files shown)


backnotprop:
### Code review

PR #264, opened 2026-03-10
feat: Improve Mermaid diagram viewing experience

flex-yj-kim:
## Summary

This PR improves the end-to-end Mermaid diagram viewing experience, especially for large architecture diagrams that were difficult to inspect in the default plan width.

## What changed

- Introduced a stronger Mermaid viewing flow centered around:
	- expanded fullscreen diagram mode
	- practical zoom controls (zoom in/out, fit-to-view)
	- quick source/diagram toggling
- Updated fit behavior to re-center and fit diagrams to the current viewport/container, rather than only resetting zoom.
- Improved wheel zoom behavior in expanded view so navigation feels consistent during review.
- Rendered the expanded Mermaid overlay at the document level to avoid layout-context clipping.
- Improved control clarity and readability (icon tuning, control grouping, zoom badge).
- Increased viewer widths in both plan view and plan-diff view to better support wide diagrams.
- Ensured the new fullscreen Mermaid viewing flow works consistently in Safari by stabilizing overlay and viewport behavior.

## Screenshots

image

image

PR #266, opened 2026-03-10
feat: Add Graphviz diagram rendering in plan mode

flex-yj-kim:
## Summary

- Extend the shared plan renderer (`Viewer`) to render Graphviz fenced code blocks by supporting `graphviz`, `dot`, and `gv` fence languages.
- Add a new `GraphvizBlock` powered by `@viz-js/viz` with Mermaid-aligned UX: source/diagram toggle, zoom/pan, and expanded view.
- Add a Graphviz example to the sample plan content (`packages/editor/App.tsx`) so the behavior is easy to validate locally.

## Screenshots

image

image

image

### DIFF
--- a/packages/ui/components/GraphvizBlock.tsx
+++ b/packages/ui/components/GraphvizBlock.tsx
@@ -161,9 +161,13 @@ export const GraphvizBlock: React.FC<{ block: Block }> = ({ block }) => {
       try {
         const viz = await getVizInstance();
         const renderedSvg = await viz.renderString(block.content, { format: 'svg' });
+        const cleaned = renderedSvg
+          .replace(/ width="[^"]*"/, ' width="100%"')
+          .replace(/ height="[^"]*"/, ' height="100%"')
+          .replace(/ style="[^"]*"/, '');
         if (!cancelled) {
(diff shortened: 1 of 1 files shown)


backnotprop:
### Code review

Found 2 issues:

1. `event.ctrlKey` guard in `handleWheelZoom` blocks trackpad pinch-to-zoom on macOS. Browsers synthesize `WheelEvent` with `ctrlKey: true` for pinch gestures, so this early return silently disables the most natural zoom gesture on trackpads. This is the same bug that was flagged in PR #264 for MermaidBlock.

https://github.com/backnotprop/plannotator/blob/75220c2a05f345925ee9779a2e9e480cd09c3837/packages/ui/components/GraphvizBlock.tsx#L131-L135

2. `handleWheelZoom` uses React's synthetic `onWheel` handler with `event.preventDefault()`, but React registers wheel events as passive listeners, meaning `preventDefault()` is silently ignored. This causes the page to scroll while the user tries to zoom the diagram. MermaidBlock solved this with an imperative `addEventListener('wheel', handler, { passive: false })` instead.

https://github.com/backnotprop/plannotator/blob/75220c2a05f345925ee9779a2e9e480cd09c3837/packages/ui/components/GraphvizBlock.tsx#L126-L140

PR #274, opened 2026-03-11
fix: VS Code extension doesn't work in SSH remote sessions

### LINKED ISSUE #259: VS Code extension doesn't work in SSH remote sessions
## Summary

The VS Code extension relies on a local IPC server (`127.0.0.1`) to receive URLs from the router script. In SSH remote sessions, the router script runs on the remote machine and cannot reach the VS Code IPC server running locally, making the extension non-functional over SSH.

## Current behavior

- The IPC server in `ipc-server.ts` binds to `127.0.0.1`
- The cookie proxy in `cookie-proxy.ts` also binds to `127.0.0.1`
- The `open-in-vscode` router script uses `curl` to POST to `localhost:PLANNOTATOR_VSCODE_PORT`
- In SSH sessions, this `curl` runs on the remote host and can't reach the local IPC server
- Editor annotations (`editor-annotations.ts`) also only work locally

## What works today

| Scenario | Status |
|----------|--------|
| Local terminal | Works |
| Devcontainer | Works (VS Code bridges `BROWSER` env var) |
| SSH + `PLANNOTATOR_REMOTE=1` | Works, but **without** the extension (browser-only) |
| SSH + VS Code extension | Does not work |

## Workaround

Users can fall back to browser mode with port forwarding:
```bash
export PLANNOTATOR_REMOTE=1
export PLANNOTATOR_PORT=9999
```
Then forward port 9999 manually. This bypasses the extension entirely.

## Possible approaches

- Leverage VS Code's `vscode.env.asExternalUri()` (already used in `panel-manager.ts`) to tunnel the IPC server port
- Use VS Code's remote port forwarding API to auto-forward the IPC port
- Alternative IPC mechanism that works across SSH (e.g., VS Code's `postMessage` or filesystem-based signaling)

## Related

- `packages/server/remote.ts` — remote session detection
- `apps/vscode-extension/src/ipc-server.ts` — IPC server
- `apps/vscode-extension/src/cookie-proxy.ts` — cookie proxy
- `apps/vscode-extension/bin/open-in-vscode` — router script

7tg:
## Links

Closes #259

## Summary

In VS Code Remote SSH sessions, the Plannotator extension never opens a panel because `handleServerReady()` skips `openBrowser()` when it detects a remote session. However, the VS Code extension sets `PLANNOTATOR_BROWSER` to its own `open-in-vscode` handler, which should be respected regardless of session type.

**Fix:** When `PLANNOTATOR_BROWSER` is set, always call `openBrowser()` — even in remote sessions. This env var is the explicit signal that a custom handler (the VS Code extension) exists.

This is a one-line condition change in the shared handler used by all three modes (plan review, code review, annotate).

## Validation

- In a VS Code Remote SSH session, open an integrated terminal and run Claude Code with the plannotator plugin
- Trigger a plan review — the panel should now open in VS Code
- Verify code review (`/plannotator-review`) and annotate (`/plannotator-annotate`) also work in remote sessions
- Verify local (non-SSH) sessions are unaffected

### DIFF
--- a/packages/server/shared-handlers.ts
+++ b/packages/server/shared-handlers.ts
@@ -112,13 +112,13 @@ export function handleDraftDelete(contentKey: string): Response {
   return Response.json({ ok: true });
 }
 
-/** Open browser for local sessions. Used by all 3 servers. */
+/** Open browser for local sessions or when a custom handler (e.g. VS Code extension) is configured. */
 export async function handleServerReady(
   url: string,
   isRemote: boolean,
   _port: number,
 ): Promise<void> {
-  if (!isRemote) {
+  if (!isRemote || process.env.PLANNOTATOR_BROWSER) {
     await openBrowser(url);
   }
 }


backnotprop:
@7tg

Thanks for [#274](https://github.com/backnotprop/plannotator/pull/274).

7tg:
Thanks @backnotprop

PR #276, opened 2026-03-11
fix: enable linked doc navigation in annotate mode

dgrissen2:
## Summary

- **Bug:** Clicking `.md` links in annotated files (e.g., `[37% IQM](comparison_iqm_37.md)`) showed "Failed to connect to server" because the annotate server had no `/api/doc` endpoint — requests fell through to the SPA catch-all, returning HTML instead of JSON.
- **Root cause:** The `/api/doc` handler (`handleDoc`) was only registered in the plan server (`packages/server/index.ts`), not the annotate server (`packages/server/annotate.ts`).
- **Fix:** Three targeted changes:
  1. **`annotate.ts`** — Add `/api/doc` route, auto-injecting the source file's parent directory as `?base=` so relative sibling links (e.g., `comparison_iqm_37.md` from `plots/gamma_segments/comparison_iqm_35.md`) resolve from the correct location
  2. **`reference-handlers.ts`** — `handleDoc()` now accepts an optional `?base=` query param and tries resolving relative to it before falling back to the project-root glob search
  3. **`App.tsx`** — `handleOpenLinkedDoc` passes the current file's directory as `?base=` when navigating linked-doc → linked-doc, so chained relative links also resolve correctly

## Test plan

- [ ] `plannotator annotate` a markdown file that contains relative `.md` links to sibling files
- [ ] Click a `.md` link — linked doc should load in-place with back button
- [ ] From the linked doc, click another `.md` link — should resolve relative to *that* doc's directory
- [ ] Click back — should return to the original annotated file
- [ ] Plan server linked doc navigation still works (no regression)

### DIFF
--- a/packages/editor/App.tsx
+++ b/packages/editor/App.tsx
@@ -506,9 +506,19 @@ const App: React.FC = () => {
     if (vaultBrowser.activeFile && vaultPath) {
       linkedDocHook.open(docPath, buildVaultDocUrl(vaultPath));
     } else {
-      linkedDocHook.open(docPath);
+      // Pass the current file's directory as base for relative path resolution
+      const baseDir = linkedDocHook.filepath
+        ? linkedDocHook.filepath.replace(/\/[^/]+$/, '')
+        : imageBaseDir;
+      if (baseDir) {
+        linkedDocHook.open(docPath, (path) =>
+          `/api/doc?path=${encodeURIComponent(path)}&base=${encodeURIComponent(baseDir)}`
+        );
+      } else {
+        linkedDocHook.open(docPath);
+      }
     }
-  }, [vaultBrowser.activeFile, vaultPath, linkedDocHook, buildVaultDocUrl]);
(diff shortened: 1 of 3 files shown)


backnotprop:
### Code review

Found 1 issue:

1. `imageBaseDir` is referenced but never defined in `App.tsx`. It is not declared as state, a prop, an import, or a constant anywhere in the file. This will cause a `ReferenceError` at runtime when a user clicks a relative `.md` link in annotate mode and `linkedDocHook.filepath` is null (i.e., the first navigation from the root annotated file). The fallback `imageBaseDir` is always `undefined`, so `if (baseDir)` is false, and the code falls through to `linkedDocHook.open(docPath)` without any `?base=` param — defeating the relative path resolution for the initial hop. It appears the intent was to capture `filePath` from the `/api/plan` response (which `annotate.ts` already sends at line 121) into component state, but that wiring is missing.

https://github.com/backnotprop/plannotator/blob/0477ae3e89982b5d8afc6474a644025e1bbd6dc5/packages/editor/App.tsx#L508-L521

PR #282, opened 2026-03-12
feat: merge Send Feedback and Approve into single adaptive button

theflysurfer:
## Summary

Replace the two separate action buttons (Send Feedback + Approve) with a single button that adapts based on annotation count:

- **Has annotations** → **Send Feedback** (accent color, chat icon on mobile)
- **No annotations** → **Approve** (green/success, OK on mobile)

## Motivation

Simplifies the review UI by removing the need for users to choose between two actions. The correct action is always obvious based on context.

## Changes

- Single `<button>` that switches behavior/style based on `totalAnnotationCount`
- Removed `showApproveWarning` state (no longer needed — Approve only shows when there are 0 annotations)
- Removed the `ConfirmDialog` warning for approving with annotations
- Removed the separate Approve button and its hover tooltip
- Net: **-40 lines** (23 added, 63 removed)

### DIFF
--- a/packages/review-editor/App.tsx
+++ b/packages/review-editor/App.tsx
@@ -147,7 +147,6 @@ const ReviewApp: React.FC = () => {
   const [isSendingFeedback, setIsSendingFeedback] = useState(false);
   const [isApproving, setIsApproving] = useState(false);
   const [submitted, setSubmitted] = useState<'approved' | 'feedback' | false>(false);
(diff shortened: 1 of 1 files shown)


backnotprop:
Cool idea, I'll give it a run today.

backnotprop:
@theflysurfer whyd you close?

PR #286, opened 2026-03-14
feat: remote session notifications via Tailscale and ntfy

### LINKED ISSUE #192: Remote Claude Code sessions: no way to open plan review UI without a local browser

tfriedel:
## Summary

Fixes #192 — remote Claude Code sessions can now open the plan review UI without port forwarding.

**Note:** This is not a general-purpose solution — it relies on Tailscale, tmux, and/or ntfy, which are not common dependencies. It's shared as a working proof-of-concept and inspiration for how remote access can work. A more general solution (e.g., a session router on a fixed port, or first-class support for Claude Code surfacing hook messages) is still needed for broad adoption. See #192 for discussion.

### What's here

- **Tailscale auto-detection**: Detects Tailscale hostname via `tailscale status --self --json`, so the server URL is directly reachable (e.g. `http://mybox.ts.net:PORT`). Also supports explicit `PLANNOTATOR_HOSTNAME` env var.
- **Random ports in remote mode**: Removed the fixed port 19432 default, so parallel Claude sessions work without port conflicts.
- **`0.0.0.0` binding**: Server binds to all interfaces when a non-localhost hostname is detected.
- **tmux popup**: When `$TMUX` is set, shows a `tmux display-popup` with the server URL — works inside Claude Code's TUI since it's a tmux-layer overlay that bypasses Claude's stdio capture.
- **ntfy push notification**: When `CLAUDE_HOOKS_NTFY_URL` is set (compatible with existing claude-code ntfy hooks), sends a push notification with a clickable link.
- **Short URLs via paste service**: Uses the existing paste service (zero-knowledge encrypted) for compact share links when the server isn't directly reachable.

### Why this isn't the general solution

The core problem is that **Claude Code doesn't surface hook stderr to the user** — it captures it to a socket. This means there's no built-in way for a hook to communicate a URL (or anything else) to the user while it's running.

The workarounds here are all external channels:
- **Tailscale** — requires the user to have a mesh VPN set up
- **tmux popup** — requires the user to be running inside tmux
- **ntfy** — requires a push notification service

For users without any of these, the only option remains manual SSH port forwarding (`ssh -L PORT:localhost:PORT`), which is fragile since ports are now random.

### Ideas for a general solution

1. **Claude Code feature request**: A hook progress/message channel that surfaces stderr (or a dedicated stream) to the user's terminal. This would solve the problem at the root.
2. **Session router**: A long-lived server on a fixed port that multiplexes multiple plan sessions on subpaths (`/session/<id>`). Users forward one port, all sessions are accessible. More complex but works without external dependencies.
3. **VS Code Remote auto-forwarding**: VS Code Remote already auto-forwards ports. If the server prints a `localhost:PORT` URL to stderr, VS Code picks it up. This works today for VS Code Remote users — the missing piece is just surfacing the URL.

## Test plan

- [x] Tested on remote SSH session with Tailscale — tmux popup shows `http://hostname:port`, opens in local browser, full approve/deny works
- [x] Tested ntfy push notification — clickable link arrives on phone
- [ ] Verify local (non-remote) mode is unchanged — should still open browser directly
- [x] Verify parallel Claude sessions get different random ports
- [ ] Verify fallback when Tailscale is not installed (falls back to localhost + share URL)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -133,6 +133,9 @@ claude --plugin-dir ./apps/hook
 | `PLANNOTATOR_SHARE` | Set to `disabled` to turn off URL sharing entirely. Default: enabled. Can also be set via `~/.plannotator/config.json` (`{ "share": "disabled" }`); the env var takes precedence. |
 | `PLANNOTATOR_SHARE_URL` | Custom base URL for share links (self-hosted portal). Default: `https://share.plannotator.ai`. |
 | `PLANNOTATOR_PASTE_URL` | Base URL of the paste service API for short URL sharing. Default: `https://plannotator-paste.plannotator.workers.dev`. |
+| `PLANNOTATOR_HOSTNAME` | Explicit hostname for remote URLs (e.g. `mybox.ts.net`). Auto-detected from Tailscale if not set. |
+| `CLAUDE_HOOKS_NTFY_URL` | Full ntfy URL (e.g. `https://ntfy.sh/mytopic`). When set, remote review URLs are sent as push notifications. |
+| `CLAUDE_HOOKS_NTFY_TOKEN` | Optional ntfy auth token (Bearer). |
 | `PLANNOTATOR_ORIGIN` | Explicit agent-origin override at the top of the detection chain. Valid values: `claude-code`, `amp`, `droid`, `opencode`, `codex`, `copilot-cli`, `gemini-cli`, `kiro-cli`, `pi`. Invalid values silently fall through to env-based detection. Unset by default. |
 | `PLANNOTATOR_JINA` | Set to `0` / `false` to disable Jina Reader for URL annotation, or `1` / `true` to enable. Default: enabled. Can also be set via `~/.plannotator/config.json` (`{ "jina": false }`) or per-invocation via `--no-jina`. |
 | `JINA_API_KEY` | Optional Jina Reader API key for higher rate limits (500 RPM vs 20 RPM unauthenticated). Free keys include 10M tokens. |
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -126,9 +126,9 @@ claude --plugin-dir ./apps/hook
 
 | Variable | Description |
 |----------|-------------|
-| `PLANNOTATOR_REMOTE` | Set to `1` / `true` for remote mode, `0` / `false` for local mode, or leave unset for SSH auto-detection. Uses a fixed port in remote mode; browser-opening behavior depends on the environment. |
+| `PLANNOTATOR_REMOTE` | Set to `1` / `true` for remote mode, `0` / `false` for local mode, or leave unset for SSH auto-detection. Uses a random port and binds beyond localhost in remote mode; browser-opening behavior depends on the environment. |
 | `PLANNOTATOR_AGENT_TERMINAL_REMOTE` | Set to `1` / `true` to enable the annotate-mode agent terminal while `PLANNOTATOR_REMOTE` is active. Off by default because remote mode binds beyond localhost. |
-| `PLANNOTATOR_PORT` | Fixed port to use. Default: random locally, `19432` for remote sessions. |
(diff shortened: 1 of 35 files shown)


backnotprop:
Thank you. I will take a look at this and test it out.

tfriedel:
I've been trying this out since yesterday and for me the the ntfy system works best. I prefer that over the tmux popups. 
I get a notifaction popup on my desktop, click on it and the plan opens immediately.

This is the script I use for that:
https://github.com/joshsymonds/nix-config/blob/main/home-manager/claude-code/hooks/ntfy-notifier.sh

I also use it to get notifications when claude asks for permissions.

backnotprop:
Closing this older, conflicted proof of concept while remote capabilities are investigated under #899. The original issue #192 is already closed. Thank you for documenting and testing the Tailscale, tmux, and ntfy approach.

PR #310, opened 2026-03-16
Unify review core across Bun and Pi

### LINKED ISSUE #307: Bug: still missing untracked files in pi-extension
On code review using pi, it's still not showing untracked file
Maybe related: https://github.com/backnotprop/plannotator/pull/227

backnotprop:
## Summary
- extract a runtime-agnostic review core for diff assembly, file-content lookup, worktree handling, and stage/unstage operations
- switch the Bun review path to the shared core and copy the same source into the Pi package at build time
- bring the Pi review server up to the shared review API surface, including diff switching, file content, drafts, uploads/images, agents, and editor annotations
- add regression coverage for the shared review core and the Pi review server endpoints

Fixes #307.

## Testing
- bun test packages/shared/review-core.test.ts apps/pi-extension/server.test.ts
- bun test

## Notes
- The full test run still has one unrelated pre-existing failure in `packages/ui/components/diagramLanguages.test.ts` because `@viz-js/viz` is not installed in this environment.

### DIFF
--- a/apps/marketing/public/favicon.svg
+++ b/apps/marketing/public/favicon.svg
@@ -1,5 +1,5 @@
 <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 64 64">
-  <rect width="64" height="64" rx="14" fill="#6B3FA0"/>
+  <rect width="64" height="64" rx="14" fill="#070b14"/>
   <rect x="12" y="28" width="40" height="14" rx="3" fill="#E0BA55" opacity="0.35"/>
   <text x="32" y="46" text-anchor="middle" font-family="Inter,system-ui,sans-serif" font-weight="800" font-size="42" fill="white">P</text>
 </svg>


backnotprop:
### Code review

Found 1 issue:

1. Pi `plannotator-review` handler silently drops inline annotations. The review server's `waitForDecision()` returns `{ approved, feedback, annotations, agentSwitch }` (line 693-699 of `server.ts`), but the command handler in `index.ts` only reads `result.feedback` and ignores `result.annotations`. Since this PR adds the full annotation API surface to the Pi review server, users can now create inline code annotations in the review UI that are accepted by the server but never forwarded to the agent — the annotations are silently lost.

https://github.com/backnotprop/plannotator/blob/39cf90c33dd96c701fe8f7fffd172bb805585e6e/apps/pi-extension/index.ts#L300-L311

The Bun hook server avoids this because it bakes annotations into the `feedback` string via `exportAnnotations()` before submission. The Pi handler needs equivalent logic — either call `exportAnnotations()` to merge annotations into the feedback text, or handle `result.annotations` separately.

Issue #313, opened 2026-03-16
Feature Request: Collaborative Rooms


grubmanItay:
Problem: Getting team feedback on a plan requires a tedious export/import cycle per person — export link, teammate reviews locally, exports their link back, you import it, repeat for every reviewer.

  Proposed solution: A "Create Room" button in the Share tab that creates a live, shared session. The creator gets a URL to send to teammates. Everyone who opens it sees the same plan and can annotate it simultaneously, with
  annotations syncing across all participants via polling (~2.5s). Only the room creator can approve/deny. The room URL grants access (no auth required).

  Architecture notes from prototyping:
  - Rooms can be hosted on the existing paste service — no WebSocket/tunneling needed. The paste service stores room metadata + an append-only annotation log keyed by room:{id}:meta and room:{id}:ann:{timestamp}
  - Requires adding a list(prefix) method to the PasteStore interface (FS store scans dataDir, KV store uses native list API)
  - 5 new API endpoints: create room, get room state, push annotations, poll annotations (with cursor), close room
  - E2E encrypted with AES-256-GCM — key lives in the URL fragment, never sent to server. Creator token is hashed server-side (SHA-256)
  - Client needs: useRoom hook (manages join/poll/push lifecycle), RoomBanner component (status bar with participants), name prompt on join
  - Polled annotations must go through applySharedAnnotations() to anchor to the DOM — same pipeline as shared URL import
  - Room URL should use window.location.origin as base (not the share portal) so it works both locally and in production
  - The annotate server needs pasteApiUrl passed through (currently missing from AnnotateServerOptions)
  - Use theme-aware colors (semantic tokens like bg-primary, bg-secondary, etc.) for participant avatars, not hard-coded Tailwind colors

backnotprop:
We should have a discussion on this. I've definitely been thinking about this. There's been a few other apps that kind of create a collaborative live room for agents we could evaluate

backnotprop:
https://github.com/backnotprop/plannotator/pull/316

More context from the group, if collaborating, would be good

schelip:
any updates on this issue, @backnotprop? really looking forward to it!

backnotprop:
This is live in beta: https://room.plannotator.ai

Upload a plan (or paste markdown, or fetch any URL), create an encrypted room, share the link. Collaborators see your plan and annotations in real time. Their changes sync to you.

Key properties:
[...]

PR #316, opened 2026-03-16
feat: add collaborative review sessions

eprj453:
Add server-side collaborative review sessions allowing multiple team members to review and annotate a plan using a single shared URL.

**Problem Solved:**
Previously, to get feedback from N reviewers, you needed N separate share URLs (each reviewer creates their own annotated version and sends it back). Now, create one session URL and all reviewers add annotations to the same session.

**Backend Changes:**
- Add ReviewSession, CreateReviewSessionRequest, AddAnnotationsRequest types
- Extend PasteStore interface with session methods
- Implement session storage for filesystem and Cloudflare KV
- Add 3 new API endpoints:
  - POST /api/review-session (create new session)
  - GET /api/review-session/:id (fetch session state)
  - PATCH /api/review-session/:id/annotations (add annotations with optimistic locking)

**Frontend Changes:**
- Add useCollaborativeSession hook for session management
- Add CollaborativeSessionButton UI component
- Integrate collaborative session button in plan editor toolbar

**Key Features:**
- One fixed URL for all reviewers (no N-way URL ping-pong)
- Server merges annotations automatically (deduplicates)
- Optimistic locking prevents version conflicts
- Auto-expires after 7 days (same as paste service)
- Works offline (local filesystem) or cloud (Cloudflare KV)

**Files Changed:**
- packages/ui/types.ts (+50 lines)
- apps/paste-service/core/storage.ts (+10 lines)
- apps/paste-service/stores/fs.ts (+50 lines)
- apps/paste-service/stores/kv.ts (+30 lines)
- apps/paste-service/core/handler.ts (+210 lines)
- packages/editor/App.tsx (+10 lines)
- CLAUDE.md (+40 lines)

**Files Created:**
- packages/ui/hooks/useCollaborativeSession.ts (260 lines)
- packages/ui/components/CollaborativeSessionButton.tsx (170 lines)
- COLLABORATIVE_REVIEW_GUIDE.md (comprehensive testing guide)

**Tested:**
- ✅ Session creation
- ✅ Session retrieval
- ✅ Annotation merging from multiple reviewers
- ✅ Optimistic locking (version conflict handling)
- ✅ Deduplication
- ✅ Filesystem storage persistence

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -150,6 +150,34 @@ User annotates markdown, provides feedback
 Send Annotations → feedback sent to agent session
 ```
 
+## Collaborative Review Flow
+
+```
+User clicks "Start Collaborative Review" in plan editor
+        ↓
+POST /api/review-session → creates session with unique ID
+        ↓
+Share URL displayed: /s/<session-id>
+        ↓
+User shares URL with team members (Reviewer A, B, C...)
+        ↓
+Each reviewer:
+  1. Opens /s/<session-id> → auto-joins session
+  2. Adds their annotations
+  3. Clicks "Submit" → PATCH /api/review-session/:id/annotations
+     (Optimistic locking: fails if version mismatch)
+        ↓
+Original user clicks "Refresh" → fetches all merged annotations
+        ↓
+User approves/denies plan with all team feedback → sent to agent
+```
+
+**Key features:**
+- One fixed URL for all reviewers (no N-way URL ping-pong)
+- Server merges annotations automatically (deduplicates)
+- Version conflicts auto-resolve with refresh
+- Works offline (local filesystem) or cloud (Cloudflare KV)
+
 ## Server API
 
 ### Plan Server (`packages/server/index.ts`)
@@ -205,8 +233,18 @@ All servers use random ports locally or fixed port (`19432`) in remote mode.
 | --------------------- | ------ | ------------------------------------------ |
 | `/api/paste`          | POST   | Store compressed plan data, returns `{ id }` |
 | `/api/paste/:id`      | GET    | Retrieve stored compressed data            |
+| `/api/review-session` | POST   | Create collaborative review session, returns `{ session, shareUrl }` |
+| `/api/review-session/:id` | GET | Retrieve review session with all annotations |
+| `/api/review-session/:id/annotations` | PATCH | Add annotations to session (optimistic locking) |
 
 Runs as a separate service on port `19433` (self-hosted) or as a Cloudflare Worker (hosted).
+
+**Collaborative Review Sessions:**
+- Fixed session URL (`/s/<id>`) for multiple reviewers
+- Server-side annotation merging with deduplication
+- Optimistic locking (version field) prevents conflicts
+- Auto-expires after 7 days (same as paste TTL)
+- Supports both filesystem and Cloudflare KV storage
 
 ## Plan Version History
 
--- a/COLLABORATIVE_REVIEW_GUIDE.md
+++ b/COLLABORATIVE_REVIEW_GUIDE.md
@@ -1,0 +1,345 @@
+# Collaborative Review Sessions - Implementation Guide
+
+## Overview
+
+This PR adds collaborative review sessions to Plannotator, allowing multiple team members to review and annotate a plan using a single shared URL.
+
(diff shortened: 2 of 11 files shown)


backnotprop:
Hey @eprj453 what was the motivation behind this, what type of testing have you done

osujin:
We're running plannotator on a self-hosted EKS deployment and this is the feature that would make it genuinely useful for team review. The paste service extension approach is clean — it builds on existing infrastructure without requiring a separate real-time backend.

grubmanItay:
Related to #313.                                                                                                                                                                                                                  
                                                                                                   
  Great work @eprj453 — this covers a lot of ground and tackles the hardest part: the full storage layer, API endpoints, both FS and KV backends, the hook, and the UI integration. The overall architecture (extending the paste   
  service rather than adding a separate real-time backend) is exactly the right call. Really nice to see this taking shape.                                                                                                         

  A few things I ran into while prototyping the same feature locally that might save you some debugging. Tagging @backnotprop for his take on these.

  **Encryption**

  The existing share flow uses AES-256-GCM with the key in the URL fragment so the server never sees plan content. This PR stores plans and annotations as plaintext JSON on the paste service. That's a regression from
  plannotator's current security model — anyone with server/KV access can read everything. I'd expect this to use the same `@plannotator/shared/crypto` encrypt/decrypt flow with a `#key=...` fragment in the session URL.

  **No real-time sync**

  Reviewers have to manually click "Refresh" to see others' annotations and "Submit" to push their own. For a collaborative session this is a dealbreaker in practice — you end up with version conflicts and stale state. A polling
   loop (even a simple 2-3s setInterval) would make this actually usable.

  **Remote annotations won't render as highlights**

  `refreshSession()` merges server annotations via `setAnnotations()` directly, but annotations from `fromShareable()` come back with `blockId: ''`, `startOffset: 0`, and no `startMeta`/`endMeta`. They need to go through
  `viewerRef.current.applySharedAnnotations()` to get anchored to the DOM — otherwise they exist in state but are invisible in the document. I hit this exact bug when prototyping this.

  **Optimistic locking on KV is racy**

  The KV `updateSession` does get-then-put (non-atomic) which the code acknowledges. With 3+ reviewers submitting around the same time, annotations get silently dropped. An append-only log (one key per annotation batch, e.g.
  `session:{id}:ann:{timestamp}`) avoids the read-modify-write race entirely and scales better since you're not rewriting the full session on every update.

  **No session lifecycle**

  There's no concept of a session owner, closing a session, or approve/deny. In practice the person who created the session needs to be able to lock it when review is done.

  **Theme**

  Minor: `CollaborativeSessionButton` uses hard-coded colors (`bg-purple-600` etc.) — should use the semantic theme tokens (`bg-primary`, `bg-accent`, etc.) so it works with both dark and light themes.

  @backnotprop what's your thinking on this? Happy to share more detail from my prototype if it helps.

backnotprop:
Yes all these things matter @grubmanItay - also willing to meet over call for this work

eprj453:
Thanks everyone for the thoughtful feedback — this is very helpful.

You’re right that many of these points should have been discussed earlier in an issue/discussion before going this far in implementation. I’ll make sure to do that for design-heavy changes going forward.

I still believe that enabling shared review workflows for even small teams can make plannotator much more broadly useful and drive wider adoption of the project. If follow-up issues or PRs are opened around this direction, I’d be happy to actively support and contribute.

backnotprop:
Diving deeper into the PR this weekend

backnotprop:
Hey @eprj453 — thank you for this PR and for pushing collaborative sessions forward. We took a different direction with the implementation to solve two things we felt were non-negotiable:

1. **Privacy** — plans are end-to-end encrypted. The server stores and relays ciphertext only. Room secrets live in the URL fragment and never reach the server. No Supabase, no third-party data store that can read your content.
2. **Performance** — the room engine runs on Cloudflare Durable Objects with WebSocket multiplexing, SQLite-backed event sequencing, and a zero-dependency protocol. No external database round-trips.

The result is live at https://room.plannotator.ai — multiplayer annotation for plans, specs, markdown, anything. E2E encrypted, rooms auto-expire, and agents can join as first-class peers.

This ships in #591. Closing this PR in favor of that work, but we genuinely appreciate the contribution and the signal that this was something people wanted. 🙏

PR #323, opened 2026-03-17
fix: opencode review cwd context flow

jwyce:
## What
- Fix OpenCode review to use the project directory from command context (`ctx.directory`) instead of relying on implicit process cwd.
- Add `cwd` to `GitContext` and add `runGitDiffWithContext()` so review calls use one consistent context object.
- Update follow-up review endpoints (`/api/diff/switch`, `/api/file-content`, `/api/git-add`) to reuse `gitContext.cwd`.
- Keep worktree behavior unchanged: explicit worktree paths still override default cwd.
## Why
- In some setups (tmux/server/attach flows), process cwd can differ from the repository being reviewed.
- That could produce empty or incorrect diffs after the initial review open.
- This change makes review behavior deterministic by carrying cwd as explicit session context.
## Validation
```bash
bun test packages/shared/review-core.test.ts
bun run build:review && bun run build:hook && bun run build:opencode
Passed locally.
```
## Notes
- Scope is review flow only.
- Annotate flow still uses process.cwd() and is intentionally unchanged in this PR.

### DIFF
--- a/apps/opencode-plugin/index.ts
+++ b/apps/opencode-plugin/index.ts
@@ -31,7 +31,7 @@ import {
   startAnnotateServer,
   handleAnnotateServerReady,
 } from "@plannotator/server/annotate";
-import { getGitContext, runGitDiff } from "@plannotator/server/git";
+import { getGitContext, runGitDiffWithContext } from "@plannotator/server/git";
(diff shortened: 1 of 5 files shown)


backnotprop:
@jwyce

Thanks for [#323](https://github.com/backnotprop/plannotator/pull/323).

PR #344, opened 2026-03-19
feat(review): add global diff search

### LINKED ISSUE #343: [FEATURE REQUEST] ability to search code diffs
hey, I do not know if this exists somewhere in the UI, but what do you think about adding search on the diffs? would look something like this

Image

sercantor:
## Summary
- add global diff search to the review UI
- highlight matches in the diff and search results sidebar
- closes #343


backnotprop:
Commits pulled and merged with #347

Issue #348, opened 2026-03-20
Code Review Enhancements & GitHub PR Integration


backnotprop:
## Overview

Extend the PR review mode with richer GitHub integration — contextual PR information, issue linking, direct PR actions, and multi-PR navigation.

Currently PR review shows the diff and supports annotations, but lacks the surrounding context that makes a review actionable: PR metadata, comments, merge status, linked issues, and the ability to act on the PR directly from the UI.

---

## 1. PR Context Panel

Display rich PR metadata in the review UI — a panel or tab alongside annotations.

- [x] **PR summary** — title, description, author, branch info (metadata already fetched, just not displayed)
- [x] **PR comments** — conversation thread from GitHub
- [x] **Merge status** — checks passing/failing, review approvals, merge conflicts
- [x] **Linked issues** — issues referenced in the PR description or commits

### Open questions
- ~Tab within the annotation panel, sidebar section, or collapsible area?~ → Icon tabs inline in the panel header

## 2. Issue Attachment in Feedback

Let reviewers reference GitHub issues in their annotations/feedback, so when feedback is sent back to the agent or posted as a PR comment, it links to relevant issues for context.

- [ ] Attach existing GitHub issues to annotations
- [ ] Issue references included in exported feedback markdown

## 3. PR Actions

Actions that execute directly against the GitHub PR from the review UI.

- [ ] **Approve PR** — posts a GitHub approval review (not just local LGTM)
- [ ] **Post feedback as PR comment** — annotations become a review comment on GitHub
- [ ] **Squash & merge** — after approval, optionally merge the PR

### Open questions
- Direct automation (Plannotator calls `gh` itself) vs agent-mediated (send "merge this PR" as feedback to the agent) vs both as user-configurable options?
- Should squash & merge be a separate explicit action, or offered as a follow-up after approval?

## 4. PR Navigation

- [ ] **Browse open PRs** — switch between PRs without restarting (like the diff-type switcher for local reviews)
- [ ] **Navigate to linked issues** — from the context panel, open referenced issues

backnotprop:
Iterating on this.

NaikSoftware:
Add support for Github enterprise URLs, please. Like company.ghe.com/my/repo/pull/123

backnotprop:
GitHub Enterprise support shipped in [v0.16.4](https://github.com/backnotprop/plannotator/releases/tag/v0.16.4). PR review now works with any GHE host. See PR #460.

PR #352, opened 2026-03-20
feat: Approve and review PR to github

rockneurotiko:
I really love the pull request review, and I've seen on #348 the action 3 to approve/review to github directly.

That's something I really wanted, here's a PR for it. Only the approve/review, the squash&merge is left for other PR because it requires discussion around the usage.

This PR adds:

- Adds a toggle on review mode for PRs to change between agent and github:
image
image

- Use `gh api` to send the reviews, since `gh review` only allow general comments and not per-file-line comments

- Since reviews don't have global comments, when clicking on send review or approve, a popup will appear to write the general review comment.
image
image

- On submitting, it will open the PR link, and send a message to the agent with:
  - feedback is set to "Pull request approved on GitHub: <url>" or "Pull request reviewed on GitHub: <url>"
  - annotations is empty []
  - approved is false (so the agent doesn't treat it as a code approval)

### DIFF
--- a/packages/review-editor/App.tsx
+++ b/packages/review-editor/App.tsx
@@ -640,7 +640,10 @@ const ReviewApp: React.FC = () => {
     }).filter(c => c.body.length > 0);
 
     // Editor annotations (VS Code extension) — always on new/RIGHT side
+    // Only include annotations targeting files in the diff to avoid GitHub API rejection
+    const diffPaths = new Set(files.map(f => f.path));
     for (const ea of editorAnnotations) {
+      if (!diffPaths.has(ea.filePath)) continue;
       const body = ea.comment || `> ${ea.selectedText}`;
       if (!body.trim()) continue;
       const isMultiLine = ea.lineStart !== ea.lineEnd;
@@ -657,7 +660,7 @@ const ReviewApp: React.FC = () => {
     }
 
     return { action, body, fileComments };
-  }, [annotations, editorAnnotations]);
+  }, [annotations, editorAnnotations, files]);
 
   // Submit a review directly to GitHub
   const handleGitHubAction = useCallback(async (action: 'approve' | 'comment', generalComment?: string) => {
@@ -770,7 +773,8 @@ const ReviewApp: React.FC = () => {
 
       if (githubMode) {
         // GitHub mode: No annotations → Approve on GitHub, otherwise → Post Review
-        if (totalAnnotationCount === 0) {
+        const isOwnPR = !!ghUser && prMetadata?.author === ghUser;
+        if (totalAnnotationCount === 0 && !isOwnPR) {
           setGithubGeneralComment('');
           setGithubCommentDialog({ action: 'approve' });
         } else {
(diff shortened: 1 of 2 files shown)


backnotprop:
awesome will take a look

backnotprop:
Looks like it is working great @rockneurotiko

backnotprop:
Bug

  1. Loading state text is wrong — When submitting in GitHub mode, the buttons show
  "Approving..." even when doing "Send Feedback". The loading text should match the action
  being taken.

  UX Improvements

  2. Button labels should differ between modes — In GitHub mode, "Send Feedback" / "Approve"
  should say something like "Post Comments" / "Post Approval" (or similar) to make it clear
  the action goes to GitHub, not the agent. Helps delineate the two modes.
  3. "Open PR in browser" should be opt-in — Currently it always opens a new tab after
  submitting. This should be a checkbox or toggle in the general comment dialog, something
  like "Open PR link after submitting". Not always wanted.

  Tested & Working

  4. Sending feedback via the general comment dialog — works
  5. Inline code annotations posted as GitHub comments — works well

backnotprop:
Tested this out locally, the core flow works well. Here's what I found:

## Bugs

1. **Loading state text is wrong** — When submitting "Send Feedback" in GitHub mode, both buttons show a submitting state, but the Approve button text changes to "Approving..." even though that's not the action being taken. The approve button should just become disabled without a text change.

2. **Multi-line annotations collapse to a single line on GitHub** — When selecting a range of lines (e.g., 304-318) and submitting via GitHub mode, the comment only anchors to the last line. The payload sends `line` but not `start_line`/`start_side`, so GitHub can't render it as a ranged comment. This also breaks multi-line suggestions since "Apply suggestion" won't map to the correct span. Happy to take a stab at fixing this.

## UX Improvements

3. **Button labels should differ between modes** — In GitHub mode, "Send Feedback" / "Approve" should say something like "Post Comments" / "Post Approval" to make it clear the action targets GitHub, not the agent.

4. **"Open PR in browser" should be opt-in** — Currently it always opens a new tab after submitting. A checkbox in the general comment dialog like "Open PR link after submitting" would be better. Not always wanted.

## Tested & Working

- Sending feedback via the general comment dialog works
- Inline code annotations posted as GitHub comments work well

backnotprop:
Im happy to address these btw! I think i might in order to get next release today.

rockneurotiko:
@backnotprop thanks for the amazing review! 

Feel free to chime in with changes, it's already late here (Europe timezone) and I don't know if I'll be able to adress the issues today

backnotprop:
I got it & will make the changes - no worries! @rockneurotiko , great work!! have a good weekend

apps/review/package.json
```diff
@@ -13,6 +13,7 @@
     "@plannotator/server": "workspace:*",
     "@plannotator/ui": "workspace:*",
     "@pierre/diffs": "^1.1.0-beta.19",
+    "dompurify": "^3.3.3",
     "react": "^19.2.3",
     "react-dom": "^19.2.3",
     "tailwindcss": "^4.1.18",
```
backnotprop: Testing approval with annotations.

backnotprop (dismissed):
Comment for approved annotation.

backnotprop (commented):
Another test.

rockneurotiko:
@backnotprop thanks for all the changes, it required much more tweaks than I expected, sorry for that 😅

PR #353, opened 2026-03-20
feat: show repo identifier in tab title (#342)

### LINKED ISSUE #342: feat: Identifier in tab title
When having multiple plannotator tabs open in the browser, the current page title makes it hard to navigate them, especially in cases when these are for different repos.

This issue is for requesting to think of a quick-win solution which could be shipped in the next version(s).

One very simple could be: use repo or dir name like:
`repo/name ・ Plannotator`, or
`dir-name ・ Plannotator`

Below screenshot is from Arc Browser, having 2 plannotator tabs on 2 different repos:

|  |
| - |

VincentHardouin:
## Summary
- Adds dynamic page title with repo/directory name for both plan review and code review tabs
- Format: `{repoName} · Plannotator` or `{repoName} · Code Review`
- Makes it easy to distinguish multiple Plannotator tabs from different repos

Closes #342

### DIFF
--- a/packages/editor/App.tsx
+++ b/packages/editor/App.tsx
@@ -95,6 +95,13 @@ const App: React.FC = () => {
   const [shareBaseUrl, setShareBaseUrl] = useState<string | undefined>(undefined);
   const [pasteApiUrl, setPasteApiUrl] = useState<string | undefined>(undefined);
   const [repoInfo, setRepoInfo] = useState<{ display: string; branch?: string } | null>(null);
+
+  useEffect(() => {
+    if (repoInfo) {
+      document.title = `${repoInfo.display} · Plannotator`;
+    }
+  }, [repoInfo]);
+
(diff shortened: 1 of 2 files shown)


backnotprop:
Hey thanks! I will double check, I think we need some type of fallback incase there isnt repo info - there are edge cases where it can be null:
  ```
  const [repoInfo, setRepoInfo] = useState<{ display: string; branch?: string } | null>(null);
  ```
  
  You could fall back to just "Plannotator"  - ensure no " · "

VincentHardouin:
Done! Added explicit fallback — title now resets to 'Plannotator' / 'Code Review' when repoInfo is null.

backnotprop (approved):

backnotprop:
@VincentHardouin

Thanks for [#353](https://github.com/backnotprop/plannotator/pull/353).

PR #355, opened 2026-03-20
fix(parser): support nested markdown code fences

brian-malinconico-axomic:
## Summary

- Count backticks in the opening fence and require the same number (or more) to close it, instead of always closing on any ` ``` ` line
- Fix language tag extraction to slice by actual fence length rather than hardcoded 3
- Add `devbox.json` with `devbox run test` wired to `bun test`
- Add `packages/ui/utils/parser.test.ts` — first test file for the parser, 8 tests all passing

## The problem

A plan that shows a markdown example — e.g. explaining how to write a code block — uses a 4-backtick fence to wrap content that itself contains triple backticks:

````markdown
Here is how to write a code block in your plan:

```js
console.log("hello");
```
````

Before this fix, the parser closed the outer block at the first ` ``` ` line, producing two broken blocks instead of one. The nested content was either lost or misattributed to a following paragraph.

## Spec context

CommonMark (§4.5) defines the rule clearly: a closing fence must use **the same character** (backtick or tilde) and have **at least as many** of them as the opening fence. A line with fewer backticks is treated as content, not a delimiter.

However, CommonMark is not a universal standard — support for nested fences varies significantly across tools:

| Tool / renderer | Nested fence support |
|---|---|
| GitHub Flavored Markdown | ✅ Follows CommonMark — 4-backtick fence wraps 3-backtick content correctly |
| VS Code preview | ✅ Follows CommonMark |
| Obsidian | ✅ Supported |
| older Pandoc defaults | ⚠️ Varies by output format and flag; `--from markdown+fenced_code_blocks` handles it |
| Reddit / older Reddit MD | ❌ No nested fence support — only triple backtick, no length distinction |
| Some static site generators | ❌ Depend on underlying library; many only implement triple-backtick |

In practice, plans submitted by Claude frequently contain markdown documentation examples, shell scripts with heredocs, or instructions that themselves reference code blocks. These are the exact cases where a 4-backtick outer fence is necessary and expected to work.

## Changes

```
packages/ui/utils/parser.ts       — 3-line fix in the code-block branch
packages/ui/utils/parser.test.ts  — new file, 8 tests
devbox.json / devbox.lock         — pin bun for running tests locally
bun.lock                          — updated lockfile
```

## Test plan

- [x] `devbox run test` — 267 tests pass across 23 files
- [x] Triple-backtick fences unaffected (regression test)
- [x] 4-backtick fence with inner ` ``` ` treated as content
- [x] 5-backtick fence treats both ` ``` ` and ```` ```` ```` as content
- [x] Closing fence with more backticks than opener closes the block
- [x] Nested-fence block does not swallow surrounding paragraphs
- [x] Unclosed fence at EOF produces a code block (CommonMark §4.5)
- [x] Fence opener as last line produces an empty code block (no crash)
- [x] Manually verified in browser via the plannotator plan review UI


backnotprop:
@brian-malinconico-axomic

Thanks for [#355](https://github.com/backnotprop/plannotator/pull/355).

PR #366, opened 2026-03-22
fix: annotate-last resolves wrong session after cd

janah01:
## Problem

`/plannotator-last` annotates the wrong message when the user `cd`'s during a Claude Code session. No error, no warning — just the wrong content in the annotation UI.

### What happens

`findSessionLogsForCwd()` builds a project slug from `process.cwd()`. Claude Code's session log lives under the slug for the *original* session directory, not the current shell CWD.

1. Session starts in `/Users/me` → log stored at `~/.claude/projects/-Users-me/<id>.jsonl`
2. User runs `cd ~/projects/my-app`
3. `/plannotator-last` uses CWD → looks in `~/.claude/projects/-Users-me-projects-my-app/`
4. Finds a stale session from a previous day that happened to use that directory
5. Opens the annotation UI with that old session's last message

The user has no way to tell something went wrong until they read the text and realize it's not what they expected. By then they may have already written annotations against the wrong content.

### Who this affects

Anyone who `cd`'s during a session. That's a common workflow — `cd` into a repo, run commands, then annotate the assistant's response.

This is a Claude Code-only bug. Codex uses `CODEX_THREAD_ID`, OpenCode and Pi use their SDK APIs — none of them resolve sessions via CWD. The fix is scoped to the Claude Code `else` branch; no other harness is touched.

## Fix

Three-tier session resolution (most precise first):

**1. PPID session metadata** (new, primary path)

Claude Code writes `~/.claude/sessions/<pid>.json`:

```json
{"pid": 17326, "sessionId": "112c3961-...", "cwd": "/Users/me", "startedAt": ...}
```

When `/plannotator-last` runs, its parent process is the Claude Code instance. `process.ppid` → read the metadata → get the exact session ID and original CWD → open the exact JSONL. O(1), no scanning, no mtime heuristics.

**2. CWD slug match** (existing behavior, unchanged)

Works when the shell CWD hasn't diverged from the session's project directory.

**3. Ancestor walk** (new fallback)

Walk up the directory tree trying parent slugs. Handles the case where PPID metadata is unavailable and the user `cd`'d deeper into a subdirectory.

### Other changes

- `getLastRenderedMessage()` now catches read errors instead of throwing on missing/corrupt files, so the fallback cascade doesn't crash mid-resolution.
- Extracted `tryLogCandidates()` to deduplicate the candidate iteration loop (was copy-pasted between CWD and ancestor fallbacks).

## Test plan

- [x] 44 tests pass (`bun test apps/hook/server/`)
- [x] 3 new tests for `findSessionLogsByAncestorWalk` (root returns empty, walks up to parent, skips exact CWD)
- [x] This is a Claude Code-only bug. Codex uses `CODEX_THREAD_ID`, OpenCode and Pi use their SDK APIs — none of them resolve sessions via CWD. The fix is scoped to the Claude Code `else` branch; no other harness is touched.

### DIFF
--- a/apps/hook/server/index.ts
+++ b/apps/hook/server/index.ts
@@ -51,7 +51,7 @@ import { registerSession, unregisterSession, listSessions } from "@plannotator/s
 import { openBrowser } from "@plannotator/server/browser";
 import { detectProjectName } from "@plannotator/server/project";
 import { planDenyFeedback } from "@plannotator/shared/feedback-templates";
-import { findSessionLogsForCwd, getLastRenderedMessage, type RenderedMessage } from "./session-log";
+import { findSessionLogsForCwd, resolveSessionLogByPpid, findSessionLogsByAncestorWalk, getLastRenderedMessage, type RenderedMessage } from "./session-log";
 import { findCodexRolloutByThreadId, getLastCodexMessage } from "./codex-session";
 import path from "path";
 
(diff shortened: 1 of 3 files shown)


backnotprop:
nice work

backnotprop:
making one adjustment for windows

backnotprop:
@janah01

Thanks for [#366](https://github.com/backnotprop/plannotator/pull/366).

PR #368, opened 2026-03-22
Add OpenPackage manifest

TiagoJacinto:
Enables installation via opkg install plannotator

Adds openpackage.yml manifest for opkg compatibility.

### DIFF
--- a/openpackage.yml
+++ b/openpackage.yml
@@ -1,0 +1,8 @@
+name: plannotator
+version: 0.8.0
+description: Interactive Plan Review for Claude Code. Annotate plans visually, share with team, and send feedback to agents.
+author: backnotprop
+license: MIT
+
+includes:
+  - apps/hook/**/*


backnotprop:
Hey thanks @TiagoJacinto - I'm not familiar with OpenPackage. Can you just give me an idea of what this does and enables?

TiagoJacinto:
OpenPackage is a package manager for AI. It drops the files where your editor expects them. It works with Claude, Cursor, Windsurf, OpenCode, and dozens more.

You already have a marketplace.json, the difference is that marketplace.json only reaches Claude users, while openpackage.yml includes that file plus gets you into every other platform. It adds cleaner uninstalls, discoverability on openpackage.dev, opkg install backnotprop/plannotator works everywhere.

backnotprop:
@TiagoJacinto

Thanks for [#368](https://github.com/backnotprop/plannotator/pull/368).

PR #383, opened 2026-03-24
feat: resizable split pane for diff viewer

backnotprop:
## Summary
- Adds a draggable divider between left/right panes in split diff view so users can resize each side
- Uses CSS custom properties that inherit into Pierre's shadow DOM — drag updates inline style vars, not `unsafeCSS`, so zero Pierre re-renders during drag
- Ratio clamped to 20-80%, persisted via cookie storage, double-click resets to 50/50
- Handle hidden for add-only/delete-only files (Pierre renders these as single-column even in split mode)
- Adds `pierre-guard` skill (`.agents/skills/`) documenting the full @pierre/diffs API surface we depend on

## Test plan
- [ ] `bun run dev:review` — open split diff view, drag the divider between panes
- [ ] Verify ratio persists after page refresh
- [ ] Double-click divider — should reset to 50/50
- [ ] Switch to unified view — divider should disappear
- [ ] Select an add-only or delete-only file — divider should not appear
- [ ] Verify theme colors on hover/active match expectations (uses `bg-primary` tokens)
- [ ] Check console for regressions (existing `parseLineType` warning is pre-existing)


backnotprop:
### Code review

PR #393, opened 2026-03-25
feat: Sync github Viewed files

rockneurotiko:
Bidirectional sync between plannotator's file-viewed checkboxes and GitHub's native "Mark as viewed" feature:

On load: fetches per-file viewerViewedState via GraphQL and pre-populates the viewed checkmarks

On toggle: fires a background markFileAsViewed/unmarkFileAsViewed GraphQL mutation when a file is checked/unchecked

GitLab: no-op, GitLab's viewed state is localStorage-only with no API

Also adds prNodeId (GraphQL node ID) to GithubPRMetadata, fetched alongside existing gh pr view call, to be able to call github with the viewed status. Implemented across both the Bun server and Pi extension to maintain route parity.

### DIFF
--- a/apps/pi-extension/server/pr.ts
+++ b/apps/pi-extension/server/pr.ts
@@ -10,7 +10,9 @@ import {
 	fetchPRContext as fetchPRContextCore,
 	fetchPR as fetchPRCore,
 	fetchPRFileContent as fetchPRFileContentCore,
+	fetchPRViewedFiles as fetchPRViewedFilesCore,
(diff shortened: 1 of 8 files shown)


backnotprop:
This is cool, when a file changes on GitHub, its viewed status gets changed to dismissed. 

Not that we need to do anything here, but just curious if you have thoughts on how we would address this.

backnotprop:
Sorry about the noise on the earlier review comment — disregard. LGTM, merging!

PR #395, opened 2026-03-25
Unix update command for wsl users on UpdateBanner

alexandresilvestri:
I use Plannotator in WSL, so when I copy the update command, I get it for Windows instead of for Unix.

I exported the "isWSL" function that checks if the user is using WSL to the function that defines the update command.


backnotprop:
@alexandresilvestri

Thanks for [#395](https://github.com/backnotprop/plannotator/pull/395).

PR #420, opened 2026-03-28
feat: print support with export menu integration and keyboard shortcut

Yecats:
## Summary

- Adds a full print stylesheet for clean white-paper output from the plan review UI
- Moves the print button from a standalone toolbar button into the export dropdown menu
- Adds `Ctrl+P` / `⌘P` keyboard shortcut for printing, following existing shortcut patterns
- Registers the shortcut in the keyboard shortcuts display panel
- Adds print option to the mobile hamburger menu

## Details

**Print stylesheet** (`packages/ui/print.css`): Comprehensive `@media print` rules — A4 layout, hidden UI chrome, forced light colors for code blocks, proper typography, page break hints. Uses a `plannotator-print` CSS class (toggled via `beforeprint`/`afterprint` events) to override highlight.js dark theme in code blocks.

**Keyboard shortcut**: New `useEffect` handler following the same pattern as the existing `Ctrl+S` save shortcut — skips when focus is in input fields, skips when modals are open, prevents default browser print to ensure the `usePrintMode` hook fires properly.

**Platform support**: Uses `(e.metaKey || e.ctrlKey)` for the shortcut and `modKey` from `packages/ui/utils/platform.ts` for display — renders as `⌘P` on Mac, `Ctrl+P` on Windows/Linux.


apps/hook/commands/annotate.md
backnotprop: Is there a purpose for this file

backnotprop:
@Yecats Im going to remove those additional command files, cool?

They are redundant:

`apps/hook/commands/annotate.md` is new, but `apps/hook/commands/plannotator-annotate.md` already exists.

lmk if you think we should have better shorthand commands

Yecats:
@backnotprop  Absolutely - I didn't mean to include that, so sorry! I had started to make the hook before I realized that you already had what I needed. I'd meant to roll back, but apparently I didn't.

PR #317, opened 2026-03-16
feat: bidirectional AI conversation, context persistence, and approve-with-notes

sudorest:
## Summary

This PR addresses two major user complaints and adds three interconnected features inspired by [Octto](https://github.com/vtemian/octto)'s branch-based Q&A patterns:

1. **Approve with Notes** — Users can now add annotations AND approve the plan in one step. No more forced choice between "approve silently" or "deny with feedback."

2. **Context Persistence** — Review sessions now survive across deny/revise cycles. Previous iterations' feedback, questions, and answers are accumulated and fed back to the AI with each revision.

3. **Bidirectional AI Conversation** — The AI can now embed clarification questions in plans that appear as interactive question cards in the review UI. Users answer visually; responses flow back through the approve/deny feedback loop.

### Phase 1: Approve with Notes (smallest change, immediate value)

- **`packages/shared/feedback-templates.ts`**: New `planApproveWithNotesFeedback()` template
- **`apps/hook/server/index.ts`**: Pass user annotations through as `message` field in the `allow` decision (Claude Code)
- **`packages/editor/App.tsx`**: Replace "Annotations Won't Be Sent" warning with "Approve with Notes" confirmation dialog; always show full-green Approve button

### Phase 2: Context Persistence

- **`packages/shared/questions.ts`** (NEW): Types for `ReviewIteration`, `ReviewSession`, `ClarificationQuestion`, `QuestionAnswer`; plan metadata parsing (`extractSessionId`, `extractQuestions`, `stripPlanMetadata`)
- **`packages/server/review-session.ts`** (NEW): Session persistence to `~/.plannotator/sessions/`, with auto-cleanup of sessions older than 7 days
- **`packages/shared/feedback-templates.ts`**: Enhanced `planDenyFeedback()` now includes accumulated iteration history and Q&A context
- **`apps/hook/server/index.ts`** & **`apps/opencode-plugin/index.ts`**: Session tracking (extract/generate ID, load previous iterations, record after decision)
- **`packages/server/index.ts`**: New `GET /api/session` endpoint for UI to fetch session state
- **`packages/ui/components/sidebar/ReviewHistory.tsx`** (NEW): Review History sidebar tab with iteration timeline, expandable feedback summaries, and Q&A display
- **`packages/ui/components/sidebar/SidebarContainer.tsx`** & **`SidebarTabs.tsx`**: History tab integration
- **`packages/ui/hooks/useSidebar.ts`**: Added `"history"` to `SidebarTab` type

[...]

PR #425, opened 2026-03-29
fix: harden follow-up flows for review sessions

sudorest:
## Summary
- fix Review History clarification answer matching so stored answers render by question id
- harden approve/deny/annotate/question submission flows so failed API responses no longer appear successful
- correct the shared questions import path used by the server question session helper

## Verification
- bun test packages/ui/utils/parser.test.ts packages/ui/utils/bear.test.ts apps/vscode-extension/src/panel-manager.test.ts
- attempted bun run build:hook *(blocked in this environment because  was not available on PATH)*

## Context
Follow-up to #317 to tighten correctness and UX trust around the new review session and clarification question flows.

### DIFF
--- a/packages/editor/App.tsx
+++ b/packages/editor/App.tsx
@@ -473,14 +473,18 @@ const App: React.FC = () => {
         headers: { 'Content-Type': 'application/json' },
         body: JSON.stringify(answer),
       });
-      if (res.ok) {
-        const data = await res.json();
-        if (data.answers) {
-          setClarificationAnswers(data.answers);
-        }
+      if (!res.ok) {
+        throw new Error('Failed to submit clarification answer');
+      }
+
+      const data = await res.json();
+      if (data.answers) {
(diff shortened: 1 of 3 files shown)


sudorest:
Follow-up to #317 to tighten correctness around the new review-session flows.

Reasoning for this PR:
- The new clarification question flow added more API-dependent UI paths, so failed , , , or  responses could otherwise look successful from the editor.
- Review History was matching answers against question text instead of question id, which could hide previously captured clarification answers even when they were stored correctly.
-  referenced the wrong shared import path, which is a small but important correctness/build consistency fix.

The goal here is intentionally narrow: improve trust in the new UX from #317 without changing the feature design or broadening scope.

sudorest:
Closing this because I based it on the wrong upstream context. I intended to create a follow-up to #364, not #317, so I will open a corrected replacement PR.

Issue #431, opened 2026-03-30
feat: Context Anchoring — preserve plan decision reasoning across sessions


aviadshiber:
## Background

This is inspired by [Martin Fowler's article on context anchoring](https://martinfowler.com/articles/reduce-friction-ai/context-anchoring.html), which describes the pattern of externalizing AI decisions into a persistent document to prevent context loss across sessions.

The core insight: *code shows what was built, not what was rejected or why*. Today, when Claude denies a plan and Claude re-plans, the agent only gets the current denial feedback. It doesn't know:
- What approaches were already tried and rejected in earlier iterations
- *Why* they were rejected (not just what to change)
- What constraints emerged from previous review rounds
- Whether the same mistake is being made again

## Proposal: Iterative Decision Chronicle

When a plan goes through multiple review cycles (deny → revise → re-review), plannotator should accumulate a **decision chronicle** alongside the version history. This chronicle would:

1. **Record what was rejected and why** — each denied iteration contributes a "rejected: reason" entry extracted from annotations
2. **Surface the chronicle in subsequent denial feedback** — when the agent resubmits after a denial, the chronicle is prepended to the feedback:
   ```
   ## Previous Iteration Decisions
   - Iteration 1 denied: [user annotation summary] — "Don't use Redis caching here, latency budget is <5ms"
   - Iteration 2 denied: [user annotation summary] — "Stop proposing background jobs, this must be synchronous"

   ## Current Feedback
   ...
   ```
3. **Per-feature decision file** — following the existing convention of one file per slug (`~/.plannotator/plans/{slug}-approved.md`, `{slug}-denied.md`), write a parallel `~/.plannotator/decisions/{project}/{slug}.md` that accumulates ADR entries for that feature across iterations. One file per feature, not one document for all decisions.

## Why This Maps to Context Anchoring

The article describes a two-layer structure:
- **Project-level priming** (stable) — tech stack, conventions
- **Feature-level document** (evolving) — decisions made, rejected alternatives + reasoning, constraints, open questions

Plannotator's archive (`~/.plannotator/plans/`) already stores *what* was decided. The gap is **why** — the reasoning embedded in annotations is currently only used for single-session feedback and then discarded from the agent's perspective.

## Concrete Implementation Ideas

1. **Decision Chronicle / Lightweight ADRs** (`packages/shared/storage.ts`): alongside version history, accumulate a `chronicle.md` per plan slug in ADR format. Each denied iteration contributes a record:
   ```markdown
[...]

PR #432, opened 2026-03-30
feat(feedback): add context anchoring instructions to plan denial

aviadshiber:
## Summary

Implements the context anchoring pattern from [Martin Fowler's article on context anchoring](https://martinfowler.com/articles/reduce-friction-ai/context-anchoring.html), applied to both the plan denial and approval flows.

**The problem:** when an agent gets a plan denied and revises, it only receives the current round's feedback — not what was rejected in prior rounds. It can re-propose the same approaches without knowing they were already rejected, creating iteration loops. And on approval, nothing reminds the agent to carry forward the decision history into implementation.

### `packages/shared/feedback-templates.ts`

**`planDenyFeedback()`** — adds a `## Context Anchoring` section instructing the agent to maintain a `## Decisions Log` in the plan itself:

```
## Context Anchoring

Before revising your plan:
1. Add (or update) a `## Decisions Log` section at the bottom of the plan.
2. For each rejected approach from this feedback, add an entry:
   - **Rejected:** [brief description]  **Why:** [reason from this feedback]
3. Do NOT re-propose approaches already listed in the Decisions Log — it is your cross-session memory.
```

**`planApproveFeedback()`** — new function used on plan approval, reminding the agent to reference the Decisions Log during implementation:

```
Plan approved!

If your plan contains a `## Decisions Log`, keep it as a reference during
implementation — it documents the rejected alternatives that shaped this design.
```

### `apps/opencode-plugin/index.ts`

Uses `planApproveFeedback()` for both approval paths (with notes and without), replacing the previous inline strings.

### `packages/shared/feedback-templates.test.ts`

Adds 4 tests covering the context anchoring behavior (all passing):
- Deny includes `Decisions Log` instructions, `Rejected:` entry format, and cross-session memory note
- Approve includes `Decisions Log` reminder
- Approve with notes includes both implementation notes and `Decisions Log` reminder
- Approve with saved path surfaces the file path

## Why this approach

The plan is already the persistent artifact (saved to version history on every submission). By embedding rejected alternatives and their reasoning directly in the plan markdown, the ADR pattern works at zero infrastructure cost. The Decisions Log:
- Travels with every revision
- Is visible in the plan diff view (shows what was added/changed between iterations)
- Is archived with the final approved/denied decision
- Survives context window resets — it's in the document, not just the conversation

## Test plan

- [x] Deny → agent instructed to build `## Decisions Log` with rejected approaches
- [x] Deny again → Decisions Log grows ("add or update" instruction)
- [x] Approve → agent reminded to reference Decisions Log during implementation
- [x] Approve with notes → notes + Decisions Log reminder both appear

Part of #431

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/apps/opencode-plugin/index.ts
+++ b/apps/opencode-plugin/index.ts
@@ -40,7 +40,7 @@ import {
   handleArchiveCommand,
   type CommandDeps,
 } from "./commands";
-import { planDenyFeedback } from "@plannotator/shared/feedback-templates";
+import { planDenyFeedback, planApproveFeedback } from "@plannotator/shared/feedback-templates";
 import {
   normalizeEditPermission,
   stripConflictingPlanModeRules,
@@ -473,20 +473,7 @@ Do NOT proceed with implementation until your plan is approved.`);
               }
             }
 
-            if (result.feedback) {
-              return `Plan approved with notes!
-${result.savedPath ? `Saved to: ${result.savedPath}` : ""}
-
-## Implementation Notes
-
-The user approved your plan but added the following notes to consider during implementation:
-
-${result.feedback}
-
-Proceed with implementation, incorporating these notes where applicable.`;
-            }
-
-            return `Plan approved!${result.savedPath ? ` Saved to: ${result.savedPath}` : ""}`;
+            return planApproveFeedback(result.feedback, result.savedPath);
           } else {
             return planDenyFeedback(result.feedback || "", "submit_plan", {
               planFilePath: sourceFilePath,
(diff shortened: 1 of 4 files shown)


backnotprop:
Hey @aviadshiber , thanks for this PR and sorry for taking so long to get back to you. The context anchoring idea is great.

We ended up going a different direction with the implementation though. Instead of hardcoding the context anchoring into the denial template, we built a configurable feedback pipeline that lets users customize all plan/annotation/review messages via `~/.plannotator/config.json` with `{{variable}}` templates. Your context anchoring pattern is now a first-class config example in the docs rather than a default for everyone.

The new docs page walks through how to set it up: users drop your exact denial + approval templates into their config and they're done. You're credited by name with a link to your GitHub profile as the contributor behind the example.

We're going to close this PR, but **you will get co-author credit on the commit** that ships the feedback pipeline. Thanks again for the contribution, it directly shaped how we documented and designed the customization system.

PR #450, opened 2026-03-31
fix(review): resolve repo root for untracked file discovery

### LINKED ISSUE #449: Bug: untracked files missing from code review when agent CWD is a subdirectory
## Description

When running `/plannotator-review`, untracked (new) files are missing from the file tree. Only untracked files that happen to be inside the agent's current working directory are shown, and with incorrect relative paths. Tracked/modified files display correctly.

## Steps to reproduce

1. In a monorepo, have the agent `cd` into a subdirectory (e.g., `packages/foo/`)
2. Create untracked files both inside and outside that subdirectory
3. Modify some tracked files
4. Run `/plannotator-review`

**Expected:** All untracked files appear in the review with repository-root-relative paths.
**Actual:** Only untracked files under the agent's CWD appear, with CWD-relative paths (e.g., `lib/Bar.ts` instead of `packages/foo/lib/Bar.ts`). Untracked files in other directories are missing entirely.

## Workaround

Prompt the agent to change back to the repository root before running the review:

> Change your cwd back to the root of the git repo.

## Root cause

In `packages/shared/review-core.ts`, `getUntrackedFileDiffs()` passes the process CWD through to `git ls-files --others --exclude-standard`. Unlike `git diff HEAD` (which always covers the full repo with root-relative paths regardless of CWD), `git ls-files` scopes to the CWD subtree and returns CWD-relative paths.

When the hook process inherits a CWD that is a subdirectory of the repo, `git ls-files` only discovers untracked files in that subtree, and the resulting diff headers use short CWD-relative paths that don't match the root-relative paths from the tracked diff.

## Suggested fix

Resolve the repo root before running `git ls-files` and the per-file `git diff --no-index` commands:

```
async function getUntrackedFileDiffs(runtime, srcPrefix, dstPrefix, cwd) {
  // Resolve repo root so ls-files covers the entire repo
  const toplevel = await runtime.runGit(["rev-parse", "--show-toplevel"], { cwd });
  const rootCwd = toplevel.exitCode === 0 ? toplevel.stdout.trim() : cwd;

  const lsResult = await runtime.runGit(
    ["ls-files", "--others", "--exclude-standard"],
    { cwd: rootCwd },  // was { cwd }
  );

  // ... same file filtering logic ...

  const diffs = await Promise.all(
    files.map(file =>
      runtime.runGit(
        ["diff", "--no-ext-diff", "--no-index", ...prefixes, "/dev/null", file],
        { cwd: rootCwd },  // was { cwd }
      ).then(r => r.stdout)
    ),
  );

  return diffs.join("");
}
```

## Environment

- Plannotator v0.16.2
- macOS, Claude Code plugin
- Reproducible whenever the agent changes its working directory to a subdirectory before review is invoked

blimmer:
## Summary

- Resolve repo root in `getUntrackedFileDiffs` so `git ls-files` lists all untracked files, not just those under the process CWD
- Use the resolved root as CWD for `git diff --no-index` so paths are root-relative, matching tracked-diff output
- Add regression test that runs `runGitDiff` from a subdirectory

## Problem

`git ls-files --others --exclude-standard` scopes to the CWD subtree and returns CWD-relative paths. `git diff HEAD` always covers the full repo with root-relative paths. When the hook process inherits a subdirectory CWD (e.g., an agent that `cd`'d into `packages/foo/`), untracked files outside that subtree are silently dropped and the one that is found gets a wrong relative path in the diff header.

## AI

Note that I used Opus 4.6 to assist with identifying, filing the GitHub issue, and the TDD-resolution of this bug.

Fixes #449

### DIFF
--- a/packages/shared/review-core.test.ts
+++ b/packages/shared/review-core.test.ts
@@ -1,6 +1,6 @@
 import { afterEach, describe, expect, test } from "bun:test";
 import { spawnSync } from "node:child_process";
-import { mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join, resolve as resolvePath } from "node:path";
 import {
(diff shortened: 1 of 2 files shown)


backnotprop:
looks good thanks!

PR #474, opened 2026-04-02
fix(review): expand file tree folders by default on initial render (#473)

### LINKED ISSUE #473: File tree folders are collapsed by default on initial render
## Bug

When opening a code review via `/plannotator-review`, the file tree sidebar starts with all folders collapsed. The user has to manually click "Expand all folders" every time.

## Root Cause

In `FileTree.tsx`, the `expandedFolders` state is initialized as an empty `Set`:

```ts
const [expandedFolders, setExpandedFolders] = useState<Set<string>>(new Set());
const [prevTree, setPrevTree] = useState(tree);

// Expand all folders when tree changes (initial render + diff switch)
if (tree !== prevTree) {
  setPrevTree(tree);
  setExpandedFolders(new Set(getAllFolderPaths(tree)));
}
```

The comment says "initial render + diff switch" but `prevTree` is initialized to `tree` on line 134, so `tree !== prevTree` is `false` on the first render. The expand-all logic never fires on mount, only on subsequent tree changes (e.g., switching diffs).

## Fix

Initialize `expandedFolders` with all folder paths:

```ts
const [expandedFolders, setExpandedFolders] = useState<Set<string>>(
  () => new Set(getAllFolderPaths(tree))
);
```

---

> This issue was debugged and the fix was generated with assistance from Claude Code.

blimmer:
## Summary

- Fix file tree folders starting collapsed on initial render by initializing `expandedFolders` state with all folder paths instead of an empty set

Fixes #473

---

> This PR was generated with assistance from Claude Code.

### DIFF
--- a/packages/review-editor/components/FileTree.tsx
+++ b/packages/review-editor/components/FileTree.tsx
@@ -130,7 +130,7 @@ export const FileTree: React.FC<FileTreeProps> = ({
 
   const tree = useMemo(() => buildFileTree(files), [files]);
 
-  const [expandedFolders, setExpandedFolders] = useState<Set<string>>(new Set());
+  const [expandedFolders, setExpandedFolders] = useState<Set<string>>(() => new Set(getAllFolderPaths(tree)));
   const [prevTree, setPrevTree] = useState(tree);
 
   // Expand all folders when tree changes (initial render + diff switch)


backnotprop:
Nice, thanks!

PR #488, opened 2026-04-05
fix(annotate): support @ markdown file references

Exloz:
## Summary

- Fixes markdown path resolution for annotate when the input uses OpenCode-style file references like `@file.md`.
- Adds a safe fallback in the shared resolver: try the original input first, and only if it is `not_found`, retry after stripping leading `@`.

## Problem

In OpenCode, `/plannotator-annotate README.md` worked, but `/plannotator-annotate @README.md` failed to open annotate mode for the file.
The shared resolver treated `@README.md` as a literal filename and returned `not_found` when only `README.md` existed.

## Root Cause

`resolveMarkdownFile` in `packages/shared/resolve-file.ts` did not support `@`-prefixed file reference syntax.

## Changes

- Refactored current resolution logic into an internal core function (`resolveMarkdownFileCore`).
- Kept `resolveMarkdownFile` as a wrapper that now:
  - trims input
  - accepts wrapped quotes (`"..."` / `'...'`)
  - runs primary resolution using the original normalized input
  - if primary result is `not_found` and input starts with `@`, retries with leading `@` removed
  - preserves the original user input in `not_found`/`ambiguous` responses for clearer logs/errors

## Why this is safe

- If a real file named `@spec.md` (or similar) exists, it is found by the primary lookup and returned directly.
- `@` normalization is only applied after the primary lookup fails.
- Existing non-`@` behavior remains unchanged.
- The fix lives in shared resolver logic, so it is consistent across surfaces that reuse it (annotate + `/api/doc` paths).

## Tests

Added coverage in `packages/server/resolve-file.test.ts`:

- resolves `@README.md` via fallback when `@README.md` does not exist
- prioritizes real `@README.md` when it exists
- resolves quoted input `"@README.md"`
- returns `not_found` for `@nope.md`
- returns `ambiguous` for fallback case like `@plan.md` with multiple `plan.md` matches

Validation run:

- `bun test packages/server/resolve-file.test.ts` (pass)

## Scope

- `packages/shared/resolve-file.ts`
- `packages/server/resolve-file.test.ts`

No command template changes were required.

### DIFF
--- a/packages/server/resolve-file.test.ts
+++ b/packages/server/resolve-file.test.ts
@@ -106,6 +106,33 @@ describe("resolveMarkdownFile", () => {
     });
   });
 
+  test("resolves @ filename via fallback when @ file does not exist", async () => {
+    const root = createTempProject({ "README.md": "# Hello" });
+    const result = resolveMarkdownFile("@README.md", root);
+    expect(result).toEqual({
+      kind: "found",
+      path: resolve(root, "README.md"),
+    });
+  });
+
+  test("prioritizes real @ filename before fallback", async () => {
+    const root = createTempProject({ "@README.md": "# At" });
+    const result = resolveMarkdownFile("@README.md", root);
+    expect(result).toEqual({
+      kind: "found",
+      path: resolve(root, "@README.md"),
+    });
+  });
+
+  test("resolves quoted @ filename", async () => {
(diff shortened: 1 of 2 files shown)


backnotprop:
@Exloz

Thanks for [#488](https://github.com/backnotprop/plannotator/pull/488).

PR #494, opened 2026-04-06
fix(review): show P4 added files in code review

l10veu602:
## Summary
- Fix P4 `add`ed files not appearing in the review UI on Windows
- Normalize `\r\n` to `\n` in P4 command stdout to prevent stray `\r` in file paths
- Replace fragile index-based `p4 where` parsing with `p4 -ztag where` structured output

## Root Cause
On Windows, `p4 where` stdout uses `\r\n` line endings. Splitting on `\n` leaves `\r` at the end of every path except the last line. `Bun.file("path\r")` silently fails to find the file, causing `getNewFileDiff()` to return an empty string and the added file to be omitted from the review.

## Changes
- `runP4()`: strip `\r\n` → `\n` in stdout for all P4 commands
- `getNewFileDiff()`: strip `\r\n` from file content read via `Bun.file()`
- `batchResolveDepotPaths()`: switch from index-based `p4 where` line parsing to `p4 -ztag where` field-based parsing (also more robust for stream depots and overlay mappings)

## Test plan
- [x] Tested on a real P4 workspace with 4 `p4 add` files + 12 `p4 edit` files
- [x] Before fix: 13 files shown (only 1 of 4 added files). After fix: 17 files shown (all 4 added files)
- [x] Verified `Stray \r in patch: 0` after fix

### DIFF
--- a/packages/server/p4.ts
+++ b/packages/server/p4.ts
@@ -32,7 +32,7 @@ async function runP4(
     proc.exited,
   ]);
 
-  return { stdout, stderr, exitCode };
+  return { stdout: stdout.replace(/\r\n/g, "\n"), stderr, exitCode };
 }
 
 // --- Path helpers ---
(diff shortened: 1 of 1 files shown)


backnotprop:
@l10veu602

Thanks for [#494](https://github.com/backnotprop/plannotator/pull/494).

Issue #287, opened 2026-03-14
Support turn by turn and session level diff for code review diff view


prashanth057:
Harnesses such as Claude code supports /diff command that can show diffs which happened in the last turn(last conversation) or current session which is different than git diff. 
Can you add support for such views please?

githendrik:
+1 on this. It could be achieved with staged vs. unstaged hunks in git (staged can have additional diff to unstaged), which easily gives the diff on diff option of what was approved vs. what was touched last

backnotprop:
I think it's out of scope to support previous turn diffs per Harness; because they are not really natively integrated into Git. It becomes difficult to actually track what exactly a last turn was. And I don't want to try to do things like bring in hooks or rationalize what actually a turn is from outside the harness. 

**Unstaged is the closest thing you get if you're staging changes between turns.** - this is now the default.

A future item could try to create file-timestamps at the beginning and end of turns via hooks.

PR #521, opened 2026-04-08
add --unstaged option to view only unstaged files/hunks

githendrik:
Related to this issue: https://github.com/backnotprop/plannotator/issues/287
My personal flow is to stage hunks that I approved in previous steps. This flag allows me to view only the most recent changes.

### DIFF
--- a/apps/hook/server/index.ts
+++ b/apps/hook/server/index.ts
@@ -12,6 +12,10 @@
  *    - Triggered by /review slash command
  *    - Runs git diff, opens review UI
  *    - Outputs feedback to stdout (captured by slash command)
+ *    - Flags:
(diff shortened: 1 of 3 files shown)


backnotprop:
Hey Hendrik, thanks for this PR! Your workflow of staging approved hunks and reviewing only what's left is a great pattern.

I ended up going a slightly different direction with this. Instead of a CLI flag, I made the default diff type a persistent user setting (configurable in Settings > Display, stored in config.json). I also added a first-run dialog that prompts users to pick their preferred default the first time they open a local review. And I changed the hardcoded default from "uncommitted" to "unstaged" to match `git diff` semantics, which aligns with what you were going for.

You have co-authorship on the commit in #531 since your PR is what kicked this off. Appreciate you taking the time to contribute.

PR #531, opened 2026-04-10
feat(review): configurable default diff type

backnotprop:
## Summary

- Changes the default review diff type from `uncommitted` to `unstaged` (matching `git diff` semantics)
- Makes the default user-configurable via Settings > Display, persisted to `~/.plannotator/config.json`
- Adds a first-run setup dialog on the first local review session so users can choose their preferred default
- Preserves P4 default (`p4-default`) for Perforce users

Inspired by #521 from @hrichert. Rather than a CLI flag, we made the default a persistent setting with a first-run prompt.

Related: #287

## Test plan

- [ ] Run `/plannotator-review` on a fresh install (clear `plannotator-diff-type-setup-done` cookie) — verify the setup dialog appears
- [ ] Select a non-default option, click Done — verify `~/.plannotator/config.json` has `diffOptions.defaultDiffType`
- [ ] Close and re-run `/plannotator-review` — verify the dialog does not reappear and the selected default is used
- [ ] Open Settings > Display — verify "Default Diff View" control appears and reflects the saved value
- [ ] Change the setting in Settings — verify it persists across sessions
- [ ] Test with AI providers configured — verify the diff type dialog waits for the AI setup dialog to complete (no stacking)
- [ ] Test PR review mode — verify the dialog does not appear
- [ ] Verify the diff type switcher dropdown still works to toggle between views during a session

### DIFF
--- a/apps/hook/server/index.ts
+++ b/apps/hook/server/index.ts
@@ -64,6 +64,7 @@ import {
   handleAnnotateServerReady,
 } from "@plannotator/server/annotate";
 import { type DiffType, getVcsContext, runVcsDiff, gitRuntime } from "@plannotator/server/vcs";
+import { loadConfig, resolveDefaultDiffType } from "@plannotator/shared/config";
 import { fetchRef, createWorktree, removeWorktree, ensureObjectAvailable } from "@plannotator/shared/worktree";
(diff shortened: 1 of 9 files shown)


githendrik:
Seems like a reasonable default (same as with git by default actually). Thanks!

Issue #456, opened 2026-04-01
[Feature Request] Ask questions about the plan


andreivladmatei:
Something I realized would make a lot of sense is: once you have a plan, maybe you want to ask Claude questions about it without altering the plan.

E.g. I did a lot of back and forth with comments and global comment and then, just before approving it, I may want to ask: "Can you tell me if this is....?" something that normally you would use `/btw` for 

Can this be implemented?

backnotprop:
I have an `ask` feature in the works, where you don't leave the UI & can talk to a forked session (similar to btw)? Will this suffice?

andreivladmatei:
yep, something like that. A way to clarify the plan by asking questions to cc

davinpkn:
@backnotprop Also, it is possible to after plan, when approving (or a custom button for that) it just write the .md file and don't start implementation? Sometimes i want to switch models for non-critical implementations

backnotprop:
@daviziks for opencode, you can disable agent switching, does this work?

<img width="747" height="491" alt="Image" src="https://github.com/user-attachments/assets/6977efa2-dab9-40c6-ad3b-f26507120cc1" />

davinpkn:
@backnotprop I'm using pi, i just find out i can configure the behaviour on the custom `plannotator.json` config, ty so much for this tool!
But the `ask` tool will be huge!

hossainemruz:
> [@daviziks](https://github.com/daviziks) for opencode, you can disable agent switching, does this work?
[...]

PR #123, opened 2026-01-30
feat: Structured review workflow (tags, MACRO, annotate, markers, evolution)

theflysurfer:
## Summary

This PR adds 5 features that transform Plannotator into a comprehensive document review system for AI-assisted development workflows:

### 1. Review Tags
Methodology-based annotation tags for clear AI feedback:
- **Modification**: `@TODO`, `@FIX`, `@CLARIFY`, `@MISSING`, `@ADD-EXAMPLE`
- **Verification**: `@VERIFY`, `@VERIFY-SOURCES`, `@CHECK-FORMULA`, `@CHECK-LINK`
- **Validation**: `@OK`, `@APPROVED`, `@LOCKED`

### 2. [MACRO] Flag
Toggle to mark annotations with cross-document impact. Signals to AI that related files should be checked.

### 3. Annotate Mode
Annotate any markdown file (not just plans):
```bash
plannotator annotate README.md
/plannotator-annotate CLAUDE.md
```

### 4. Persistent Validation Markers
Save `@OK`/`@APPROVED`/`@LOCKED` tags to source files as HTML comments:
```markdown
<!-- @APPROVED by="alice" date="2026-01-30" -->
## Database Schema
```

### 5. Plan Evolution Tracking
Version comparison with git-diff style split view:
- Auto-saves plan history on each review
- LCS-based block diffing algorithm
- Synchronized scroll between columns

## Files Changed

| Feature | Key Files |
|---------|-----------|
| Review Tags | `packages/core/types.ts`, `packages/ui/components/AnnotationToolbar.tsx` |
| MACRO Flag | `packages/ui/components/AnnotationToolbar.tsx` |
| Annotate Mode | `packages/server/annotate.ts`, `apps/hook/commands/plannotator-annotate.md` |
| Markers | `packages/core/markers.ts` |
| Evolution | `packages/core/planDiff.ts`, `packages/server/planHistory.ts`, `packages/ui/components/PlanEvolutionPanel.tsx` |

## Test Plan

- [x] Review tags appear in toolbar dropdown
- [x] MACRO flag toggles and exports correctly
- [x] Annotate mode opens any markdown file
- [x] Markers save to source file as HTML comments
- [x] Evolution panel shows version comparison
- [x] Scroll sync works bidirectionally

## Screenshots

See individual PR documentation in `docs/` folder for detailed descriptions.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #111, opened 2026-01-24
Feature Request: See what has been changed since last iteration on the plan


PetrBrabec:
I'd like to see highlighted what has been changed since last iteration on the plan so I don't have to go over it all again.

_Originally posted by @PetrBrabec in https://github.com/backnotprop/plannotator/discussions/110_

backnotprop:
Good idea @PetrBrabec , I guess `plan diff` could be the feature

backnotprop:
API exists now, I'm just filling out the UI, and it should be released today.

backnotprop:
This capability is now merged via #176. Plan diff UI with sidebar, dual view modes (rendered + raw), and version history browsing will be included in the next release.

backnotprop:
Plan Diff is here! 🎉

Released in [v0.9.0](https://github.com/backnotprop/plannotator/releases/tag/v0.9.0) — when your agent revises a plan after feedback, you'll see exactly what changed with two view modes (rendered and raw), automatic version history, and a sidebar for browsing past iterations.

Blog post with details: https://plannotator.ai/blog/plan-diff-see-what-changed/

There's also a [YouTube video demo](https://youtu.be/uIWkFCg60Lk) walking through the feature.

Thanks for requesting this!

PR #562, opened 2026-04-14
feat(viewer): show color swatches next to hex color codes in plans

Pran-Ker:
## Summary

When a plan contains a hex color value (`#rgb`, `#rrggbb`, `#rgba`, `#rrggbbaa`), the `InlineMarkdown` renderer now displays a small filled swatch inline — immediately to the left of the monospace hex code — so the actual color is visible at a glance without leaving the document.

This is especially useful for **frontend development plans** where colours are frequently referenced (design tokens, Tailwind overrides, CSS variables, component palette decisions). Without this, reviewers have to mentally decode hex strings or open a colour picker just to follow the author's intent.

**Before:** `use #ff6600 for the CTA button`
**After:** `use 🟧 #ff6600 for the CTA button` *(small inline swatch, not an emoji)*

## What changed

**`packages/ui/components/Viewer.tsx`** — two edits, ~20 lines total:

1. **New pattern block** in the `InlineMarkdown` loop (after inline-code, before wikilinks):
   - Matches `#rgb`, `#rgba`, `#rrggbb`, `#rrggbbaa` with a negative lookahead that prevents false-positives on URL anchors (`#section`), CSS id selectors (`#sidebar`), and any identifier that continues with word characters.
   - Renders a `14×14` rounded swatch (`<span style={{ backgroundColor: hex }}>`) followed by the hex code in a `<code>` tag.

2. **`nextSpecial` regex** now includes `#` so the scanner stops at a `#` mid-sentence rather than swallowing it as plain text.

## Security notes

- **No CSS injection risk.** `style={{ backgroundColor: hex }}` uses React's JS-object style prop, not raw `cssText`. React sets individual CSS properties, so it is not possible to break out into arbitrary CSS via this value.
- **Value is regex-constrained.** The only string that ever reaches `backgroundColor` is `#` followed by exactly 3, 4, 6, or 8 hex digits (`[0-9a-fA-F]`). Payloads like `expression()`, `url()`, or `;color:red` suffixes are structurally impossible to inject.
- **No `innerHTML` or `dangerouslySetInnerHTML`.** All rendering is via React JSX — standard XSS protections apply.
- **No ReDoS risk.** The regex uses fixed-length alternatives checked from longest to shortest with no nested quantifiers.

## Tests

`packages/ui/components/hexColorSwatch.test.ts` — 19 tests, all passing (`bun test`):

| Group | What is tested |
|-------|----------------|
| Valid colors | 3-, 4-, 6-, 8-digit hex; mixed case; word-boundary termination |
| No false positives | 5- and 7-digit sequences, non-hex chars, bare `#`, URL anchors, CSS id names, 9+ digit strings |
| Surrounding context | Color mid-sentence, followed by punctuation, two colors in sequence |
| Security | Injection payloads never reach `backgroundColor`; captured value always passes strict `/^#[0-9a-fA-F]{3,8}$/` check |

```
bun test packages/ui/components/hexColorSwatch.test.ts
 19 pass
  0 fail
```

### DIFF
--- a/packages/ui/components/Viewer.tsx
+++ b/packages/ui/components/Viewer.tsx
@@ -836,6 +836,25 @@ const InlineMarkdown: React.FC<{ text: string; onOpenLinkedDoc?: (path: string) 
       continue;
     }
 
+    // Hex color swatch — 3/4-digit forms need an a-f letter to avoid matching issue refs like #123.
+    match = remaining.match(/^(#(?:[0-9a-fA-F]{8}|[0-9a-fA-F]{6}|(?=[0-9a-fA-F]*[a-fA-F])[0-9a-fA-F]{4}|(?=[0-9a-fA-F]*[a-fA-F])[0-9a-fA-F]{3}))(?![0-9a-fA-F\w])/);
+    if (match) {
+      const hex = match[1];
+      parts.push(
+        <span key={key++} className="inline-flex items-center gap-1 align-middle">
+          <span
+            className="inline-block w-3.5 h-3.5 rounded-sm border border-black/20 dark:border-white/20 flex-shrink-0"
+            style={{ backgroundColor: hex }}
+            title={hex}
+          />
+          <code className="px-1.5 py-0.5 rounded bg-muted text-sm font-mono">{hex}</code>
+        </span>
+      );
+      remaining = remaining.slice(match[0].length);
+      previousChar = match[0][match[0].length - 1] || previousChar;
+      continue;
+    }
+
     // Wikilinks: [[filename]] or [[filename|display text]]
     match = remaining.match(/^\[\[([^\]|]+)(?:\|([^\]]+))?\]\]/);
     if (match) {
@@ -972,7 +991,7 @@ const InlineMarkdown: React.FC<{ text: string; onOpenLinkedDoc?: (path: string) 
     }
 
     // Find next special character or consume one regular character
-    const nextSpecial = remaining.slice(1).search(/[\*_`\[!~\\<]/);
+    const nextSpecial = remaining.slice(1).search(/[\*_`\[!~\\<#]/);
     if (nextSpecial === -1) {
       parts.push(remaining);
       previousChar = remaining[remaining.length - 1] || previousChar;


backnotprop:
Nice feature

packages/ui/components/Viewer.tsx
```diff
@@ -836,6 +836,25 @@ const InlineMarkdown: React.FC<{ text: string; onOpenLinkedDoc?: (path: string)
       continue;
     }
 
+    // Hex color: #rrggbb, #rgb, #rrggbbaa, #rgba — render a color swatch next to the code
+    match = remaining.match(/^(#(?:[0-9a-fA-F]{8}|[0-9a-fA-F]{6}|[0-9a-fA-F]{4}|[0-9a-fA-F]{3}))(?![0-9a-fA-F\w])/);
+    if (match) {
```
backnotprop, line 841: **False positive on issue/PR numbers** every decimal digit (`0–9`) is a valid hex digit, so `#123`, `#456`, `#1234` etc. all match the current regex and render as color swatches. Since development plans routinely cite GitHub issues, this will fire constantly on unintended content.

**Root cause:** the 3- and 4-digit alternatives don't distinguish numeric-only sequences from actual CSS shorthand colors.

**Fix:** add a lookahead requiring at least one `a–f` letter for the short forms. 6- and 8-digit patterns are unambiguous and can stay as-is.

```suggestion
    // Hex color: #rrggbb, #rgb, #rrggbbaa, #rgba ... render a color swatch next to the code.
    // 3- and 4-digit forms require at least one a-f letter to avoid false positives on
    // issue/PR references (#123, #1234) whose digits are all valid hex chars.
    match = remaining.match(/^(#(?:[0-9a-fA-F]{8}|[0-9a-fA-F]{6}|(?=[0-9a-fA-F]*[a-fA-F])[0-9a-fA-F]{4}|(?=[0-9a-fA-F]*[a-fA-F])[0-9a-fA-F]{3}))(?![0-9a-fA-F\w])/);
```

**Tradeoff:** pure-decimal 3-digit grays (`#000`, `#111`…`#999`) no longer trigger a swatch, since they're indistinguishable from issue numbers. `#000000` (6-digit) still works fine.

| Input | Before | After |
|---|---|---|
| `#123` (issue ref) | swatch ❌ | plain text ✅ |
| `#1234` (issue ref) | swatch ❌ | plain text ✅ |
| `#f60` `#fff` `#abc` | swatch ✅ | swatch ✅ |
| `#fade` `#cafe` | swatch ✅ | swatch ✅ |
| `#ff6600` `#000000` | swatch ✅ | swatch ✅ |

The existing test suite should add cases for `#123` and `#1234` (expect no match) to cover this.

Pran-Ker:
Add the change @backnotprop applied the more accurate regex. 

The 3- and 4-digit alternatives now require at least one `a–f` letter via a lookahead, so `#123` and `#1234` pass through as plain text. 6- and 8-digit forms are unchanged.

backnotprop:
@Pran-Ker

Thanks for [#562](https://github.com/backnotprop/plannotator/pull/562).

PR #563, opened 2026-04-14
fix: arrival-save plans to ~/.plannotator/plans/ on server startup

### LINKED ISSUE #556: No more plan file created in ~/.planotator/plan if I don't prove/deny
Recentily I found no more plan created in ~/.planotator/plan  if I don't prove/deny.

I usally need use opus(one key) to create plan, them use another sonnet key to execute plan,  I write a claude command to read last plannotator plan and execute , but I find I had to approve/deny to generate the plan file.  
And once I quit the askquestion prompt in claude,  the plannotator page doesn't working any more ,  save to obsidian also fail 

On windows 10

backnotprop:
Closes #556.

## Summary

- Auto-save the plan to `~/.plannotator/plans/{slug}.md` on server startup — before the browser opens — so users who quit Claude Code without clicking approve/deny still find the plan in the directory they expect.
- Migrate `planSave` preferences (enabled flag + custom path) from browser cookies to `~/.plannotator/config.json` so the server can honor them at arrival-save time. Migration is lazy, one-shot, and clears cookies only after the server confirms the write.
- Preserve full precedence: `PLANNOTATOR_PLAN_SAVE` / `PLANNOTATOR_PLAN_SAVE_ON_ARRIVAL` env vars > approve/deny body session override > `config.json` > defaults. Env vars now authoritatively win over the client body (previously a client default payload could silently re-enable saves an operator disabled at the environment level).
- Validate `customPath` at the `/api/config` boundary (rejects `..` segments) as defense-in-depth.

## Backwards compatibility

Existing users keep their preferences — the first UI load after upgrade reads legacy cookies, POSTs them to `/api/config`, and clears the cookies on success. From then on, `config.json` is authoritative.

One-session transient window is accepted and documented inline in `planSave.ts`: for a user who had saves disabled via cookies, the first post-upgrade session can produce harmless arrival + decision files in the default directory before migration fires. Same blast radius as the arrival-save trade-off itself — self-corrects next session.

## Test plan

- [ ] Fresh user (no cookies, no config.json): submit a plan → `~/.plannotator/plans/{slug}.md` exists immediately; approve → `{slug}-approved.md` lands alongside.
- [ ] Legacy user with `plannotator-save-path=/tmp/myplans` cookie: open UI → `config.json` gains `planSave.customPath: "/tmp/myplans"`, cookies cleared. Next plan's arrival save lands in `/tmp/myplans/`.
- [ ] Legacy user with `plannotator-save-enabled=false`: after UI loads, `config.json` has `enabled: false`; next session's arrival save is skipped.
- [ ] Settings UI: toggle enable/disable + change custom path → `config.json` reflects change, no cookies written; approve immediately after → snapshot uses the new values.
- [ ] Env var: `PLANNOTATOR_PLAN_SAVE_ON_ARRIVAL=false plannotator ...` → no arrival save, regardless of config.
- [ ] Env var: `PLANNOTATOR_PLAN_SAVE=false` → no approve/deny snapshot even if client body sends `enabled: true`.
- [ ] Path-traversal guard: `POST /api/config` with `planSave.customPath: "../../etc"` returns 400 and leaves config.json unchanged.
- [ ] Archive browser: open `plannotator archive` with `customPath` set in config.json → list and first-plan content both load from the custom directory, not defaults.
- [ ] Pi server parity: `bun run build:pi`, submit plan → same behavior as Bun server.
- [ ] Regression: full approve/deny flow with annotations — `{slug}-approved.md` content unchanged.

For provenance purposes, this PR was AI assisted.


backnotprop:
Closing this 92-day-old conflicted implementation as part of backlog cleanup. The linked issue is also being closed; please reopen it with a current reproduction if the behavior is still a problem.

PR #40, opened 2026-01-07
Add configurable agent switching for OpenCode

backnotprop:
## Summary
- Add Settings UI option to configure agent switching behavior after plan approval
- Users can choose "Build" (default) or "Disabled"
- Useful for oh-my-opencode users who have their own agent switching logic
- Add configurable plan saving with custom path support

## Changes
- `packages/ui/utils/agentSwitch.ts` - New utility for storing preference
- `packages/ui/utils/planSave.ts` - New utility for plan save settings
- `packages/ui/components/Settings.tsx` - Added "Agent Switching" and "Save Plans" settings
- `packages/editor/App.tsx` - Pass settings to API endpoints
- `packages/server/index.ts` - Accept and pass through settings
- `packages/server/storage.ts` - Support custom save paths
- `apps/opencode-plugin/index.ts` - Conditionally switch agent based on setting

## Test plan
- [x] Open Settings, verify "Agent Switching" dropdown appears (OpenCode only)
- [x] Set to "Disabled", approve plan, verify no agent switch occurs
- [x] Set to "Build" (default), approve plan, verify agent switches to build
- [x] Disable "Save Plans", verify no files saved on approve/deny
- [x] Set custom path, verify plans save to custom location

Closes #37
Closes #36

Co-Authored-By: TheOutdoorProgrammer <joey@apollorion.com>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #564, opened 2026-04-14
don't change current agent

andreineculau:
agent_cycle is not the correct command as the user may have several agents defined, not only build and plan.

I do think that the changes introduced in #40 make sense, that is "not only run the plan with the build agent, but also switch the agent to build" - but in its current format, it is simply buggy as it assumes only a build and plan agent.

I checked the opencode sourcecode and I'm not sure what is the alternative. Cycle and validate that the current agent is the target agent, if not repeat?!

### DIFF
--- a/apps/opencode-plugin/index.ts
+++ b/apps/opencode-plugin/index.ts
@@ -465,14 +465,6 @@ Do NOT proceed with implementation until your plan is approved.`);
 
             if (shouldSwitchAgent) {
               try {
-                await ctx.client.tui.executeCommand({
-                  body: { command: "agent_cycle" },
-                });
-              } catch {
-                // Silently fail
-              }
-
-              try {
                 await ctx.client.session.prompt({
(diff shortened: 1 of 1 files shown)


backnotprop:
There is a setting in the UI, does that not work for you?

You can disable agent switching

andreineculau:
@backnotprop it's 2 different things. I want that option enabled, so that an approval triggers the build agent to implement the plan. Disabling the option disables the agent cycle but also the automatic implementation in
https://github.com/andreineculau/plannotator/blob/f893a9d9874d4738bd1938b5c01354d0cf9c1438/apps/opencode-plugin/index.ts#L467-L475

backnotprop:
I understand. That's an oversight on my part.

backnotprop:
@andreineculau

Thanks for [#564](https://github.com/backnotprop/plannotator/pull/564).

Issue #479, opened 2026-04-03
OpenCode: add manual-review mode and /plannotator-last-plan to avoid eager submit_plan


backnotprop:
## Summary

OpenCode support is currently mixing two different product models:

1. **Plannotator owns the planning loop** — the agent is expected to call `submit_plan` automatically.
2. **Plannotator is an explicit review tool** — users want to call it only when they decide the plan is worth reviewing.

A growing cluster of feedback says the current default is too eager and too broad.

## Pain points

Common sentiment:

- "I really enjoy Plannotator when I want to call it, but invoking itself on the OpenCode Plan mode is a bit annoying."
- "The agent eagerly calls it."
- "I don't want the build agent to call it."
- "I'd love a way if I could only manually invoke Plannotator on the last plan submitted by the agent."

What this looks like in practice:

- `submit_plan` is still available to all **primary** agents, not just the planning agent.
- The `build` agent is excluded from extra prompt injection, but it can still access the tool.
- Post-approval behavior is still biased toward implementation handoff, which clashes with users who want to stop, stay in planning, or manually choose the next step.
- The plugin still assumes the planning agent is literally named `plan`, which makes alternate workflows harder.
- Users who like OpenCode's native planning flow feel like Plannotator is hijacking it instead of layering on top of it.

## Why this matters

This is not just prompt tuning.

The complaints around eager invocation, build-agent access, auto-switching, and hardcoded plan-agent behavior all point to the same underlying issue: **the OpenCode contract is ambiguous**.

Right now the plugin is trying to serve both:

- users who want automatic plan interception
- users who want manual, on-demand review

Those should probably not share one default behavior.

## Existing issues this seems to unify

- #159 - option for ending planning without automatically implementing
- #118 - stop/hold behavior after approval, especially in OpenCode web/desktop
- #139 - oh-my-opencode compatibility around agent switching / workflow assumptions
- #365 - `submit_plan` still feels too prompt-dependent / ambiguous
- #422 - hardcoded reliance on the agent name `plan`

I also think this should be resolved before expanding agent-facing OpenCode power further, e.g. #295.

## Proposal

Introduce an explicit OpenCode workflow mode instead of relying on one global default.

### 1. `manual-review` (recommended default)

Behavior:

- No prompt injection into agent planning.
- No `submit_plan` tool exposed to agents.
- User lets OpenCode plan normally.
[...]

Issue #365, opened 2026-03-22
submit_plan: replace dual-mode auto-detect with a single deterministic path


rcdailey:
I've been following (and commenting on) the plan mode changes since v0.13.1. My feedback in #328 covers the full arc, but the core problem hasn't been addressed so I think it's worth a dedicated issue.

## Problem

`submit_plan` accepts both inline markdown and file paths, with auto-detection choosing the mode based on what the agent passes. This means the plugin's behavior depends on prompt engineering successfully steering the agent toward the right mode.

In practice, this has been fragile:

- v0.13.1 introduced file-based plans
- v0.14.0 hit permission conflicts with file writing
- v0.14.1 collapsed two tools into one with auto-detect
- v0.14.2 replaced the plan mode prompt and the agent reverted to inline markdown (my last comment on #328, which went unanswered)
- v0.14.3 and v0.14.4 shipped without addressing this

Four releases in five days, most dealing with prompt/permission interactions around the dual-mode design. The agent choosing "wrong" shouldn't be a failure mode.

## Why it matters

The whole point of file-based plans is token efficiency. Repeating the entire plan as inline markdown on every feedback cycle is wasteful. But the current architecture only delivers that benefit when the agent happens to choose file mode, which is prompt-dependent and (as of v0.14.2) not working.

I also think offering both modes makes the tool harder for agents to use correctly. A single clear path removes the prompt engineering burden entirely.

## Ideas

I don't know what constraints you're working with, so here are a few approaches to consider rather than a single prescription.

**Option A: Plugin manages the file internally.** The agent always passes inline markdown. The plugin writes it to a `.md` file behind the scenes (`.opencode/plans/` follows OC's documented convention). On denial, the feedback tells the agent "your plan is at /path/to/plan.md, edit it and resubmit." If the agent follows that hint, it uses `edit` for targeted changes (token-efficient). If it ignores the hint and passes full markdown again, the plugin just overwrites the file. The file becomes an optimization the plugin enables but doesn't strictly require.

**Option B: Tool only accepts a file path.** The agent must write a `.md` file and pass the path. No inline markdown accepted. This is the simplest contract but depends on the agent reliably writing files first, which might have its own prompt challenges in plan mode given OC's edit restrictions.

These are just starting points. Maybe there's a cleaner approach I'm not seeing. The important thing (from my perspective) is that there's one deterministic path, not two modes the agent has to choose between.

backnotprop:
[...]

Issue #295, opened 2026-03-15
[FEATURE] submit_review tool for Opencode extension


feiwhang:
A single slash command `/plannotator-review` limit the potential usefulness of review UI. 

- Needs to be triggered manually, requires human intervention in the loop.
- Cannot submit committed changes
- Agent should be the one submitting the review not human.

backnotprop:
Thanks for the report. I am consolidating the older backlog so related work has one current tracking issue.

This specific request is now tracked in #905 as part of the programmatic review input surface work. During the June 13, 2026 older-than-two-month triage, I copied the relevant requirement from this issue into #905 and explained why it belongs with the related work.

I am closing this older issue so the backlog points at the current tracking issue. If there is a newer reproduction or a detail that is missing from #905, please comment there.

Issue #159, opened 2026-02-17
[FEATURE REQUEST] Option for End planning without implementing


gustavocaiano:
Im Using version 0.8.1 in Opencode

When plannotator pops up , we can either approve which starts automatically the build implementation, or annotate and send feedback that will cycle back to the same 2 options.

I think a simple feature would be adding a "Implement" button. 

Approve button behaviour would be to end planning (normal behaviour of opencode without plannotator)

emanuelsan:
Agree with @gustavocaiano 
I'm also encountering this issue where I always respond with the "Provide feedback" even if I want to approve the plan, because the "Approve" button sends the prompt "Proceed to implementation" back to opencode. This confuses the agent that submitted the plan for review because I don't want it to be the one doing the actual code implementation.

### Workarounds Tried
I tried using the current global comment functionality, but it assembles a longer response like "Plan approved with notes! [...] Proceed with implementation, incorporating these notes where applicable". This again causes confusion to the agent receiving this because it's not the agent that need to do the implementation.

### Suggested Feature Approach
From my perspective, an ideal approach would be to allow to set a "global verdict" or "global comment" that plays nicely with the "Approve" "Provide feedback" buttons. 

### Notes
I'm using Opencode 1.2.26 and Plannotator 0.13.1

arden-shackelford-q2:
I'd definitely agree that something other than just "Approve, so implement" would be great here.

My opinion here would be that there's an option for "Approve" but its a dropdown where things like "Save plan to file" is an option. That way you can come back to it later, as the plan is saved locally (configurable default path per project).

pencoyd:
Really appreciate plannotator and using it with Claude Code. I'm +1 on having a way to approve a plan without triggering implementation — sometimes I want to review and hold.

PR #571, opened 2026-04-16
Scope OpenCode submit_plan to planning agents

### LINKED ISSUE #422: opencode plugin: remove hardcoded reliance on agent name "plan", to allow user to connect other agents to plannotator
I am happy to code this, this issue is a request for comment. 
 

User story:

Users wants an alternative plan agent with its own name, allowing  the out of the box OpenCode behaviour untouched. 

however, in
apps / opencode-plugin / index.ts

the "plan" name is hard-coded. 

So alternate plan agents can't work with planotator.

A minimal change would be to provide an env var that gives an alternate name (I have not tested this yet, but llm analysis makes it look like an easy change)

Allowing a set of names would be a bit more work since the current code assumes there is only one candidate agent acting as "plan". A simple alternate name configuration would on its bring bring benefits, so I suggest a simple approach first.

backnotprop:
## Summary
- add explicit OpenCode workflow modes: `plan-agent` default, `manual` commands-only, and `all-agents` legacy behavior
- make `plan-agent` the new default for `0.19.1`: keep `submit_plan` available to OpenCode's built-in `plan` agent, allow any extra agents listed in `planningAgents`, and hide or deny it for `build` and other non-planning primary agents
- scope prompt injection, OpenCode permission mutation, and runtime execution checks around that workflow behavior
- keep manual features first-class across modes: `/plannotator-last`, `/plannotator-annotate`, `/plannotator-review`, and `/plannotator-archive`
- add OpenCode folder annotation support and fix folder draft collisions so drafts are isolated per folder instead of leaking between folders
- document the `0.19.1` migration, config snippets, troubleshooting, local sandbox flow, and manual test plan

## Issues
Closes #422.

Refs #479. This PR implements the workflow split, the `plan-agent` default, and narrower default tool exposure, but `/plannotator-last-plan` and approval stay/stop semantics remain follow-up work.

Refs #118, #139, #159, #365, #295.

## Testing
- `bun test apps/opencode-plugin`
- `bun test apps/opencode-plugin/workflow.test.ts`
- `bun run build:opencode`
- `bun run build:pi`
- `bun run build:marketing`


oorestisime (approved):
Looking good. havent tested locally though. 

As a side note: I am not sure why the annotate command is limited to just md and html extensions. Why would someone not be able to annotate and share even specific source code files?

backnotprop:
> Looking good. havent tested locally though. 
> 
> As a side note: I am not sure why the annotate command is limited to just md and html extensions. Why would someone not be able to annotate and share even specific source code files? 

We will get there!

PR #572, opened 2026-04-16
Add ~ support for user-entered file paths

AlexanderKolberg:
This change adds support for `~` in user-entered file paths by expanding home-relative paths in the shared path resolver instead of treating the tilde literally. That shared resolver is now used by the annotate entrypoints and the Bun and Pi reference handlers, so file, folder, vault, and linked-document paths all handle `~` the same way.


backnotprop:
@AlexanderKolberg

Thanks for [#572](https://github.com/backnotprop/plannotator/pull/572).

PR #573, opened 2026-04-16
Fix for RSS feed, for the blog

dotemacs:
An easy one: RSS link for the blog added.

Advertise the blog feed in the shared marketing layout so feed readers and browsers can discover it automatically.


backnotprop:
Thanks!

backnotprop:
@dotemacs

Thanks for [#573](https://github.com/backnotprop/plannotator/pull/573).

PR #577, opened 2026-04-17
Add Codex Stop-hook plan review

ivanov17andrey:
## Summary
- add Codex `Stop` hook plan review support that extracts the latest plan from rollout transcripts, opens Plannotator for fresh plans, and returns continuation feedback on denial
- cover the new Codex parsing and `Stop` flow with server tests and update Codex-facing installation and plan-review docs
- add a reproducible local Codex E2E harness plus checked-in screenshots showing the initial review, revised review, versions sidebar, diff view, and archive sidebar
- ignore the local `.playwright-cli/` cache so Playwright-driven validation does not dirty the repo

## Test plan
- [x] `/bin/bun test apps/hook/server/*.test.ts`
- [x] `/bin/bun run --cwd apps/review build`
- [x] `/bin/bun run --cwd apps/hook build`
- [x] `PLANNOTATOR_BROWSER=/usr/bin/true ./tests/manual/local/test-codex-plan-review-e2e.sh --keep --skip-build --root-dir /tmp/plannotator-codex-stop-e2e-rerun3` with Playwright-driven review, denial, revision, versions, diff, archive, and approval flow

### Initial plan review
Initial plan review

### Feedback before denial
Feedback comment

### Revised plan review
Revised plan review

### Versions sidebar after denial
Versions sidebar

### Diff view against v1
Diff view

### Archive sidebar
Archive sidebar


backnotprop:
Big one!

Give me some time to review but will priortize

ivanov17andrey:
Hey! Any luck to be checked soon?

backnotprop:
Testing now

backnotprop:
Quick question on the Stop hook output format.

On approve, the hook outputs `{}` and exits 0, which means the turn just completes normally. But nothing triggers the transition to Default mode or submits "Implement the plan" like the TUI does. So approval is effectively a no-op from Codex's perspective, right? The user would still need to manually switch modes and kick off implementation.

ivanov17andrey:
> Quick question on the Stop hook output format.
> 
> On approve, the hook outputs `{}` and exits 0, which means the turn just completes normally. But nothing triggers the transition to Default mode or submits "Implement the plan" like the TUI does. So approval is effectively a no-op from Codex's perspective, right? The user would still need to manually switch modes and kick off implementation.

Short answer: **yes for the hook protocol / `codex exec`, but not for the interactive TUI/app**.

**Why approval is a no-op at the hook layer**
- `plannotator/apps/hook/server/index.ts:1019-1020` prints `{}` on approve.
- In Codex, the `Stop` hook output has no special “approved” field. The universal schema is just `continue`, `stopReason`, `suppressOutput`, and `systemMessage`, with `continue` defaulting to `true`: `codex/codex-rs/hooks/src/schema.rs:57-65` and `codex/codex-rs/hooks/src/schema.rs:513-514`.
- `parse_stop()` only derives `should_block` from `decision:"block"`; `{}` therefore means “continue processing, don’t block, don’t stop”: `codex/codex-rs/hooks/src/engine/output_parser.rs:176-193`.
- The core turn loop only has special handling for `should_block` and `should_stop`. If neither is set, it just falls through and completes the turn normally: `codex/codex-rs/core/src/codex/turn.rs:540-567`.

So if you mean **the Stop hook itself**, then yes: approval is just “let the turn finish.” It does **not** itself switch to Default mode or inject `Implement the plan.`

**Why that is not the whole story in the TUI/app**
- After a task completes, the TUI calls `maybe_prompt_plan_implementation()`: `codex/codex-rs/tui/src/chatwidget.rs:2430-2432`.
- That method checks that the current mode is `Plan` and that a plan item was seen this turn, then opens the implementation prompt: `codex/codex-rs/tui/src/chatwidget.rs:2448-2472`.
- The TUI defines `PLAN_IMPLEMENTATION_CODING_MESSAGE` as `"Implement the plan."` and, on “Yes,” submits it in **Default** mode: `codex/codex-rs/tui/src/chatwidget.rs:248-251` and `codex/codex-rs/tui/src/chatwidget.rs:2475-2484`.
- There is an explicit test for this behavior: `codex/codex-rs/tui/src/chatwidget/tests/plan_mode.rs:30-39` asserts the submitted text is `Implement the plan.` and the collaboration mode is `ModeKind::Default`.

**Conclusion**
- In **`codex exec`**: your statement is correct. Approval is effectively just normal completion, so the user would need to manually start implementation.
- In the **interactive TUI/app**: not quite. The hook approval itself is still a no-op, but once the approved plan turn completes, the TUI’s normal post-plan flow should offer “Implement this plan?” and submit `Implement the plan.` in Default mode if the user accepts.

backnotprop:
I'm trying one thing with the app server. If it doesn't work, we'll merge this approach.

backnotprop:
Releasing this today

backnotprop:
@ivanov17andrey

Thanks for [#577](https://github.com/backnotprop/plannotator/pull/577).

PR #594, opened 2026-04-20
fix(pi): loosen plan-mode write gate to directory scope

backnotprop:
## Summary

- Plan-mode `write`/`edit` gate did an exact-path comparison against `planFilePath`, so `/plannotator plans/` blocked *every* file inside `plans/` and the agent fell back to `bash cat > file` to bypass it.
- Gate is now directory-scoped when `planFilePath` ends with a separator, has no file extension, or points to a subdirectory. Default `PLAN.md` behavior is unchanged.
- Extracted the logic to a pure function (`isPlanWritePathAllowed`) in `tool-scope.ts` with 6 new unit tests.

## Repro that now works

```
$ /plannotator plans/
$ "make a simple plan for snake game"
→ agent writes plans/snake_game_plan.md (allowed), no bash fallback
```

## Test plan

- [x] `bun test apps/pi-extension/tool-scope.test.ts` — 9 pass, 0 fail
- [x] `tsc --noEmit -p apps/pi-extension/tsconfig.json` — clean
- [ ] Manual: `/plannotator plans/` in a fresh project, confirm plan writes succeed
- [ ] Manual: `/plannotator PLAN.md` (default), confirm behavior unchanged and writes to `src/` still blocked
- [ ] Manual: `/plannotator plans/auth.md`, confirm sibling writes (`plans/notes.md`) allowed, writes outside blocked

### DIFF
--- a/apps/pi-extension/index.ts
+++ b/apps/pi-extension/index.ts
@@ -53,6 +53,7 @@ import {
 } from "./plannotator-events.js";
 import {
 	getToolsForPhase,
+	isPlanWritePathAllowed,
 	PLAN_SUBMIT_TOOL,
 	type Phase,
 	stripPlanningOnlyTools,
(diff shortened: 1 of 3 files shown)


backnotprop:
Superseded by #595, which redesigns plan mode around agent-supplied file paths instead of the configured-path gate. That approach avoids both the sibling-unlock regression this PR introduced and the pre-existing submit-on-directory failure.

Issue #560, opened 2026-04-14
Word level diff / Diff display options for Plan version comparison


pbowyer:
I see this has been done for review mode in https://github.com/backnotprop/plannotator/pull/428 but it'd be great to have in Plan mode.

# The Problem
When I want to see the changes Claude has made to a plan, it would be helpful to see the inserted/deleted words, not the entire block that has changed.

Examples:

<img width="754" height="328" alt="Image" src="https://github.com/user-attachments/assets/ff743ae5-6a90-45fd-b17c-4a849c731305" />

The one below in particular takes a bit of reading before you find the difference:

<img width="747" height="297" alt="Image" src="https://github.com/user-attachments/assets/b6ad6e75-02fd-420f-95db-ee326c3a9688" />

backnotprop:
@pbowyer is this legible to you, or would you still prefer stacked -/+ with just highlights

<img width="797" height="317" alt="Image" src="https://github.com/user-attachments/assets/ebc46db3-8423-4524-a29d-a590c7cb0d30" />

pbowyer:
@backnotprop Thanks, that's legible to me. Shame about the splitting of `early-access` and `production-grade` but that's the component you're using and the choices it's made over what counts as a 'word'.

PR #603, opened 2026-04-23
feat(ui): quieter plan diffs on prose edits

pbowyer:
I appreciate you adding #565 for #560 which has made finding changes easier in my plans - but I wanted to make reading my diffs even easier 😀 

Plan diffs use `diffWordsWithSpace`, which breaks on word boundaries. For most edits that's the right level of granularity, but it struggles with three specific shapes: markdown emphasis delimiters, adjacent word-level swaps, and hyphenated compounds. The results are technically correct but hard to read at a glance. This PR layers three passes on top of the existing engine to fix them.

## The three cases

**Bold phrase swaps (known issue ⑯).** For `**preliminary analysis**` → `**final analysis**`, the opening `**` attaches to the first word-run while the closing `**` slides into the unchanged tail. The result has no balanced `**…**` pair outside the `<del>`/`<ins>` wrappers, so `InlineMarkdown`'s bold regex misses it and the asterisks render literally.

**Adjacent word swaps.** When a rewritten sentence changes several nearby words, each swap is highlighted independently. Output alternates red and green between thin spaces — accurate, but the eye can't pick out the shape of the edit.

**Hyphenated compounds.** `ninety-five` → `ninety-nine` diffs as an unchanged `ninety-` prefix and a swapped `five`/`nine` suffix. The hyphen ends up on the wrong side of the change and the compound reads as two separate tokens.

## What changed

- **`9e2eace` — balanced emphasis pairs.** Before diffing, replace each balanced `**…**`, `__…__`, `~~…~~`, `*…*`, `_…_` (and the triple forms) with a unique word-char sentinel. Restore after. Same pattern used by the existing code-span and link passes. Pair matching follows CommonMark's flanking rules so stray asterisks like `2**3` aren't captured.
- **`f455ac6` — coalesce adjacent change sites.** After the diff returns, walk the token stream and merge any run of two or more change sites separated only by "thin" unchanged tokens (whitespace, `, . ; : — – " '`). Isolated swaps pass through, which keeps word-level highlighting for the common case. Parens and brackets remain hard boundaries so inline links aren't absorbed into coalesced runs.
- **`b004737` — hyphen atomization.** A final sentinel pass replaces infix hyphens between word chars with a fixed marker. Leading and trailing dashes, and em-dash-like separators (a dash with whitespace on one side), are left alone.

### The clearest demonstration is case ③ — a paragraph with many adjacent word-level changes. Before, the output interleaves red and green at the word level. After, each side reads as one coherent phrase:

before-03-inline-code
after-03-inline-code

### Case ① — scattered edits, including the two-word "load balancer" → "service mesh" swap and the `ninety-five` → `ninety-nine` compound:

before-01-scattered-edits
after-01-scattered-edits

### Case ② — bold phrases (a quieter improvement; included for completeness):

before-02-bold-phrases
after-02-bold-phrases

### Case ⑯ — the previously-known bold-phrase limitation, now resolved. Before: literal `**` asterisks leak out and `analysis` loses its weight. After: a clean bold-struck → bold-green swap:

before-16-bold-phrase
after-16-bold-phrase

## Approaches considered an discarded

Rather than inventing my own I looked at existing algorithms first and Google's diff-match-patch came up as an alternative: a 45kb dependency using Myers diff with semantic cleanup, and well-tested. I considered it, but it's character-level by default, which would fragment `ninety-five` → `ninety-nine` more than the current approach does, plus the library's documented word-mode is a DIY tokeniser, so every token-shape decision in this PR would still need to be made after the swap.

## Test plan

All tested locally and working.

- [ ] `bun test packages/ui/utils/planDiffEngine.test.ts` passes. Covers each atomization pass, the coalescing rules, and anti-regressions for isolated swaps and hard boundaries.
- [ ] The ⑯ demo at `packages/editor/demoPlanDiffDemo.ts:244` renders as a single bold-struck → bold-green swap.
- [ ] A paragraph with three adjacent word swaps coalesces into one phrase-level change.
- [ ] An isolated single-word swap still highlights at word level.
- [ ] A swap next to an unchanged inline link splits cleanly at the link boundary.
- [ ] Annotation capture in `PlanCleanDiffView` continues to work — block-level, so it should be unaffected, but worth a manual check.

### DIFF
--- a/apps/hook/dev-mock-api.ts
+++ b/apps/hook/dev-mock-api.ts
@@ -483,9 +483,9 @@ module.exports = { authenticate };
 
 ---
 
-## ⑯ Known Limitation — Word Swap Inside a Multi-Word Bold Phrase
+## ⑯ Fixed — Word Swap Inside a Multi-Word Bold Phrase
 
-> **What to watch for (this is a known glitch):** When a single word inside a multi-word bold phrase changes — like **preliminary analysis** becoming **final analysis** — the engine splits the bold markers across the change boundary. You will likely see raw \`**\` asterisks rendered as literal text and the word "analysis" lose its bold styling. This is a boundary case we haven't fixed yet; it was surfaced by an adversarial audit of the engine.
+> **What to watch for:** When a single word inside a multi-word bold phrase changes — like **preliminary analysis** becoming **final analysis** — the diff engine atomizes the whole balanced \`**…**\` pair so it renders as a clean old-bold-struck + new-bold-green swap. Previously the closing \`**\` orphaned into unchanged-tail text and rendered as a literal asterisk; that limitation is resolved.
 
 Before the leadership steering committee signs off on the external rollout phase, the team must complete a full pass over the **preliminary analysis** of load testing results, confirm that the error budget still permits the planned migration window, and escalate any unresolved dependencies to the program lead. Any open question at this stage must be either resolved or formally deferred to the post-launch review with named owners and dates.
 
--- a/packages/editor/demoPlanDiffDemo.ts
+++ b/packages/editor/demoPlanDiffDemo.ts
@@ -239,9 +239,9 @@ export async function authenticate(request: Request): Promise<AuthContext | null
 
 ---
 
-## ⑯ Known Limitation — Word Swap Inside a Multi-Word Bold Phrase
+## ⑯ Fixed — Word Swap Inside a Multi-Word Bold Phrase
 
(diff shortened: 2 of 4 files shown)


backnotprop:
Ok great! Yea I wasnt totally sure about the original approach, yours is looking better, Ill get a review in

backnotprop:
nice work

backnotprop:
@pbowyer

Thanks for [#603](https://github.com/backnotprop/plannotator/pull/603).

PR #605, opened 2026-04-23
fix(ui): in-page anchor navigation for headings

dgrissen2:
## Summary
- Intercept `#`-prefixed link clicks in `InlineMarkdown` and smooth-scroll
  within the sticky scroll viewport. The native browser jump didn't work
  because plan content lives inside a scrollable container, not the
  document root.
- Expose both a canonical slug and a legacy slug per heading via
  `data-anchor-aliases` so previously shared URLs keep resolving. Canonical
  collapses runs of separators; legacy preserves them.
- Handle `hashchange` and the initial hash so direct links land on the
  right heading after load.
- New `packages/ui/utils/anchors.ts` with `normalizeAnchorText`,
  `decodeAnchorHash`, `slugifyHeadingAnchor`, `legacySlugifyHeadingAnchor`,
  `getHeadingAnchorAliases` + unit tests.
- Thread `onNavigateAnchor` through `BlockRenderer`, `AlertBlock`,
  `Callout`, `TableBlock`, `TablePopout`, `proseBody`.
- Ignore `.serena/` tooling state.

## Test plan
- [ ] `bun test packages/ui/utils/anchors.test.ts`
- [ ] Render a plan with `## Part I: Overview: Scope` and click a TOC link
      like `[Scope](#part-i-overview-scope)` — scrolls smoothly, sticky
      header isn't overlapping the target.
- [ ] Open a shared URL that uses the legacy double-hyphen slug
      (`#part-i--overview--scope`) — still resolves via alias.
- [ ] Load a page with `#some-heading` in the URL directly — auto-scrolls
      on first render.
- [ ] Anchor links inside tables, callouts, alerts, and blockquotes all
      navigate in-page.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,7 @@ apps/pi-extension/review-core.ts
 # Claude Code session-local runtime state (lock files, scheduled-task state).
 # Machine-specific; never belongs in the repo.
 .claude/
+.serena/
 *.ntvs*
 *.njsproj
 *.sln
--- a/packages/ui/components/BlockRenderer.tsx
+++ b/packages/ui/components/BlockRenderer.tsx
@@ -119,7 +119,7 @@ export const BlockRenderer: React.FC<{
       return <hr className="border-border/30 my-8" data-block-id={block.id} />;
 
     case 'html':
-      return <HtmlBlock block={block} imageBaseDir={imageBaseDir} onOpenLinkedDoc={onOpenLinkedDoc} />;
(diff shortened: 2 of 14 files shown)


backnotprop:
what's the legacy slug alias for? heading links never worked before this PR so we shouldn't have old URLs to preserve.

dgrissen2:
Sloppy coding, good call out.  Will work on addressing and removing some complexity.

PR #608, opened 2026-04-23
fix(commands): use background Bash to avoid shell timeout killing servers

akdoan1:
## Summary

- Replace `!` shell preprocessing with `run_in_background: true` Bash instructions in review, annotate, and annotate-last slash commands
- Fixes servers being killed after ~2 minutes due to Claude Code's shell preprocessing timeout
- Prevents annotation loss when the server dies mid-review

## Problem

The `/plannotator-review`, `/plannotator-annotate`, and `/plannotator-last` slash commands use the `!` preprocessing syntax, which runs `plannotator` as a blocking shell command. The server blocks on `waitForDecision()` until the user submits, but Claude Code's shell preprocessing has a ~120s timeout that kills the process during long sessions.

Plan mode doesn't have this issue because it uses a PermissionRequest hook with a 96-hour timeout (`hooks.json`).

When the process is killed:
1. The draft file survives on disk, but only contains annotations up to the last successful auto-save
2. Auto-save failures are silent (`.catch(() => {})`), so the user has no indication their work isn't being saved
3. Reopening recovers partial annotations but anything after the last save is lost

## Fix

Replace `!`plannotator review $ARGUMENTS`` with skill instructions that tell Claude to run the command via the Bash tool with `run_in_background: true`. Background Bash tasks have no timeout and notify Claude when complete.

No server code changes, no build required — only the 3 `.md` skill files are modified.

## Test results

Tested with `run_in_background: true` Bash tasks (the approach this PR uses). Both servers survived 135 seconds (past the old 120s kill threshold) and successfully captured annotations on submission:

| Test | Survived 135s | Annotations submitted | Feedback captured |
|------|:---:|:---:|:---:|
| `plannotator review` | Yes | Yes | Yes — review feedback on diff |
| `plannotator annotate` | Yes | Yes | Yes — annotation on markdown |

### Review output (after 135s wait):
```
# Code Review Feedback

## test.txt

### Line 1 (new) — ``hello world`` (chars 0-11)
test

The reviewer has identified issues above. You must address all of them.
```

### Annotate output (after 135s wait):
```
# File Feedback

I've reviewed this file and have 1 piece of feedback:

## 1. Feedback on: "Test file"
> test
```

## Environment

- Claude Code v2.1.108
- Plannotator v0.19.0
- macOS 26.4

### DIFF
--- a/apps/hook/commands/plannotator-annotate.md
+++ b/apps/hook/commands/plannotator-annotate.md
@@ -4,9 +4,11 @@ allowed-tools: Bash(plannotator:*)
 disable-model-invocation: true
 ---
 
-## Markdown Annotations
+## Instructions
 
-!`plannotator annotate $ARGUMENTS`
+Run `plannotator annotate $ARGUMENTS` using the Bash tool with `run_in_background: true`. This starts an annotation server and opens the browser. The user will annotate the document and submit feedback — this may take a long time.
+
+**Wait for the background task to complete before proceeding.** Do not interrupt or take other actions while the user is annotating.
 
 ## Your task
 
--- a/apps/hook/commands/plannotator-last.md
+++ b/apps/hook/commands/plannotator-last.md
@@ -4,9 +4,11 @@ allowed-tools: Bash(plannotator:*)
 disable-model-invocation: true
 ---
 
-## Message Annotations
+## Instructions
 
-!`plannotator annotate-last $ARGUMENTS`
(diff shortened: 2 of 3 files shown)


backnotprop:
Hey, I'm happy to review this. I'll give it a shot. I haven't actually run into this problem before. 

Usually it gets auto background. 

> Fixes servers being killed after ~2 minutes due to Claude Code's shell preprocessing timeout

And now the slash command would force the agent to call it, where before it was deterministic and happened automatically.

backnotprop:
Closing this older, conflicted implementation because the command files it modifies no longer exist after the move to installed skills. No replacement issue is being opened; a fresh report can be filed if the current skill-based flow reproduces the timeout.

PR #621, opened 2026-04-28
Add source line numbers to exported feedback annotations

backnotprop:
## Summary
This PR enhances the feedback export functionality to include source line numbers for annotations, with special handling for documents converted from HTML/URL sources where line numbers are caveated.

## Key Changes

- **Line number extraction**: Added `lineLabelForAnnotation()` helper to resolve source line numbers from block metadata, filtering out global comments and diff-view annotations
- **Export options interface**: Created `ExportAnnotationsOptions` to support the `sourceConverted` flag, which adds a caveat note when markdown was converted from HTML/URL via Turndown/Jina
- **Main document feedback**: Updated `exportAnnotations()` to include line labels in annotation headers and display a caveat when `sourceConverted` is true
- **Linked document feedback**: Enhanced `exportLinkedDocAnnotations()` to:
  - Accept `LinkedDocAnnotationEntry` with optional `blocks` and `isConverted` fields
  - Include line numbers for linked doc annotations
  - Display conversion caveat in section headers
- **Document caching**: Modified `useLinkedDoc` hook to cache markdown and conversion status alongside annotations, enabling line number computation during export
- **Server-side tracking**: Updated all annotation entry points (extension, hook, OpenCode plugin) to detect and propagate the `sourceConverted` flag when:
  - Loading from URLs (Jina conversion)
  - Loading from `.html` files (Turndown conversion)
- **API responses**: Modified document handlers in both extension and Bun servers to include `isConverted` flag in responses
- **State management**: Added `sourceConverted` state to App component and threaded it through to export functions and linked doc hook

## Implementation Details

- Line labels are only shown for annotations with valid block references (excludes diff-view and global annotations)
- Conversion detection is automatic based on file extension (`.html`/`.htm`) or URL source
- Caveats are added at both the document level (linked docs) and globally (main feedback) to ensure users understand line number limitations
- The markdown content is cached per linked document to enable block parsing during export without re-fetching

### DIFF
--- a/apps/hook/server/index.ts
+++ b/apps/hook/server/index.ts
@@ -559,6 +559,7 @@ if (args[0] === "sessions") {
   let folderPath: string | undefined;
   let annotateMode: "annotate" | "annotate-folder" = "annotate";
   let sourceInfo: string | undefined;
+  let sourceConverted = false;
 
   // --- URL annotation ---
   const isUrl = /^https?:\/\//i.test(filePath);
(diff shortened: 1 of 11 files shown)


backnotprop:
Superseded by a clean reimplementation that fixes the frontmatter line offset bug, adds line ranges for multi-line blocks, and correctly tracks the conversion flag per-document instead of using the root flag.

PR #630, opened 2026-04-30
fix(review): gutter hover button + bump @pierre/diffs to 1.1.20

backnotprop:
## Summary

- **Fix gutter hover button not rendering** — `renderHoverUtility` was calling `getHoveredLine()` during React render to decide whether to show the button. But Pierre updates hover state imperatively via pointer events (no React re-render), so the button was permanently null. Now the button always renders; `getHoveredLine()` is only called at click time.
- **Fix button styling** — matched Pierre's own `[data-utility-button]` spec: `1lh` sizing, `--diffs-modified-base` background, correct margin/padding. Previously used hardcoded `1.25rem` and `oklch()` purple that didn't match Pierre's theme.
- **Bump `@pierre/diffs` from 1.1.12 → 1.1.20** — picks up split-view scroll fix, WorkerPool race condition fix, hydration fixes, CSS refactoring, and empty-file-as-deleted fix.

## Test plan

- [ ] Hover over line numbers in split and unified diff — the `+` button should appear on every line hover
- [ ] Click the `+` button — comment toolbar should open for that line
- [ ] Button color should match Pierre's blue theme, not purple
- [ ] Button should not clip or overflow the line number cell
- [ ] Verify no regressions with hunk expansion, line selection, annotations

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -8,7 +8,7 @@
         "@anthropic-ai/claude-agent-sdk": "^0.2.92",
         "@openai/codex-sdk": "^0.118.0",
         "@opencode-ai/sdk": "^1.3.0",
-        "@pierre/diffs": "^1.1.12",
+        "@pierre/diffs": "^1.1.20",
         "diff": "^8.0.4",
         "dockview-react": "^5.2.0",
         "dompurify": "^3.3.3",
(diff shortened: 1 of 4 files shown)


backnotprop:
**Note for follow-up work:** The `<File>` component is available from `@pierre/diffs/react` as of 1.1.20 — it renders a single syntax-highlighted file (no diff). We don't use it in this PR, but it's there for any future work that needs plain file rendering (e.g. viewing unchanged files, full-file context, etc.).

PR #641, opened 2026-05-01
Localize Pi extension review flow strings

jerryfan:
I found one focused UX issue in the Pi extension: plan/review/annotation flows have status and error text exactly where users approve plans, request changes, open code review, or annotate files. If those messages are misread, the user can miss whether a review is closed, approved, missing assets, or waiting for feedback.

This PR makes that Pi-extension path optionally localizable while preserving the current English strings as fallbacks.

- No new dependency
- No behavior change without an i18n provider
- English remains the default/fallback
- Locales: ja, zh-CN, es — common developer workflow locales for review/approval UI
- Covers Pi command descriptions and high-friction review/annotation/archive notifications
- Does not touch generated browser UI bundles

Validation:
- npm pack --dry-run from apps/pi-extension

Note: bun is not available in this environment, so I could not run the repo's bun test/typecheck scripts here.

### DIFF
--- a/apps/pi-extension/i18n.ts
+++ b/apps/pi-extension/i18n.ts
@@ -1,0 +1,147 @@
+import type { ExtensionAPI } from "@mariozechner/pi-coding-agent";
+
+type Params = Record<string, string | number>;
+type Translate = (key: string, fallback: string, params?: Params) => string;
+
+let translate: Translate = (_key, fallback, params) => format(fallback, params);
+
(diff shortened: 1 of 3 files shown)


jerryfan:
Retracting this PR. I used locale choices that were too broad and not sufficiently supported by explicit repo/product signals. That was not a good fit for a first-touch contribution. I will come back only with a smaller, better-scoped PR using the project’s actual signals and/or a safer Western-default locale set.

PR #644, opened 2026-05-01
Feat: add Codex Plannotator skills

leoreisdias:
## Summary
- add Codex-callable Plannotator skills for `review`, `annotate`, and `last`
- document the Codex skill entrypoints in the README
- keep plan mode out of scope; this only improves the existing manual Codex flows

## Why
Codex app does not accept `!` commands in chat, so the current direct shell-style Plannotator invocations are awkward or unavailable in the app experience.

There is also a workflow gap when Plannotator returns annotations in Codex CLI: the agent receives the feedback, but it does not proactively act on it unless the user sends another explicit follow-up message.

These skills help in both places:
- they make Plannotator callable from Codex app through `$plannotator-*` skills
- they instruct the agent to run the command itself, wait for the browser session to finish, and then act on the returned feedback without requiring an extra user prompt

## Screenshots
Codex App - Skill execution and waiting responses:
image

Codex App - Receiving annotation and acting on it:
image

## Scope notes
This PR focuses only on the user-invoked Codex flows that already map well to Plannotator today.

## Test plan
- verify the new skill folders are present under `apps/skills/`
- verify the README documents the Codex skill entrypoints

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,12 @@ irm https://plannotator.ai/install.ps1 | iex
 **Then in Codex — feedback flows back into the agent loop automatically:**
 
 ```
+$plannotator-review          # Code review skill for current changes
+$plannotator-annotate        # Annotate a markdown file, URL, or folder
+$plannotator-last            # Annotate the last agent message
+```
+
+```
 !plannotator review           # Code review for current changes
 !plannotator review <pr-url>  # Review a GitHub pull request
 !plannotator annotate file.md # Annotate a markdown file
--- a/apps/skills/plannotator-annotate/SKILL.md
+++ b/apps/skills/plannotator-annotate/SKILL.md
@@ -1,0 +1,23 @@
+---
+name: plannotator-annotate
+description: Open Plannotator's annotation UI for a markdown file, converted HTML file, URL, or folder and then respond to the returned annotations.
+---
+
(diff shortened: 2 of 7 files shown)


backnotprop:
@leoreisdias thanks!!

backnotprop:
How will these get installed. Maybe we should ship a plugin

leoreisdias:
> How will these get installed. Maybe we should ship a plugin

thanks for the follow up @backnotprop !

about the installation, the existing installer already copies `apps/skills/*` into `~/.agents/skills` _(install.sh / install.ps1 / install.cmd)_, and codex can pick skills up from there, so a plugin wouldnt be needed as these should install from the normal Plannotator script once they re in a release

but I like the idea for speed and discoverability. I am going to take a look at how this is usually set up and whats needed. Should we do it within this PR or a separated one? thoughts? 🙂

backnotprop:
Merging, have a cool stop hook for plan mode coming in as well #577

backnotprop:
@leoreisdias

Thanks for [#644](https://github.com/backnotprop/plannotator/pull/644).

leoreisdias:
@backnotprop 

Thank for letting me know! Amazing initiative. Looking forward to know more about it as things evolve. I’m around if there’s anything else I can help with :)

PR #685, opened 2026-05-08
Add tab size option to code review

skripalschikov:
When reviewing code, indentation in the diff viewer was hard-coded to 2 spaces, which is too narrow for files that use tabs (Go, Makefiles, deeply nested code). Reviewers had no way to change it.

This PR adds a "Tab size" control to the diff options popover, letting each reviewer pick a width from 1 to 8. The choice persists across sessions and is applied to the diff viewer immediately — no reload needed.

Default stays at 2, so nothing changes for users who don't touch it.

image


backnotprop:
Awesome

backnotprop:
@skripalschikov

Thanks for [#685](https://github.com/backnotprop/plannotator/pull/685).

PR #686, opened 2026-05-08
fix(review): resolve jj default target from trunk

graemefolk:
Small update to resolve the default jj compare target from `trunk()` to the concrete bookmark that jj reports, rather than iterating a hardcoded list of potentials.


backnotprop:
Thanks @graemefolk !

PR #668, opened 2026-05-05
Expose bypass clear reminder permission mode

AgileInnov8tor:
## Summary
- Adds a synthetic persisted `bypassPermissionsClearReminder` permission mode for Settings and first-run setup.
- Decomposes that UI mode to Claude Code wire fields: `permissionMode: 'bypassPermissions'` plus `clearContextNudge: true`.
- Validates stored permission-mode cookie values before reusing them.

## Test Plan
- `bun test packages/editor/approvalBody.test.ts`
- `bun run --cwd apps/review build && bun run build:hook`
- `bun run --cwd packages/ui typecheck`
- `PATH="$PWD/packages/ui/node_modules/.bin:$PATH" bun run typecheck`
- `bun test`
- `git diff --check c7727b9..HEAD`

## Notes
- Direct push to `backnotprop/plannotator` was denied for the authenticated accounts, so this PR is opened from the `AgileInnov8tor/plannotator` fork.

PR #693, opened 2026-05-11
feat(hook): one-click clear-context + bypass via native plan-accept + keystroke injection

AgileInnov8tor:
## Summary

Proper replacement for the reverted PR #668 (Expose bypass clear reminder permission mode).

PR #668 was reverted because the hook emitted a `/clear` reminder that was never actually displayed — a silent no-op. This PR takes a different approach: instead of trying to do the impossible (hooks cannot clear context directly), it defers to Claude Code's own native primitive.

### What changed

**Four commits:**

1. **Wire truthful approval semantics** (9096eff, c7727b9) — thread `permissionMode` and a new `deferToNativeForClear` flag through the decision pipeline so the hook knows when to defer vs. when to nudge.

2. **Consent-gated native deferral** (ff79946) — when user picks "Approve + Bypass + Clear Context (native)":
   - Hook writes `showClearContextOnPlanAccept: true` to `~/.claude/settings.json` (consent-gated, atomic write with retry)
   - Hook exits 0 with **empty stdout** (native passthrough)
   - Claude Code's built-in plan-accept dialog appears and shows "Yes, clear context and bypass permissions" (because the setting is now on)

3. **Keystroke injection for one-click UX** (12cddf7, `apps/hook/server/keystrokeInjector.ts`) — before exiting 0, spawn a detached background process that fires "1\n" into the CC terminal after 600ms:
   - **tmux** (`$TMUX_PANE`): `tmux send-keys -t <pane> 1 Enter` (no accessibility permissions needed)
   - **macOS non-tmux**: `osascript` iterating `{warp, iTerm2, Terminal}` via System Events
   - **Linux/Windows without tmux**: no-op (user sees native dialog, presses 1 manually)
   - Silent-fail: accessibility denied or terminal not found → falls back to manual press, no crash

4. **Slop cleanup** (fb6c44d) — no behavior change; removes AI-generated bloat from keystrokeInjector:
   - `keystrokeInjector.ts` 78→39 lines: delete 16-line module docblock + 5-line JSDoc; inline two single-use builder functions; unify the duplicated `Bun.spawn+.unref()` into a single call
   - `keystrokeInjector.test.ts` 118→107 lines: extract `setPlatform(v)` helper replacing 5× repeated `Object.defineProperty` blocks

### Why not a hook `clearContext` field?

CC's `updatedPermissions` types are `addRules`/`setMode`/`removeRules`. There is no `clearContext` entry (verified by exhaustive string scan of the CC 2.1.138 binary). The plan-accept TUI dialog is hardcoded in CC's React renderer and cannot be bypassed from a hook. The native passthrough + keystroke injection is the only viable path that preserves true context-clearing without patching CC.

### User experience

Before: Review plan in plannotator → click → still have to press "1" in terminal

After: Review plan in plannotator → click "Approve + Bypass + Clear Context (native)" → context cleared, bypass mode set, plan executes — **zero additional keystrokes**

## Test plan

- [x] `bun test packages/server packages/editor apps/hook` — 265/265 pass
- [x] `bun run build:hook` — clean build
- [x] Manual smoke test on WarpTerminal (macOS): ExitPlanMode → plannotator UI → click entry → CC terminal auto-selects option 1 → plan executes
- [x] Headless probes (5 scenarios): with-consent, no-consent, already-enabled, no-settings, malformed-settings — all pass

## Notes

- Consent for the settings mutation lives at `~/.plannotator/consent/clear-context-setting.json`
- The `showClearContextOnPlanAccept` setting defaults to `false` in CC (changelog: "hidden by default in plan mode"); this PR turns it on consent-gated, not globally
- Keystroke delay (600ms) is configurable in code; conservative enough for CC dialog render time (~200–500ms)


backnotprop:
> Before: Review plan in plannotator → click → still have to press "1" in terminal

Can you explain what do you mean by this? 'Cause right now it's all automated. You just approve and the agent goes on.

AgileInnov8tor:
The "before" is specific to the **"Approve + Bypass + Clear Context (native)"** option, not the standard approve flow which is already zero keystrokes.

Background: `showClearContextOnPlanAccept: true` must be set in `~/.claude/settings.json` for the native dialog to appear. The hook writes this on first use (consent gated). Once set, CC waits for the user to press `1` in the terminal. This PR automates that keypress via tmux/osascript.

backnotprop:
Okay, I think I understand. So this would require the user to turn that on and then Instead of auto approving the plan, it just pauses and then the user gets the option to clear context.

AgileInnov8tor:
Yes exactly

PR #700, opened 2026-05-11
Fix OpenCode planning agent resolution for display names

pysol-dev:
## Summary
- Resolve configured OpenCode `planningAgents` to existing agent config keys before mutating permissions.
- Match display-name agent keys such as `Prometheus - Plan Builder` / `Sisyphus (Ultraworker)` and keys prefixed with zero-width characters, avoiding duplicate raw `prometheus` / `sisyphus` agent entries.
- Preserve allow permissions during the later non-planning-agent deny pass by tracking resolved planning-agent config keys.

## Verification
- `bun test apps/opencode-plugin/workflow.test.ts` ✅
- LSP diagnostics clean for `apps/opencode-plugin/workflow.ts` and `apps/opencode-plugin/workflow.test.ts` ✅
- `bun test apps/opencode-plugin` was attempted, but this temp checkout lacks workspace dependency resolution for `@plannotator/server`, so unrelated command tests fail before exercising this change.

## Context
This fixes a real OpenCode config interaction where `planningAgents: ["prometheus", "sisyphus"]` created extra raw agent entries instead of reusing existing OhMyOpenCode display-keyed agent configs.

### DIFF
--- a/apps/opencode-plugin/workflow.test.ts
+++ b/apps/opencode-plugin/workflow.test.ts
@@ -186,6 +186,38 @@ describe("applyWorkflowConfig", () => {
     expect(config.agent.build.permission.submit_plan).toBe("deny");
   });
 
+  test("plan-agent mode resolves planningAgents to existing display-named agents", () => {
+    const prometheusKey = "\u200B\u200B\u200BPrometheus - Plan Builder";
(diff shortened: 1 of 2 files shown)


backnotprop:
@pysol-dev

Thanks for [#700](https://github.com/backnotprop/plannotator/pull/700).

PR #708, opened 2026-05-12
fix: update Codex hooks feature flag

leoreisdias:
## Problem

Codex no longer wants users or installers to enable hooks with `[features].codex_hooks`. The current CLI reports that flag as deprecated and asks users to use `[features].hooks` instead.

Plannotator still wrote and documented the old flag in a few places: the Unix installer, Windows manual guidance, Codex setup docs, troubleshooting docs, local E2E harnesses, and the release smoke test. That means new installs could produce deprecation warnings, and the project documentation pointed users at a stale Codex configuration shape.

## Solution

Update Plannotator’s Codex setup path to use the current `[features].hooks = true` flag everywhere user-facing or test-facing.

The Unix installer now writes `hooks = true` for new Codex configs and also migrates an existing managed `codex_hooks = ...` entry to `hooks = true` when it updates the `[features]` table. Inline `features = ...` configs are still left untouched, with the manual guidance updated to show the new flag.

This also updates the docs, Windows installer guidance, manual Codex harness commands, and the release workflow assertion so they all agree on the same current Codex feature flag.

## Validation

- `bun test scripts/install.test.ts`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -353,7 +353,7 @@ jobs:
           run_installer
 
           test -x "$tmp_home/.local/bin/plannotator"
-          grep -q 'codex_hooks = true' "$tmp_home/.codex/config.toml"
+          grep -q 'hooks = true' "$tmp_home/.codex/config.toml"
 
           HOME="$tmp_home" node <<'NODE'
           const fs = require("fs");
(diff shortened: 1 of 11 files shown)


backnotprop:
thanks!

PR #710, opened 2026-05-12
Add plan context map sidebar

woertedetiankong:
## Summary
- add a Context sidebar tab for markdown plans with detected code-file mentions
- group mentions into a compact folder-style tree with mention counts and lightweight status badges
- let file rows jump back to the first plan block that mentions that file

## Details
- reuses the existing shared code-path extraction rules and adds a repeated-mention helper
- infers Modified, Created, and Deleted only from explicit nearby file-change headings, with conflicts falling back to Mentioned
- reuses /api/doc/exists validation via the existing validated-code-path hook so unresolved paths render muted
- hides the Context tab in archive, annotate, linked-doc, HTML, and active plan-diff views

## Validation
- bun test packages/shared/extract-code-paths.test.ts
- bun test packages/ui/utils/planContext.test.ts
- bun run typecheck
- bun run build:hook

### DIFF
--- a/packages/editor/App.tsx
+++ b/packages/editor/App.tsx
@@ -1,6 +1,7 @@
 import React, { useState, useEffect, useLayoutEffect, useMemo, useRef, useCallback } from 'react';
 import { toast, Toaster } from 'sonner';
 import { type Origin, getAgentName } from '@plannotator/shared/agents';
+import { parseCodePath } from '@plannotator/shared/code-file';
 import { parseMarkdownToBlocks, exportAnnotations, exportLinkedDocAnnotations, exportEditorAnnotations, exportCodeFileAnnotations, extractFrontmatter, wrapFeedbackForAgent, Frontmatter, type LinkedDocAnnotationEntry } from '@plannotator/ui/utils/parser';
 import { Viewer, ViewerHandle } from '@plannotator/ui/components/Viewer';
 import { HtmlViewer } from '@plannotator/ui/components/html-viewer';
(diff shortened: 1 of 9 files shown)


backnotprop:
yo this is sick

backnotprop:
I'm only concerned about consistency. reviewing some things.

backnotprop:
@woertedetiankong — "yo this is sick" was four months ago and then nothing; that is on us. The feature is not on main (no `packages/ui/utils/planContext.ts`, no `PlanContextBrowser`), and the need is still tracked as the first acceptance criterion of **#900** ("the plan review UI can show relevant planning context when available"), so this is the reference implementation for a live requirement.

The hard part: the branch is **635 commits** behind and it touches `SidebarContainer`/`SidebarTabs`/`useSidebar`, which have changed a great deal since May — the sidebar gained the Archive and Messages browsers and a folder file browser, plus a mobile shell. A rebase is likely closer to a re-apply than a merge.

Two things would make that worth your time before you start: (1) @backnotprop's "I'm only concerned about consistency" needs to become a concrete answer — does the Context tab belong beside Contents/Versions/Archive, or inside #900's broader plan-context panel? (2) The status inference (Modified/Created/Deleted from nearby headings) is the part most likely to need a decision, since #900 asks for context from real session/transcript data rather than inferred metadata. Happy to get you that answer if you still want to carry it; otherwise say so and I'll close this with the work credited in #900. Auto-closing in 14 days without a reply.

PR #730, opened 2026-05-15
feat(submit-plan): replace dual-mode with edit-based interface

rcdailey:
## Summary

Replaces the dual-mode `submit_plan` tool (inline text or file path with auto-detect) with a single edit-based interface. The tool now accepts an array of line-range edits applied against a plugin-managed backing file. On denial, the response includes the plan with line numbers so the agent can target surgical revisions instead of resubmitting the full document.

This addresses [#365](https://github.com/backnotprop/plannotator/issues/365) by collapsing the two input modes into one deterministic path. The agent no longer has to choose between text and file-path modes, and the plugin no longer depends on prompt engineering to steer the agent toward the right mode.

## How it works

First submission (full plan as a single edit):

```json
{ "edits": [{ "start": 1, "content": "# My Plan\n\n## Goals\n..." }] }
```

Revision after denial (targeted edits using line numbers from the response):

```json
{ "edits": [
  { "start": 12, "end": 14, "content": "revised section" },
  { "start": 30, "end": 30, "content": "" }
] }
```

The backing file lives at `.opencode/plans/_active-plan.md` and is managed entirely by the plugin. The agent never touches it directly.

## Tested with 3 denial rounds

Ran a real planning session with Opus across 3 rounds of increasing difficulty:

1. Simple wording change on a single line
2. Structural reorder (moved a section, deleted another) requiring multi-range edits
3. Targeted insertion + schedule change after the structural reorder

The agent tracked line numbers accurately across all rounds. The line-numbered plan in each denial response gave it a fresh ground truth, so it never drifted.

## Changes

- `apps/opencode-plugin/index.ts`: New `applyEdits`, `validateEdits`, `formatWithLineNumbers` functions. Input validation (bounds, overlap detection, 5MB size limit). Removed old `isFilePath`/`resolvePlanContent`.
- `apps/opencode-plugin/plan-mode.ts`: Removed dead code from the file-path workflow (`validatePlanPath`, `getPlanDirectory`). Only permission helpers and prompt-stripping remain.
- `apps/opencode-plugin/submit-plan.test.ts`: 25 new tests covering edit application, validation, and line formatting.
- `apps/opencode-plugin/plan-mode.test.ts`: Removed tests for deleted functions.
- `apps/marketing/.../opencode.md`: Updated "How the plugin works" for edit-based flow.
- `apps/marketing/.../custom-feedback.md`: Updated template variable descriptions.

Refs #365


lgladysz:
Probably this messed up my opencode flow :(.
Now my agent is trying to reread the long plan from md file and pass it to the submit_plan tool. This often results in a trimmed plan with edits that are present on Plannotator but not on the physical file. Please give opencode the ability to just pass md file as a source.

rcdailey:
@lgladysz The file-path mode was removed intentionally because it had its own reliability issues (the agent would sometimes pick the wrong mode, or the path wouldn't resolve correctly). The edit-based approach avoids that, but it does depend on the model handling tool calls well at larger sizes.

I suspect this might be model-dependent. I've been testing with Opus 4.6 and haven't hit truncation. Could you open a separate issue with:

- Which model you're using
- Rough plan size (line count is fine)
- A session transcript if possible (so we can see where the trimming happens)

That would help narrow down whether it's a model limitation, a prompt issue, or something else.

backnotprop:
I think we can bring it back with optional setting.

rcdailey:
@backnotprop I think it's worth waiting for more detail from @lgladysz before deciding on an approach. We don't even know what model they're using yet, and this could just be a model limitation rather than something that needs a code change. Adding the file-path mode back (even as an option) reintroduces the complexity #365 was trying to eliminate.

kylerm42:
Removing the ability to pass in a source file has definitely impacted my workflow. I have all my specs written to files in a specific place, and there have been multiple times when there's been a gap or differences in what gets submitted as a plan and what ends up in the file. This is using Sonnet 4.6 and <300 line specs, so a fairly capable model and not crazy long specs.

I know part of the reason for this change was to "remove the prompt engineering" aspect of it, but it's actually added prompt engineering for me. Now I'm having to give it explicit instructions to copy the file contents *exactly* into the `submit_plan` tool, which still doesn't typically happen. I've had issues going both ways, whether it's writing the file first, then submitting the plan, or vice versa.

I'm going to try to see if I can mimic the old behavior with agents calling `plannotator annotate <spec path>` instead of `submit_plan`, but I wish I didn't have to deal with workarounds.

rcdailey:
@kylerm42 You're actually on the right track with the annotate idea. `submit_plan` is built for agent-authored plans where the plugin owns the backing file, so it doesn't faithfully echo a file you wrote yourself. That's the gap you're hitting: it makes the agent re-type your spec into a tool call, and the agent doesn't do it exactly. The file-as-source mode got pulled in #365 because "let the agent pick the mode" was unreliable, and adding it back would just bring that problem back.

For what you're doing (specs written to files first, then reviewed) `plannotator annotate <spec path>` is the supported path, not a workaround. In OpenCode that's `/plannotator-annotate <file>`. It reads the file straight off disk and renders it as-is, so the file stays the source of truth and the review can't get out of sync with it. #570 added an approve/reject gate for exactly this spec-driven flow: pass `--gate` and you get Approve / Annotate / Close, with annotations sent back to the session.

If annotate falls short for your setup, I'd rather fix that path than put dual-mode back into `submit_plan`.

PR #736, opened 2026-05-17
Fix directory name for OpenCode commands

lfox91:
Bug: The path used for opencode's commands dir in the install.sh script doesn't match the [opencode docs](https://opencode.ai/docs/commands/#:~:text=Create%20markdown%20files%20in%20the%20commands/%20directory%20to%20define%20custom%20commands)

Proposed Fix:
I believe an s is missing from the OPENCODE_COMMANDS_DIR path
/command -> /commands


backnotprop:
ty!! I added same fixes to window scripts

backnotprop:
@lfox91

Thanks for [#736](https://github.com/backnotprop/plannotator/pull/736).

PR #747, opened 2026-05-18
fix: add .catch() guards to unguarded fire-and-forget promises

buihongduc132:
## What

Two places where `.then()` was called without `.catch()` on fire-and-forget promises, risking unhandled rejections that could crash the host process.

## Why

When a promise is fired and forgotten (not awaited or returned), any rejection becomes an unhandled promise rejection. In Node.js/Bun, these can terminate the process or trigger warning spam.

### 1. `packages/server/review.ts`
`detectRemoteDefaultCompareTarget()` is called as a non-blocking best-effort check to auto-detect the remote default branch. If git fails (no git installed, invalid cwd, network timeout), the rejection was unhandled.

Fix: Add `.catch()` that swallows the error — the detection is purely cosmetic and the server should keep running fine without it.

### 2. `packages/ai/providers/pi-sdk.ts`
`proc.exited` fires when the Pi process dies. The `.then()` callback cleans up pending requests and notifies listeners. But if the process exits abnormally, the promise itself rejects — and since no one catches it, it becomes an unhandled rejection.

Fix: Add `.catch()` that swallows it. The cleanup already ran in `.then()`.

### Not changed
Integration saves (Obsidian/Bear/Octarine) use `Promise.allSettled()` which already handles rejections — no fix needed there.

## Testing
- 3 new tests in `unguarded-promise.test.ts` covering the error paths
- Existing tests unaffected

## Scope
~5 LOC of actual fixes (just `.catch()` additions). Zero behavior change for happy paths.

### DIFF
--- a/packages/ai/providers/pi-sdk.ts
+++ b/packages/ai/providers/pi-sdk.ts
@@ -64,6 +64,9 @@ class PiProcess {
 			for (const listener of this.listeners) {
 				listener({ type: "process_exited" });
 			}
+		}).catch(() => {
+			// Swallow — proc.exited rejection shouldn't crash the host process.
+			// The cleanup above already ran in the .then() callback.
 		});
 	}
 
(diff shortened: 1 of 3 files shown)


backnotprop:
Thanks for the contribution. We’re closing this during our stale PR cleanup.\n\nThe current review startup path now handles remote-default detection rejection explicitly. The remaining Pi process change does not have a demonstrated production failure, and the added tests exercise toy promise chains rather than the production paths they claim to protect.\n\nIf an unhandled rejection is reproducible against current main, please open a focused issue or new PR with the reproduction. We’re happy to reconsider it.

PR #749, opened 2026-05-18
fix(server): respect PLANNOTATOR_SERVER_URL and PLANNOTATOR_HOST env vars

buihongduc132:
## What

All three servers hardcoded `http://localhost:{port}` for their server URL. No way to override it even though the README documents `PLANNOTATOR_SERVER_URL` for remote access.

## Why

Two real scenarios where hardcoded localhost doesn't work:

1. **Behind a reverse proxy / tunnel** — Cloudflare Tunnel, ngrok, Tailscale Funnel. The external URL is different from the listen address. You need to tell the server what URL to advertise.

2. **Binding to a specific network interface** — Tailscale IP, LAN IP, etc. You want the browser to open `http://100.x.x.x:port` not `http://127.0.0.1:port`.

### `packages/server/remote.ts`
Added two new exported functions:

- **`getServerUrl(port)`** — Returns the URL for accessing the server.
  - Priority: `PLANNOTATOR_SERVER_URL` → `PLANNOTATOR_HOST` → `http://127.0.0.1:{port}`
  - For remote sessions (0.0.0.0 binding), still advertises `localhost` since `0.0.0.0` isn't clickable

- Updated **`getServerHostname()`** — Now checks `PLANNOTATOR_HOST` before falling back to remote/local defaults.

### `packages/server/index.ts`, `annotate.ts`, `review.ts`
Replaced hardcoded ``http://localhost:${port}``` with `getServerUrl(port)` in all three servers.

## Testing
5 new tests covering all priority levels and env var combinations.

## Backwards Compatibility
100% — when no env vars are set, behavior is identical to before.

### DIFF
--- a/packages/server/annotate.ts
+++ b/packages/server/annotate.ts
@@ -11,7 +11,7 @@
  *   PLANNOTATOR_PORT   - Fixed port to use (default: random locally, 19432 for remote)
  */
 
-import { isRemoteSession, getServerHostname, getServerPort } from "./remote";
+import { isRemoteSession, getServerHostname, getServerPort, getServerUrl } from "./remote";
 import { getRepoInfo } from "./repo";
 import type { Origin } from "@plannotator/shared/agents";
 import { handleImage, handleUpload, handleServerReady, handleDraftSave, handleDraftLoad, handleDraftDelete, handleFavicon } from "./shared-handlers";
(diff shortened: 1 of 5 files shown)


backnotprop:
Thanks for the contribution. We’re closing this as superseded by the current advertised-URL design.

Plannotator now supports `PLANNOTATOR_URL_HOST` and the `urlHost` config setting, including validation and Tailscale auto-detection. That design deliberately keeps the advertised URL separate from the listening interface, whereas this PR targets older environment names and couples those concerns.

If the current URL-host support does not cover a concrete deployment, please open a new issue describing that setup.

PR #750, opened 2026-05-18
feat(server): add /api/decision poll + SSE endpoints for remote clients

buihongduc132:
## What

Two new endpoints that let remote clients (SSH, devcontainer, Docker) check the plan decision status over HTTP.

## Why

When the server runs locally, the CLI process calls `server.waitForDecision()` directly — works great. But when running remotely (e.g., `PLANNOTATOR_REMOTE=1` or behind a proxy), the client connecting via HTTP has no way to know when the user approved or denied the plan. It just has to... guess? Or poll the `/api/plan` response and look for changes? Neither is great.

### `GET /api/decision`
Simple polling endpoint. Returns `{ pending: true }` until the user decides, then returns the full decision result (`{ approved, feedback, savedPath, ... }`).

### `GET /api/decision/stream`
SSE stream that pushes the decision result as soon as it happens. More efficient than polling — especially over slow connections. If the decision was already made before connecting, it sends the result immediately and closes.

### Implementation
- Added `decisionResult` variable that stores the resolved decision (was previously only available via Promise)
- Both endpoints are purely additive — no change to existing approve/deny flows

## Testing
3 new tests covering pending state, resolved state, and SSE stream behavior.

## Backwards Compatibility
100% — these are new endpoints that don't affect anything existing.

### DIFF
--- a/packages/server/decision-endpoints.test.ts
+++ b/packages/server/decision-endpoints.test.ts
@@ -1,0 +1,109 @@
+/**
+ * PR D: Tests for /api/decision poll + SSE endpoints
+ *
+ * For remote clients (separate process from server), the only way to get
+ * the decision result is via HTTP. These endpoints provide both polling
+ * and real-time SSE options.
+ */
+import { describe, test, expect, afterAll, beforeAll } from "bun:test";
+
+describe("/api/decision endpoints", () => {
+  const controllers: AbortController[] = [];
+  let savedPort: string | undefined;
+  let savedRemote: string | undefined;
+
+  beforeAll(() => {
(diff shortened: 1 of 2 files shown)


backnotprop:
Thanks for the contribution. We’re closing this as not planned during our stale PR cleanup.

Remote mode does not currently create a separate decision consumer: the owning CLI or extension process still awaits `waitForDecision()`, while the browser submits through the existing approve/deny endpoints. There is no current client for these polling/SSE endpoints.

Adding a remotely reachable decision API would also need a concrete consumer, authentication and lifecycle semantics, and parity across the Bun and Pi server implementations. We’re happy to reconsider a new proposal built around a real integration requirement.

PR #751, opened 2026-05-18
feat(server): add optional sessionId and cwd to ServerOptions/ServerResult

buihongduc132:
## What

Two new optional fields on `ServerOptions` that get passed through to `ServerResult`:

- **`sessionId`** — optional string, auto-generates a UUID if omitted
- **`cwd`** — optional string, defaults to `process.cwd()`

## Why

I'm building a Pi coding agent extension that runs plannotator as a long-lived server with multiple concurrent sessions. Right now there's no way to identify which session a server belongs to, or what working directory it was started from.

These two fields are the foundation for session isolation without changing any existing behavior. Down the line they enable:
- Draft storage scoped per session (instead of global)
- Decision tracking per session
- Remote clients managing multiple sessions

### `ServerOptions` (input)
```ts
sessionId?: string;  // auto-generated UUID if omitted
cwd?: string;        // falls back to process.cwd()
```

### `ServerResult` (output)
```ts
sessionId: string;   // always present — provided or generated
cwd: string;         // always present — provided or process.cwd()
```

## Testing
4 tests covering:
- Custom sessionId passed through
- Auto-generated sessionId (UUID format)
- Custom cwd passed through
- Default cwd (process.cwd())

## Backwards Compatibility
100% — both fields are optional on input, and adding non-optional fields to the output interface is safe since the interface was already being constructed internally.

### DIFF
--- a/packages/server/index.ts
+++ b/packages/server/index.ts
@@ -81,6 +81,10 @@ export interface ServerOptions {
   onReady?: (url: string, isRemote: boolean, port: number) => void;
   /** OpenCode client for querying available agents (OpenCode only) */
   opencodeClient?: OpencodeClient;
+  /** Optional session ID for isolating storage and decisions per session */
+  sessionId?: string;
+  /** Working directory for the project (defaults to process.cwd()) */
+  cwd?: string;
   /** When set to "archive", server runs in read-only archive browser mode */
   mode?: "archive";
   /** Custom plan save path — used by archive mode to find saved plans */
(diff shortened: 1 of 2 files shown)


backnotprop:
Thanks for the contribution. We’re closing this because the proposed fields do not provide the session isolation described in the rationale.

The PR returns a generated `sessionId` and `cwd`, but nothing consumes them: decision state remains single-session, storage is not scoped by the ID, `cwd` is not threaded into server operations, and the server is not multiplexed. The current Pi integration now manages Pi session identity and per-server browser-session lifecycle through a different architecture.

A long-lived multi-session server would require a cohesive design around state ownership, routing, storage, lifecycle, and a concrete consumer. If that need returns, please open a new design issue or PR against the current architecture; we’re happy to reconsider it.

PR #752, opened 2026-05-18
fix(opencode): skip end bounds check on empty backing file

### LINKED ISSUE #742: submit_plan fails in plan-agent mode from Opencode after 0.19.18
## Summary

`submit_plan` appears broken when invoked through the allowed `plan` agent path from Opencode.

Direct orchestrator calls are correctly blocked in plan-agent mode, but once the request is routed through the `plan` agent, the initial submit still fails before the review flow starts.

This seems related to Opencode integration and started after the `0.19.18` changes in Plannotator.

## What works

- Direct orchestrator call is rejected as expected in plan-agent mode:
  - `submit_plan can only be called by: plan`
- Invoking the flow through a `plan` agent does reach `submit_plan`

## What fails

The first `submit_plan` call fails against an empty backing file with:

- `Error: end (1) exceeds file length (0)`

We also retried using the documented first-call payload shape:

```json
{
  "edits": [
    {
      "start": 1,
      "content": "# Smoke Test Plan\n\n1. Verify initial submit works.\n2. Confirm whether the review flow responds.\n"
    }
  ]
}
```

That still behaved as if `end: 1` were present and failed with the same error:

- `Error: end (1) exceeds file length (0)`

## Expected behavior

For the first `submit_plan` call on a new backing file, a single edit with `start=1` and `content` only should succeed and start the normal review/approval flow.

## Actual behavior

The plan-agent path appears to inject or normalize the payload in a way that behaves as though `end` is set on the initial call, causing validation to fail before the review flow appears.

## Repro notes

- Environment: Opencode
- Workflow: Plannotator plan-agent mode
- No repo files or Plannotator config were changed during testing
- This did **not** look like a permission/configuration issue; it looked like a wrapper/serialization issue in the plan-agent path

## Suspected regression window

Started after the `0.19.18` changes in Plannotator.
OpenCode `1.15.4`

rcdailey:
Fixes #742.

`validateEdits` rejects `end > lineCount` on first call when the backing file is empty (`lineCount = 0`). The plan-agent path in OpenCode includes `end` in the payload even when the agent omits it, so the first `submit_plan` call fails with `Error: end (1) exceeds file length (0)`.

`applyEdits` already handles this correctly (splice clamps on an empty array), so the fix just skips the bounds check when `lineCount === 0`. On an empty file every edit is a pure insert and `end` is meaningless.

This is the clean fix for the #730 revert in 0.19.19.

### DIFF
--- a/apps/opencode-plugin/index.ts
+++ b/apps/opencode-plugin/index.ts
@@ -185,7 +185,11 @@ export function validateEdits(existingLines: string[], edits: PlanEdit[]): strin
       if (!Number.isInteger(edit.end) || edit.end < edit.start) {
         return `end (${edit.end}) must be >= start (${edit.start})`;
       }
-      if (edit.end > lineCount) {
+      // On an empty file (lineCount === 0) every edit is a pure insert;
+      // end is semantically meaningless and applyEdits handles it via splice
+      // clamping. Rejecting here breaks first-call payloads where the agent
+      // or framework includes end (see #742).
+      if (edit.end > lineCount && lineCount > 0) {
         return `end (${edit.end}) exceeds file length (${lineCount})`;
       }
     }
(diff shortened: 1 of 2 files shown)


backnotprop:
thanks!

Issue #154, opened 2026-02-14
Browser does not open when launched via Claude Code skill (non-TTY environment)


jinhwan0724:
## Description

When `plannotator review` is executed via a Claude Code skill (slash command `/plannotator-review`), the server starts and listens on a port correctly, but the browser does not automatically open.

Running the same command directly in the terminal works fine — the browser opens and navigates to the correct URL.

## Steps to Reproduce

1. Install plannotator as a Claude Code plugin
2. Run `/plannotator-review` in Claude Code (which triggers the skill defined in `plannotator-review.md`)
3. The server starts and listens on a port (confirmed via `lsof`)
4. **The browser never opens**

## Expected Behavior

The browser should open and navigate to the plannotator review UI, just like when running `plannotator review` directly in a terminal.

## Actual Behavior

- The server starts and listens on the port successfully
- The browser does **not** open
- Manually running `open http://localhost:<port>` works as a workaround

## Root Cause Analysis

When Claude Code executes a skill command, it runs the process with:
- **stdin redirected from `/dev/null`** (`< /dev/null`)
- **No TTY attached** (non-interactive shell)

The actual shell invocation looks like:

```
/bin/zsh -c 'source ... && eval "plannotator review" < /dev/null'
```

It appears that `plannotator`'s browser-opening logic is either:
- Checking for a TTY and skipping `open` when none is detected, or
- Failing silently when attempting to open the browser in this environment

## Suggested Fix

Consider always attempting to open the browser regardless of TTY status, or provide a fallback that prints the URL to stdout so the calling agent can open it. For example:

```
Server running at http://localhost:51119
```

This would allow Claude Code (or any non-TTY caller) to detect the URL and open it.

## Environment

- macOS (Darwin 24.5.0, arm64)
- plannotator v0.8.0
- Claude Code CLI

rerorero:
Could https://github.com/backnotprop/plannotator/pull/163 be the cause?

materemias:
I am still experiencing this with v0.8.2

backnotprop:
Thanks for the report. I am consolidating the older backlog so related work has one current tracking issue.

This specific request is now tracked in #899 as part of the remote and multi-session session router work. During the June 13, 2026 older-than-two-month triage, I copied the relevant requirement from this issue into #899 and explained why it belongs with the related work.

I am closing this older issue so the backlog points at the current tracking issue. If there is a newer reproduction or a detail that is missing from #899, please comment there.

PR #756, opened 2026-05-19
fix(browser): treat no-op BROWSER values as unset so headless sessions can fall back

### LINKED ISSUE #724: Web UI does not open after ExitPlanMode on Windows + Claude Code (0.19.15) — server runs, plan captured
### Description

On Windows with Claude Code, the `PermissionRequest:ExitPlanMode` hook fires correctly and the Plannotator server starts on a random port — but the browser tab never auto-opens, so the user sees no UI and the plan-mode permission appears to be granted without review.

Structurally this looks like the same class of bug as #551 (closed; OpenCode + Linux + `xdg-open`), but with `cmd.exe /c start ${url}` failing silently in the Claude Code hook subprocess context on Windows.

### Environment

| | |
|---|---|
| OS | Windows 11 Pro 10.0.26200 |
| Agent | Claude Code 2.1.140 |
| Plannotator | 0.19.15 (plugin); bun 1.3.11 (binary) |
| Default browser | Microsoft Edge |
| Install | `/plugin install plannotator@plannotator` (user scope) |

### Steps to reproduce

1. Enable the plannotator plugin in Claude Code.
2. Enter plan mode (Shift+Tab) and ask Claude to make a plan.
3. Wait for the model to call `ExitPlanMode`.

### Expected behavior

Browser tab auto-opens at `http://localhost:<random>` showing the plan UI for annotation/approval.

### Actual behavior

- `ExitPlanMode` permission is granted effectively instantly with no UI prompt.
- No browser tab opens.
- The plannotator server **is** running. Confirmed by:
  - `~/.plannotator/sessions/<pid>.json` is created with a fresh port + URL.
  - The plan **is** captured in `~/.plannotator/history/<project>/<title>-<date>/NNN.md`.
  - `curl -sI http://localhost:<port>/` returns `HTTP 200` with the expected ~17 MB HTML payload — the UI is reachable, just never opened.

So the hook works end-to-end except for `openBrowser()`.

### Root cause hypothesis

In `packages/server/browser.ts`:

```ts
if (platform === "win32" || wsl) {
  await $`cmd.exe /c start ${url}`.quiet();
}
```

Wrapped in `try/catch` with `.quiet()`. When this fails in the Claude Code hook subprocess context — likely the same "child process can't reach the interactive desktop / env not propagated" mode that #551 hit on Linux/OpenCode — the failure is swallowed and only the VS Code IPC fallback is attempted. On a non-VS-Code Claude Code install there's no IPC registry to fall back to, so the user sees nothing.

(Bun's shell template-tag `$` interpolation may also be a factor — possible URL escaping/splitting that confuses `cmd.exe start` — but I haven't verified that vs. the subprocess-context theory.)

### Related

- #551 — Web UI does not open automatically (closed via #292, but later commenters reported the issue still occurring on Mac M4 and `PLANNOTATOR_BROWSER` didn't help; this Windows case looks like the same shape).
- #678 — Different failure mode (submit-side timeout on Windows/Codex, not open-side).

### Workaround (interactive)

The UI is reachable; you can open it manually:

```powershell
Get-Content ~/.plannotator/sessions/*.json |
  ConvertFrom-Json |
  Sort-Object startedAt |
  Select-Object -Last 1 -ExpandProperty url
```

Open the printed URL in a browser. Plan / annotation flow works normally from there.

### Notes

- Worked in 0.8.0 (which registered the hook with an absolute exe path and presumably a different open mechanism). 0.19.x regression.
- Have not yet tested `PLANNOTATOR_BROWSER=<full path to chrome.exe>` on Windows — the analogous workaround was reported as not-fixing-it in #551.
- Happy to test patches against this setup.

yonihorn:
## Summary

When `$BROWSER` (or `$PLANNOTATOR_BROWSER`) is set to a no-op sentinel like `true`, `false`, `none`, `:`, `0`, or `1` — as Claude Code's agent view does (`BROWSER=true`) — `openBrowser()` shells out via `Bun.$` and runs something equivalent to `true http://localhost:NNNN`. `true(1)` exits 0 without doing anything, the function returns success, and the server then sits in `waitForDecision()` forever. No browser appears, the URL isn't recoverable, and the session looks hung.

This is the BROWSER-sentinel branch of the same family as #163 (which fixed `BROWSER=open` / VS Code devcontainer scripts). The remaining sentinel case still affects every headless / background runner that conventionally sets `BROWSER=true` to disable browser launching.

## Fix

`packages/server/browser.ts`:

- Add `isNoOpBrowserSentinel()` recognising the documented no-op values (case- and whitespace-insensitive).
- In `openBrowser()`, treat a sentinel-valued `PLANNOTATOR_BROWSER` / `BROWSER` as if the variable were unset, so we fall through to the platform default (`open` / `xdg-open` / `cmd.exe /c start`) instead of shelling out to the sentinel.
- In `shouldTryRemoteBrowserFallback()`, treat the same sentinels as "no real handler configured" so the VS Code IPC fallback still fires in remote sessions where it would otherwise be skipped.

Net behaviour:

| Env                              | Before                              | After                                 |
| -------------------------------- | ----------------------------------- | ------------------------------------- |
| `BROWSER=true` (agent view)      | runs `true http://...` → no-op      | IPC fallback → platform default       |
| `BROWSER=/usr/bin/firefox`       | spawns firefox                      | spawns firefox (unchanged)            |
| `PLANNOTATOR_BROWSER='Google Chrome'` (macOS) | `open -a` chrome           | `open -a` chrome (unchanged)          |
| unset                            | platform default                    | platform default (unchanged)          |

No protocol or stdout changes, no new dependencies, no surface-area additions.

## Tests

Adds `packages/server/browser.test.ts` coverage:

- `shouldTryRemoteBrowserFallback` now returns `true` when `BROWSER` / `PLANNOTATOR_BROWSER` are sentinels.
- `isNoOpBrowserSentinel` recognises the documented values (incl. `TRUE`, `  none  `) and rejects real handlers (`/usr/bin/firefox`, `Google Chrome`, `open`).

`bun test` is green (1167 pass / 0 fail). `bun run typecheck` is clean.

Refs #154 (BROWSER-sentinel cause; other code paths in that issue may still apply).

Fixes #724

### DIFF
--- a/packages/server/browser.test.ts
+++ b/packages/server/browser.test.ts
@@ -1,5 +1,5 @@
 import { afterEach, describe, expect, test } from "bun:test";
-import { shouldTryRemoteBrowserFallback } from "./browser";
+import { isNoOpBrowserSentinel, shouldTryRemoteBrowserFallback } from "./browser";
 
 const savedEnv: Record<string, string | undefined> = {};
 const envKeys = ["PLANNOTATOR_BROWSER", "BROWSER"];
@@ -42,5 +42,36 @@ describe("shouldTryRemoteBrowserFallback", () => {
     clearEnv();
     process.env.PLANNOTATOR_BROWSER = "/usr/bin/browser";
     expect(shouldTryRemoteBrowserFallback(true)).toBe(false);
+  });
+
+  test("true for remote sessions when BROWSER is a no-op sentinel (e.g. agent view)", () => {
+    clearEnv();
+    process.env.BROWSER = "true";
+    expect(shouldTryRemoteBrowserFallback(true)).toBe(true);
+  });
+
+  test("true for remote sessions when PLANNOTATOR_BROWSER is a no-op sentinel", () => {
+    clearEnv();
+    process.env.PLANNOTATOR_BROWSER = "none";
+    expect(shouldTryRemoteBrowserFallback(true)).toBe(true);
+  });
+});
+
+describe("isNoOpBrowserSentinel", () => {
+  test("returns false for undefined / empty", () => {
+    expect(isNoOpBrowserSentinel(undefined)).toBe(false);
+    expect(isNoOpBrowserSentinel("")).toBe(false);
+  });
+
+  test("recognises the documented no-op values, case- and whitespace-insensitive", () => {
+    for (const v of ["true", "false", "none", ":", "0", "1", "TRUE", "  none  "]) {
+      expect(isNoOpBrowserSentinel(v)).toBe(true);
+    }
+  });
+
+  test("does not flag real browser handlers", () => {
+    expect(isNoOpBrowserSentinel("/usr/bin/firefox")).toBe(false);
+    expect(isNoOpBrowserSentinel("Google Chrome")).toBe(false);
+    expect(isNoOpBrowserSentinel("open")).toBe(false);
   });
 });
--- a/packages/server/browser.ts
+++ b/packages/server/browser.ts
@@ -10,6 +10,20 @@ import fs from "node:fs";
 const IPC_REGISTRY = path.join(os.homedir(), ".plannotator", "vscode-ipc.json");
 
 /**
+ * Common "no-op" values for $BROWSER used by headless/background environments
+ * (e.g. Claude Code's agent view sets BROWSER=true) to signal "do not actually
+ * launch a browser". Treating these as if the variable were unset prevents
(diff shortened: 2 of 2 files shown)


backnotprop:
Is this the fix for `Claude Code's agent view` ? Trying it now

backnotprop:
great work

PR #612, opened 2026-04-25
Adds XDG Base Directory support (backward compatible)

Joao-O-Santos:
# Add XDG Base Directory support (backward compatible)

First off, let me thank you for building this plugin and open sourcing it. I'm still just trying out the plugin, but I'm already seeing the benefits!

## PR Rationale

I like to keep my \$HOME directory as clean as possible, and have configured XDG_ env vars to help keep it that way. It is just a personal nit pick, but I wanted to get plannotator to respect XDG spec.

## What this PR does

This PR adds support for the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/latest/) while keeping `~/.plannotator` as the default for full backward compatibility.

### Before this PR

All plannotator data lived in a single `~/.plannotator/` directory:
- Config (`config.json`)
- Data (`plans/`, `history/`, `hooks/`)
- Cache/transient files (`drafts/`, `sessions/`, `pastes/`, debug logs,
  schemas)

### After this PR

Path resolution now follows this priority (highest wins):
1. `PLANNOTATOR_*_DIR` env var (explicit override)
2. `XDG_*_HOME` env var (spec compliance)
3. `~/.plannotator` (original default — unchanged)

| Category | Contents | Env override | XDG fallback | Default |
|----------|----------|--------------|--------------|---------| Config |
|`config.json` | `PLANNOTATOR_CONFIG_DIR` | `XDG_CONFIG_HOME` |
|`~/.plannotator` | Data | `plans/`, `history/`, `hooks/` |
|`PLANNOTATOR_DATA_DIR` | `XDG_DATA_HOME` | `~/.plannotator` | Cache |
|`drafts/`, `sessions/`, `pastes/`, logs, schemas | `PLANNOTATOR_CACHE_DIR` |
|`XDG_CACHE_HOME` | `~/.plannotator` |

### Migration strategy

- **Reads**: new resolved path first, then fallback to `~/.plannotator/`
- **Writes**: always go to the resolved path
- **No forced migration**: old data is never moved or deleted
- **Indefinite legacy support**: existing users see zero change unless they set
XDG vars

### Files changed

- `packages/shared/paths.ts` — new path resolution module
- `packages/shared/config.ts`, `storage.ts`, `draft.ts`, `improvement-hooks.ts`
— use new paths
- `packages/server/sessions.ts`, `browser.ts`, `codex-review.ts`, `tour/tour-review.ts` — use new paths
- `apps/vscode-extension/src/extension.ts` — match server's IPC path
[...]

PR #568, opened 2026-04-15
feat(paths): use XDG base directories instead of ~/.plannotator

cwrau:
## Summary

- Adds `packages/shared/paths.ts` with three helpers: `getDataBase()`, `getStateBase()`, `getConfigBase()`
- Stores persistent data (`plans/`, `history/`) under `$XDG_DATA_HOME/plannotator` (default: `~/.local/share/plannotator`)
- Stores ephemeral state (`drafts/`, `sessions/`, debug logs) under `$XDG_STATE_HOME/plannotator` (default: `~/.local/state/plannotator`)
- Stores configuration (`config.json`, improvement hooks) under `$XDG_CONFIG_HOME/plannotator` (default: `~/.config/plannotator`)
- **Existing installations are unaffected**: if `~/.plannotator` already exists on disk, all three helpers return it — zero migration required

## Call sites updated

| File | Change |
|------|--------|
| `packages/shared/storage.ts` | `getDataBase()` for plans + history |
| `packages/shared/draft.ts` | `getStateBase()` for drafts |
| `packages/shared/config.ts` | `getConfigBase()` for config.json |
| `packages/shared/improvement-hooks.ts` | `getConfigBase()` for hooks dir |
| `packages/server/sessions.ts` | `getStateBase()` for sessions |
| `packages/server/codex-review.ts` | `getStateBase()` for schema file + debug log |

## Migration for existing users

No action needed — `~/.plannotator` continues to work unchanged. To adopt XDG paths, simply move the directory:

```bash
mv ~/.plannotator ~/.local/share/plannotator   # data + state
# or symlink: ln -s ~/.local/share/plannotator ~/.plannotator
```

Config and state can be split further if desired:
```bash
mkdir -p ~/.config/plannotator ~/.local/state/plannotator
mv ~/.local/share/plannotator/config.json ~/.config/plannotator/
mv ~/.local/share/plannotator/hooks      ~/.config/plannotator/
mv ~/.local/share/plannotator/drafts     ~/.local/state/plannotator/
mv ~/.local/share/plannotator/sessions   ~/.local/state/plannotator/
```

## Test plan

- [ ] Fresh install: verify data lands in `~/.local/share/plannotator/`, state in `~/.local/state/plannotator/`, config in `~/.config/plannotator/`
- [ ] With `~/.plannotator` present: verify all paths still resolve to `~/.plannotator` (no regression)
[...]

PR #795, opened 2026-05-26
feat: add PLANNOTATOR_DATA_DIR env var to customize data directory

IstPlayer:
## Summary

Add a `PLANNOTATOR_DATA_DIR` environment variable that allows users to relocate the Plannotator data directory (default: `~/.plannotator`) to any path they prefer.

## Motivation

Plannotator currently hardcodes `~/.plannotator/` as the base data directory in multiple places — `storage.ts`, `config.ts`, `draft.ts`, `improvement-hooks.ts`, `sessions.ts`, `codex-review.ts`, `tour-review.ts`, `browser.ts`, and `pr-gitlab.ts`. This clutters the home directory and conflicts with XDG conventions (`XDG_DATA_HOME`, `XDG_CACHE_HOME`, etc.) that many users follow to keep their home directories clean.

The `PASTE_DATA_DIR` env var already exists as precedent for customizing sub-paths. This PR extends that pattern to the entire data directory.

## Changes

- **New file**: `packages/shared/data-dir.ts` — shared helper `getPlannotatorDataDir()` that checks `PLANNOTATOR_DATA_DIR` first (with `~` expansion), falls back to `~/.plannotator`
- **9 source files updated**: all hardcoded `join(homedir(), ".plannotator", ...)` calls replaced with `getPlannotatorDataDir()`
- Unused `homedir` imports removed where possible

## Usage

```bash
# Move all Plannotator data to ~/.cache
export PLANNOTATOR_DATA_DIR=~/.cache/plannotator

# Or to an XDG-compliant location
export PLANNOTATOR_DATA_DIR=$XDG_DATA_HOME/plannotator
```

## Backward Compatibility

Default behavior is unchanged — when `PLANNOTATOR_DATA_DIR` is absent or empty, the data directory remains `~/.plannotator`.


backnotprop:
Thanks for the PR @IstPlayer — clean approach and the right default behavior (nothing changes unless you explicitly set the env var).

I pushed a fix commit (`46df9f7a`) addressing the two files that still had hardcoded `~/.plannotator` paths:

- **`apps/opencode-plugin/index.ts`** — `getPlanBackingPath()` now uses `getPlannotatorDataDir()` instead of `homedir() + ".plannotator"`. Without this, OpenCode's active plan backing file would ignore `PLANNOTATOR_DATA_DIR`.
- **`apps/vscode-extension/src/extension.ts`** — The `IPC_REGISTRY` path now resolves through the same env var logic. The VS Code extension is a standalone build (can't import from `@plannotator/shared`), so the resolution logic is inlined. Without this fix, the extension and server would disagree on the IPC registry location when `PLANNOTATOR_DATA_DIR` is set.
- **`packages/shared/package.json`** — Added the `"./data-dir"` export so the OpenCode plugin can import it as `@plannotator/shared/data-dir`.
- **`AGENTS.md`** — Documented `PLANNOTATOR_DATA_DIR` in the env var table.

---

**For context:** There are two other open PRs tackling the same problem with different approaches:

- **#568** (@cwrau) — Full XDG Base Directory support with `getDataBase()`/`getStateBase()`/`getConfigBase()` and an `existsSync(~/.plannotator)` legacy guard. Gets the XDG categories right (state vs cache) but adds complexity.
- **#612** (@Joao-O-Santos) — Kitchen-sink approach with `PLANNOTATOR_*_DIR` + XDG + legacy fallback. Most comprehensive but has some issues with inline path resolution in paste service/VS Code defaulting to wrong paths.

This PR (#795) is the simplest and safest of the three — one env var, opt-in only, zero risk for anyone who doesn't set it. The co-author trailers on the fix commit credit @cwrau and @Joao-O-Santos since their work helped identify the full scope of files that needed updating.

PR #805, opened 2026-05-28
Fix Pi plan approval continuation delivery

### LINKED ISSUE #804: Pi extension errors after approving a plan
## Bug

When a user approves a Plannotator plan in the Pi extension, the extension can emit:

```
Extension "<runtime>" error: Agent is already processing. Specify streamingBehavior ('steer' or 'followUp') to queue the message.
```

## Cause

The approve path returns `terminate: true`, then the `agent_end` handler sends the continuation prompt with:

```ts
pi.sendUserMessage("Continue with the approved plan.");
```

Because the Pi runtime may still consider the agent turn active immediately after termination, `sendUserMessage` needs an explicit queued delivery mode.

The deny path does not hit this because it returns normally and does not set `justApprovedPlan`, so the `agent_end` handler does not send the continuation prompt.

## Expected fix

Send the continuation with `{ deliverAs: "followUp" }` and remove the `setTimeout(0)` workaround.

xz-dev:
## Summary

Fixes #804.

- Queue the Pi plan approval continuation as a follow-up message.
- Remove the unreliable `setTimeout(0)` workaround.
- Align the approval continuation with existing review/annotation feedback delivery.

## Why

Approved plans return `terminate: true`, so Pi may still consider the agent turn active when `agent_end` fires. Calling `pi.sendUserMessage()` without a delivery mode can therefore fail with `Agent is already processing`. `deliverAs: "followUp"` tells Pi to queue the continuation safely.

## Testing

- `git diff --check`
- Not run: `bun run typecheck` (Bun is not installed in this environment: `bun: command not found`)

### DIFF
--- a/apps/pi-extension/index.ts
+++ b/apps/pi-extension/index.ts
@@ -1165,9 +1165,9 @@ Execute each step in order. After completing a step, include [DONE:n] in your re
 	pi.on("agent_end", async (_event, ctx) => {
 		if (phase === "executing" && justApprovedPlan) {
 			justApprovedPlan = false;
-			setTimeout(() => {
-				pi.sendUserMessage("Continue with the approved plan.");
-			}, 0);
+			pi.sendUserMessage("Continue with the approved plan.", {
+				deliverAs: "followUp",
+			});
 			return;
 		}
 


backnotprop:
nice work, thanks

PR #809, opened 2026-05-28
annotate-last: pick which message to annotate (fixes #800)

### LINKED ISSUE #800: annotate-last: picks wrong message after /rewind; consider a message-picker UI
## Problem

`plannotator last` (and the `/plannotator-last` slash command) **selects the last assistant text block by position in the session transcript**. When the conversation has been rewound—e.g. the user ran `/rewind` in Claude Code—the message displayed at the bottom of the chat from the user's perspective is not the same as what's newest in the harness's on-disk transcript, which still contains the orphaned pre-rewind messages.

**The result is that Plannotator opens on a message the user intended to backtrack on and there's no affordance to correct this.**

This is similar to #739 in that the wrong message is selected, but the rewind case isn't covered by that fix and there's no existing issue tracking it.

## Repro (Claude Code)

1. Have a multi-turn conversation.
2. `/rewind` back one or more assistant responses.
3. Run `/plannotator-last`.
4. Plannotator opens on the most recent transcript entry, which after a rewind is not the message the user likely intends to annotate.

## Why it happens

Per the [annotate-last docs](https://plannotator.ai/docs/commands/annotate-last/), all four supported harnesses (Claude Code, OpenCode, Pi, Codex) select "the last assistant text block(s)" from the session, with filtering for system/noise entries but no awareness of conversation-mutation commands like `/rewind` or `/clear`. The current input contract has no flag for "which message" — only `--gate`, `--json`, `--hook`.

## Proposed fix

Show a dropdown or numbered list in the Plannotator `last` UI with recent assistant messages, ordered newest first, defaulting to the current "last" selection. The user picks which one to annotate. Each entry shows a short preview (first few dozen chars + timestamp, perhaps) so it's easy to find the intended message.

This is a strict improvement over the current behavior:
- Solves the rewind case naturally — the user just picks the message they actually see in their terminal.
- Provides general affordance to annotate any earlier message without manual transcript hunting.
- Backwards-compatible: if the user doesn't interact with the picker, the default selection matches today's behavior.

All four supported harnesses already expose ordered message lists (jsonl files for Claude Code/Codex, SDK calls for OpenCode/Pi), so the picker data source is available without per-harness work beyond message extraction (which already exists).

Optional companion: a `--index N` or `--message-id <id>` flag for scripted/hooked use cases — useful if `annotate-last` is ever wired into automation.

## Contributing

Happy to send a PR if you agree the picker UI is the right direction. Wanted to check first since you are likely inundated with slop PRs.

GLips:
Fixes #800.

## What this changes for users

Running \`/plannotator-last\` after \`/rewind\` no longer silently opens on the orphaned pre-rewind message. The annotation UI now surfaces recent messages so you can pick whichever message you like.

Two affordances, both only visible when there's more than one message to choose from:

- A **"Message N of M"** button in the document's sticky-top action bar (alongside Copy / Global comment / Attachments). Stays visible while scrolling.
- A **"Messages" tab** in the left sidebar with the full list — newest-first, preview + timestamp, default marked with ★. Mirrors the existing Files / Versions / Archive tab pattern.

Default selection (index 0) matches today's "last message" behavior, so anyone who doesn't interact with the picker sees no change.

CleanShot 2026-06-01 at 12 00 58@2x

## Coverage

All six harnesses (Claude Code, Codex, Droid, Copilot, OpenCode, Pi) ship the message picker.

I've tested and confirmed with Claude Code, which should cover CC, Codex, and Droid since they exercise the same message paths. Direct testing with OpenCode, Pi, and Copilot CLI still need to happen.

## Notes for reviewers

- The wire payload (\`recentMessages\` field on \`/api/plan\`) is omitted entirely when there's nothing to pick, so unsupported harnesses degrade cleanly with zero conditional branches in the server.
- Switching messages remounts the Viewer via a shared viewerContentKey (also feeds StickyHeaderLane's remount token). This is required because web-highlighter mutates the content DOM, which would otherwise crash React's reconciler (removeChild) on swap.
- New \`MessagesIcon\` lives in \`packages/ui/components/icons/\` — the grouped-icon convention.
- Added 6 unit tests covering \`extractRecentRenderedMessages\` (turn-boundary crossing, limit honoring, default-equals-legacy invariant, edge cases). Full suite still 1228/1228.

## Test plan

- [x] Verify default selection matches pre-PR \`/plannotator-last\` behavior in a non-rewound session
- [x] Run \`/rewind\` then \`/plannotator-last\` — confirm the picker surfaces both the orphaned and current-branch messages
- [ ] Confirm OpenCode / Pi / Copilot sessions still work (no picker, original message shown)


GLips:
Ok I gotta go through some more thorough testing myself tomorrow with Claude Code but this should be pretty close to the final form if you want to take a look @backnotprop

If not I'll explicitly mark it as ready for review once I've had a chance to check it out more deeply.

GLips:
@backnotprop ready for review, think this is good to go :D

backnotprop:
nice work will get through the review by later today

backnotprop:

PR #814, opened 2026-05-28
Support GitLab: custom-domain detection + MR list/detailed

backnotprop:
## What

Makes Plannotator's project dashboard actually work for GitLab users. Two holes that are useless apart, shipped together:

1. **Platform detection** — self-hosted GitLab on a custom domain (e.g. `code.company.com`) was misdetected as GitHub, so the daemon ran `gh` against a GitLab server and failed silently.
2. **MR list/detailed** — `fetchPRList` / `fetchPRDetailedList` returned `[]` for GitLab, so the dashboard showed "No pull requests found" even with open MRs.

## Changes

**Detection** (`packages/shared/pr-provider.ts`, `packages/server/pr.ts`, `packages/server/daemon/server.ts`)
- New `detectPlatformCore(runtime, host)` in `pr-provider.ts`, bound as `detectPlatform(host)` in `server/pr.ts` — mirrors the existing `checkAuthCore` → `checkPRAuth` pattern.
- Layered cheap → expensive: host-name fast path (`gitlab`/`github` substrings, zero I/O), then for ambiguous custom domains a single `glab auth status --hostname <host>` probe whose success means GitLab.
- `glab`-absent (ENOENT) and `glab`-unauthed both fall through to the historical github default — **no regression** for any host that worked before.
- The daemon `/daemon/projects/prs[/detailed]` endpoint now calls `await detectPlatform(host)` instead of `host.includes("gitlab")`. The existing per-cwd 30s cache means the probe runs at most once per project per window.

**MR list** (`packages/shared/pr-gitlab.ts`, `packages/shared/pr-provider.ts`)
- New `fetchGlMRList` / `fetchGlMRDetailedList` against `projects/:id/merge_requests?per_page=30&state=all`, using the existing `apiArgs` (handles `--hostname` for self-hosted) and `parsePaginatedArray` helpers.
- JSON→item mapping extracted into **exported pure functions** `mapGlMrToListItem` / `mapGlMrToDetailedItem` so they're unit-testable without spawning `glab`. Field mapping: `iid`→number/id, `author.username`→author, `web_url`→url, `source_branch`→headBranch, `target_branch`→baseBranch; state `opened`→open, `merged`→merged, `closed`/`locked`→closed.
- `pr-provider.ts` dispatches GitLab to these instead of returning `[]`.

### Deliberate GitLab limitation (confirmed tradeoff)
GitLab's MR-**list** endpoint does **not** return per-MR `additions`/`deletions`. Fetching them needs one extra API call per MR (~30 for the dashboard), which is unacceptable for a list view. Detailed items therefore set `additions: 0, deletions: 0` and `reviewDecision: ""` (GitLab approvals are a separate endpoint, out of scope for v1). Everything else populates: `commentCount` (from `user_notes_count`), `updatedAt`, `isDraft` (`draft` or legacy `work_in_progress`), branches, and state.

## Verification

- `bun test` (root): **1345 pass, 0 fail**. New `pr-gitlab.test.ts` cases cover both mappers against a hand-written `merge_requests` JSON fixture (opened/merged/closed/locked/draft, null author, falsy-zero `user_notes_count`, legacy `work_in_progress`) and `detectPlatformCore` across `github.com`, `gitlab.com`, `gitlab.example.com`, `github.acme.com`, and an ambiguous `code.company.com` (probe success → gitlab; auth-fail → github; ENOENT → github), asserting the fast paths never shell out.
- `bun run typecheck`: clean across all five tsconfigs (shared, ai, server, ui, pi-extension).
- The GitHub path is untouched — every `github` branch of every dispatcher is unchanged.

## Deferred

- **Live GitLab smoke test skipped**: `glab` is not installed/authed on this machine, so the optional `curl /daemon/projects/prs?cwd=<gitlab repo>` check was not run. Unit tests are the gate per the task spec.
- `additions`/`deletions` left at `0` for GitLab detailed items (limitation above).
- GitLab `reviewDecision` (approvals) not fetched — separate endpoint, out of scope for v1.

### DIFF
--- a/apps/frontend/src/components/landing/LandingPage.tsx
+++ b/apps/frontend/src/components/landing/LandingPage.tsx
@@ -630,9 +630,11 @@ function PRList({
       </div>
     );
   }
-  if (platform === "gitlab" && prs.length === 0) {
+  if (error === "fetch-failed") {
     return (
-      <div className="py-1 text-[11px] text-muted-foreground">GitLab MR listing coming soon</div>
+      <div className="py-1 text-[11px] text-muted-foreground">
+        Failed to load {platform === "gitlab" ? "merge requests" : "pull requests"}
+      </div>
     );
   }
   if (visible.length === 0 && !showAll) {
--- a/apps/frontend/src/components/landing/git-dashboard/PRRow.tsx
+++ b/apps/frontend/src/components/landing/git-dashboard/PRRow.tsx
@@ -68,10 +68,16 @@ export function PRRow({ pr, loading, onSelect }: PRRowProps) {
             {reviewBadge.label}
           </span>
         )}
-        <span className="flex items-center gap-1.5 font-mono text-[10px]">
-          <span className="tabular-nums text-green-600 dark:text-green-400">+{pr.additions}</span>
-          <span className="tabular-nums text-red-600 dark:text-red-400">-{pr.deletions}</span>
-        </span>
+        {pr.platform === "gitlab" ? (
+          // GitLab's MR-list endpoint omits per-MR additions/deletions, so render
+          // an em dash ("unknown") rather than a misleading +0/-0.
+          <span className="font-mono text-[10px] text-muted-foreground/50">—</span>
+        ) : (
+          <span className="flex items-center gap-1.5 font-mono text-[10px]">
+            <span className="tabular-nums text-green-600 dark:text-green-400">+{pr.additions}</span>
+            <span className="tabular-nums text-red-600 dark:text-red-400">-{pr.deletions}</span>
+          </span>
+        )}
         {pr.commentCount > 0 && (
           <span className="flex items-center gap-0.5 text-[10px] text-muted-foreground/60">
             <svg
--- a/apps/frontend/src/daemon/contracts.ts
+++ b/apps/frontend/src/daemon/contracts.ts
@@ -93,7 +93,7 @@ export interface PRDetailedListResponse {
   ok: true;
   prs: PRDetailedListItem[];
   platform: "github" | "gitlab" | null;
-  error?: "no-remote" | "no-cli" | "auth-failed";
+  error?: "no-remote" | "no-cli" | "auth-failed" | "fetch-failed";
   message?: string;
 }
 
@@ -102,7 +102,7 @@ export interface PRListResponse {
   prs: PRListItem[];
   platform: "github" | "gitlab" | null;
   defaultBranch?: string;
-  error?: "no-remote" | "no-cli" | "auth-failed";
+  error?: "no-remote" | "no-cli" | "auth-failed" | "fetch-failed";
   message?: string;
 }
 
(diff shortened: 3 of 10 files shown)


backnotprop:
## Fix pass (4-model adversarial review)

Pushed `7c88fd82` to `feat/gitlab-pr-support`.

**Fix 1 — Stop silent-empty on fetch failure (unanimous Act-On).** `fetchGlMRList`/`fetchGlMRDetailedList` no longer `return []` on a non-zero `glab` exit — a 403/network/rate-limit error was indistinguishable from "zero MRs." Both now `console.error(stderr)` and **throw** (via a shared `fetchGlMRRaw()` helper). The daemon `/daemon/projects/prs[/detailed]` catch now returns `error: "fetch-failed"` + `message` instead of a bare empty list. Surfaced in the dashboard (`git-dashboard-store.ts`) and the per-project `PRList` (`LandingPage.tsx`) reusing the existing `no-cli`/`auth-failed` display pattern ("Failed to load merge/pull requests"). GitHub benefits from the same surfacing.

**Fix 2 — Remove stale "GitLab MR listing coming soon" placeholder (Act-On).** Replaced with fall-through to the shared zero-PR empty state. Edit is confined to the placeholder region (~L633); the `buildStacks` import/call site is untouched to keep #815 merging in disjoint hunks.

**Fix 3 — Cache platform per host + bound the probe (unanimous Act-On).** `detectPlatform` (the Bun wrapper in `pr.ts`) now caches `host → platform` for the daemon lifetime, so the `glab auth status` probe runs at most once per host instead of up to ~4×/cold-load. `PRRuntime.runCommand` has no timeout option, so the probe is bounded with a 5s `Promise.race` that resolves to the historical `github` default — a slow/unreachable self-hosted host can no longer hang the dashboard. Detection logic itself is unchanged; known-host fast paths resolve synchronously and always win the race.

**Fix 4 — `+0 −0` renders as wrong data (Consider).** `PRRow` now renders an em dash ("unknown") for GitLab, whose MR-list endpoint omits per-MR additions/deletions. `platform` is threaded onto `GitDashboardPR`.

**Fix 5 — Dedup + fetch test (Consider).** The two identical `glab api` calls are extracted into `fetchGlMRRaw`. Added fake-`PRRuntime` tests asserting the exact `glab` args, `--hostname` for self-hosted, and that a non-zero exit now throws (per Fix 1).

### Verification
- `bun test`: GitLab suite 28/28 pass; root suite green except 2 pre-existing failures unrelated to this PR (missing `apps/pi-extension/generated/plugin-client.js` build artifact in a fresh worktree — confirmed identical on the clean base).
- Root `bun run typecheck` (shared/ai/server/ui/pi-extension): clean. `apps/frontend` typecheck: clean except one pre-existing error in `vitest.browser.config.ts` (missing optional `@vitest/browser-playwright` dep — identical on clean base). `apps/frontend` `bun run build`: succeeds.
- Live `glab` smoke test **unavailable — glab is not installed** in this environment.

### Notable behavior change
`/api/pr-list` in `review.ts` shares the same fetch path; on a `glab` failure it now returns a 500 ("Failed to fetch PR list") instead of a silently-empty 200. The existing `try/catch` there already handles the throw (no crash), and the 500 is more honest for the in-review PR-switcher — consistent with Fix 1's intent.

### Follow-up (out of scope for this pass)
Deeper GHE-vs-GitLab host verification beyond the credential probe was deliberately deferred. Fix 1's error surfacing already de-opaques the misclassification case (a wrong platform guess now produces a visible `fetch-failed` rather than a silent empty list).

Issue #91, opened 2026-01-18
Proposal for IDE-Based Control and Interactive Code Revision


alpgul:
I believe it would be much more efficient to implement the control mechanism through an IDE extension. This structure would allow for leaving comments directly on the additions, deletions, or modifications made by the AI, enabling the code to be re-coded or revised based on those comments in the future.

This way, developers can iterate quickly within the editor without needing a browser. Constant context switching between the browser and the IDE would significantly hinder the development process compared to the speed and fluidity of working directly within the IDE.

backnotprop:
@alpgul thanks

Do you mean for the entire experience.

Or just the code review feature, `plannotator-review`

alpgul:
Yes, I mean review feature. I believe performing checks via an IDE extension would be more efficient than a browser-based solution. Most review tools I have experienced operate individually through manual triggers. With the bulk commenting feature used in this tool, we can add multiple comments and create a more consistent code structure.

alpgul:
Let me provide a step-by-step example scenario for better understanding:

1. Suppose you have a plan that modifies the wrong files and code.
2. In this plan, we open a global or task-specific comment.
3. We then select a code block (from project files or from uncommitted code change files) within the IDE and choose the "create comment" option.
4. We add this code block and the associated comment as a sub-comment within the main task comment.
5. This approach enables more reliable and faster code production.
6. This way, while editing the plan, we would be doing so with the support of code blocks.
7. Of course, before implementing this, a plugin needs to be developed for the IDE to manage these controls.
8. This feature can be added at a later stage.

You can think of this as the popup prompt feature with code transformation capabilities found in IDEs. However, while IDEs execute these manually and individually, we run them in bulk here, which will ensure higher code consistency.

7tg:
I built a VS Code extension that addresses part of this — it opens Plannotator plan reviews inside VS Code tabs instead of an external browser, so you don't have to context-switch.

It also persists your settings across sessions and auto-closes the panel when you're done reviewing.

- **Marketplace:** https://marketplace.visualstudio.com/items?itemName=7tg.plannotator-webview
- **Source:** https://github.com/7tg/plannotator-vscode

Happy to transfer ownership of the extension/repo if you'd like to maintain it under the plannotator org.

jedzill4:
Any update on this ? it would be very helpful!

backnotprop:
I'll take a look at what @7tg built and try to pull it in with his commits. I can then release the VS Code extension around it. 

[...]

Issue #89, opened 2026-01-18
OpenCode < 1.1.10 CVE


Yair-Ziv:
Hey there!
Plannotator uses opencode as a dependency.
Opencode has [CVE](https://cy.md/opencode-rce/) on versions earlier than 1.1.10, plannotator currently enforces
```
"apps/opencode-plugin": {
      "name": "@plannotator/opencode",
      "version": "0.6.4",
      "dependencies": {
        "@opencode-ai/plugin": "^1.0.218",
      },
      "peerDependencies": {
        "bun": ">=1.0.0",
      },
    },
```
Could this be updated? 
Love the plugin

backnotprop:
## Fix in PR #90

@Yair-Ziv Thanks for reporting! 

Updated `@opencode-ai/plugin` dependency from `^1.0.218` to `^1.1.10` in PR #90.

Available for testing now:
```bash
rm -rf ~/.bun/install/cache/@plannotator

# opencode config:
{
  "plugin": ["@plannotator/opencode@beta"]
}
```

Will be in the next release.

Issue #67, opened 2026-01-13
Feature Request: Image names for referencing in feedback


luketych:
I want to be able to attach images as feedback, and then reference them in an annotation so that when the LLM is reading my annotation it can know to check the image for further feedback.

backnotprop:
This feature has been implemented in #147. Images now carry human-readable names (`ImageAttachment` with `{path, name}`) throughout the pipeline — editable inline under thumbnails, with auto-derived names from original filenames. Exported feedback uses `[name] /tmp/path...` format.

Will be merged and part of the next release, likely tomorrow.

Issue #66, opened 2026-01-13
Download fails in the Windows 10


chocomilk4759:
irm https://plannotator.ai/install.ps1 | iex
Fetching latest version...
Installing plannotator v0.5.7...
iex: Could not call the method named ‘Split’ on [System.Byte].
Location line:1 character:42
+ irm https://plannotator.ai/install.ps1 | iex
+                                          ~~~
    + CategoryInfo          : InvalidOperation: (:) [Invoke-Expression], RuntimeException
    + FullyQualifiedErrorId : MethodNotFound,Microsoft.PowerShell.Commands.InvokeExpressionCommand

:(

backnotprop:
@chocomilk4759 I just merged a fix and deployed the windows script. Can you try once more?

I had to invalidate the CloudFront cache. Maybe you got the old version. 

Literally just merged this. https://github.com/backnotprop/plannotator/pull/65

chocomilk4759:
k, i'll try when updated. thx

backnotprop:
Try now @chocomilk4759

chocomilk4759:
@backnotprop 
Yes, it downloaded successfully.

chocomilk4759:
@backnotprop 
hi, again me 😅


After completely removing all previously installed items and reinstalling, I only added the marketplace, but it shows as already installed. What should I do?

Try:
/plugin marketplace add backnotprop/plannotator
/plugin install plannotator@plannotator


<img width="629" height="176" alt="Image" src="https://github.com/user-attachments/assets/5fa26bb6-0070-4cfb-addf-67af29be81f4" />

<img width="583" height="163" alt="Image" src="https://github.com/user-attachments/assets/4521e849-e4d7-4005-89a4-6564e009e4e6" />

backnotprop:
@chocomilk4759 you should be good. Do you want to get on a google meet to debug?

backnotprop:
Oh, but yeah, it looks like the plug-in's still not installed. 

```
/plugin marketplace add backnotprop/plannotator
/plugin install plannotator@plannotator // <--- Run this one now
```

chocomilk4759:
It's not easy to meet because I'm at work.


and

That didn't hit

> /plugin install plannotator@plannotator
  ⎿  Plugin 'plannotator@plannotator' is already installed. Use '/plugin' to manage existing plugins.

backnotprop:
Can you restart claude

chocomilk4759:
[...]

PR #815, opened 2026-05-28
Fix order-dependent PR-stack grouping via leaf-rooting

backnotprop:
## The bug

`buildStacks` (project dashboard PR grouping in `LandingPage.tsx`) walked PR chains **only downward** (toward the base) and marked each PR consumed as it went. A descendant discovered *after* its parent chain was already built could not attach and was dumped into `loose`.

Concretely, a 3-deep stack `A(base=main) ← B(base=A) ← C(base=B)` grouped correctly **only** when the input arrived leaf-first (`[C, B, A]`). For base-first (`[A, B, C]`) or interleaved orderings, `B` was consumed into `[A, B]` first; then `C` walked down, hit already-consumed `B`, had length 1, and orphaned into `loose`. The grouping depended on the order the API returned PRs — which is wrong.

## The fix

Root every chain from a **leaf** — a PR whose `headBranch` is not any other PR's `baseBranch` — then walk **down** from each leaf following `baseBranch ← headBranch` edges. A leaf-rooted walk captures the full chain in **one pass regardless of input order**.

Output shape is unchanged:
- `stacks`: chains of length > 1, ordered base → leaf, labelled `#<base> → #<leaf>`.
- `loose`: everything else, in original input order.

The base PR (whose base is the default branch) is still included via the walk. Cycles fall through to `loose` (their members are never leaves, and the walk guard prevents infinite loops). Forks (a branch that is the base of two PRs) degrade to one stack plus a loose sibling without double-counting the shared ancestor.

To make it testable, `buildStacks` and the `PRStack` interface were extracted into a new pure module `apps/frontend/src/components/landing/buildStacks.ts` (no React imports). `LandingPage.tsx` now imports both; the call-site signature is unchanged.

## Test coverage

New table-driven vitest suite `apps/frontend/src/components/landing/buildStacks.test.ts`:
- 3-deep stack given leaf-first, base-first, and interleaved orderings → all produce the same single `#1 → #3` stack with 3 PRs.
- 4-deep stack (multiple orderings).
- Two independent stacks in one input.
- All-loose (every PR based on the default branch).
- A single non-default-based PR with no parent in the set → loose.
- Empty input.
- **Permutation-invariance:** every permutation of a 3-deep stack, a 4-deep stack, and two independent stacks yields identical grouping.
- Cycle (does not loop, lands in loose) and fork (no double-counted base) edge cases.

The suite was confirmed to **fail against the old downward-only algorithm** (12 failures, including the base-first 3-deep case) and **pass against the fix** (18 passing). A separate parity check over 2000 random leaf-first forests confirmed the new algorithm agrees with the old one wherever the old one was correct, i.e. the fix is a strict generalization.

## Verification

In `apps/frontend`:
- `bun run test` → 18 passed.
- `bun run typecheck` → exit 0 (the only diagnostic is a pre-existing `@vitest/browser-playwright` module-resolution error in `vitest.browser.config.ts`, present on the base branch and unrelated to this change).

Also marked the "PR stack splitting is order-dependent" backlog entry as DONE in `goals/frontend-session-lifecycle/backlog.md`.

### DIFF
--- a/apps/frontend/src/components/landing/LandingPage.tsx
+++ b/apps/frontend/src/components/landing/LandingPage.tsx
@@ -562,10 +562,7 @@ function PRList({
     [prs, showAll],
   );
   const hiddenCount = prs.length - visible.length;
-  const { stacks, loose } = useMemo(
-    () => buildStacks(visible, defaultBranch),
-    [visible, defaultBranch],
-  );
+  const { stacks, loose } = useMemo(() => buildStacks(visible), [visible]);
 
   if (loading) {
     return <div className="py-1 text-[11px] text-muted-foreground">Loading PRs…</div>;
--- a/apps/frontend/src/components/landing/LandingPage.tsx
+++ b/apps/frontend/src/components/landing/LandingPage.tsx
@@ -20,6 +20,7 @@ import { GitDashboard } from "./git-dashboard/GitDashboard";
 import { useDaemonEventStore } from "../../daemon/events/event-store";
 import { daemonApiClient } from "../../daemon/api/client";
 import { getSessionModeMeta, formatSessionLabel } from "../../shared/session-meta";
+import { buildStacks, type PRStack } from "./buildStacks";
 import type {
   ProjectEntry,
   PRListItem,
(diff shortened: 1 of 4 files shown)


backnotprop:
Fix pass applied from the 4-model interrogate review. Pushed to `fix/pr-stack-grouping`.

**Fix 1 — drop dead `_defaultBranch` (UNANIMOUS).** Removed the unused param from `buildStacks` and from the only call site (`LandingPage` `useMemo` + its deps). This also kills the latent stale-default split bug: when the real default is `master` but the UI still holds its initial `"main"`, the old `useMemo` dep produced wrong groupings. Edits are confined to the `useMemo` hunk (lines 562-565) — disjoint from #814's GitLab placeholder/error region.

**Fix 2 — deterministic forks (UNANIMOUS).** Candidate leaves are now visited in ascending PR-`number` order, so when two PRs share a base, the lowest-numbered child deterministically claims the shared ancestor and the sibling falls through to `loose` — no longer order-dependent. The fork test now asserts this across **all** permutations of `[root, child1, child2]` (lowest child wins, other loose) instead of a hardcoded `toContain(72)`. Doc comment + backlog softened: this is an explicit "one child wins, rest loose" policy (not full fork collapse), but the *choice* is order-independent.

**Fix 3 — deterministic `byHead` (Consider).** Head-branch collisions (e.g. a merged + open PR on the same branch from `state=all`) now resolve to the open PR, then the lower number, via a `preferHead` tiebreak. Added a permutation test where a tip follows a duplicated `mid` head — always resolves `#89 → #91` with the merged duplicate loose.

**Fix 4 — stable output order (Consider).** `stacks` sorted by base PR `number`, `loose` by `number`, so independent stacks never swap on-screen between 30s polls. Permutation tests now assert the **unsorted** output arrays (removed the `sorted()`/`sortedNums()` wrappers that were masking order).

**Fix 5 — cycle test + loop tidy (Consider).** Added a leaf-into-cycle permutation test pinning current bounded behaviour (terminates via the `stacked` guard, forms a `#66 → #67` stack). Simplified the length-1 release to `stacked.delete(chain[0].id)`.

**Verification.**
- Meaningfulness proven: the rewritten fork, duplicate-head, and output-order tests all produce **2 distinct outcomes** against the old downward-only / last-write-wins algorithm (i.e. they FAIL the old code), and exactly **1** against the new one.
- Brute-forced all 5040 permutations of a mixed input (linear stack + fork + loose) → 1 distinct outcome.
- `apps/frontend`: `bun run test` 20/20 pass; `bun run typecheck` clean (the only error is the pre-existing `vitest.browser.config.ts` missing-dep, unrelated).
- Root `bun test`: my changes add passing assertions and introduce zero new failures (the 2 pre-existing fail / 2 errors live in `apps/pi-extension` / `remote.test.ts`, unrelated to stack grouping).

PR #838, opened 2026-06-02
fix(ui): align list markers to first line in clean diff view

rcdailey:
List markers (bullets and ordered numbers) render vertically centered against the full wrapped height of a list item in the plan-diff clean "Show Changes" view, instead of sitting on the first line. It only shows up on items long enough to soft-wrap, and it makes it hard to tell where each item begins.

The marker component (`ListMarker`) wraps its glyph in an inner `flex items-center`, which only renders correctly when the parent row pins its children to the top. The main viewer's row does that with `items-start`; the clean diff view's row class did not, so the marker stretched to the item's full height and `items-center` then centered the glyph.

The fix makes correctness intrinsic to the shared component: `ListMarker` now carries `self-start`, so it can't stretch regardless of the parent row's `align-items`. No call site has to remember the contract, and a future fourth consumer can't reintroduce the bug. The diff row also gets `items-start` for parity with the main viewer.

A second commit extracts the duplicated marker-plus-paragraph scaffold (which is why this fix would otherwise need to be applied in three places) into a `ListItemBody` component. Each call site keeps its own row wrapper (indent, data attributes, hover props, interactivity). It's a structural-only change with no behavior difference, so feel free to drop that commit if you'd rather not take the refactor; the fix stands on its own underneath it.

### DIFF
--- a/packages/ui/components/BlockRenderer.tsx
+++ b/packages/ui/components/BlockRenderer.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import { Block } from "../types";
 import { InlineMarkdown } from "./InlineMarkdown";
-import { ListMarker } from "./ListMarker";
+import { ListItemBody } from "./ListItemBody";
 import { CodeBlock } from "./blocks/CodeBlock";
 import { HtmlBlock } from "./blocks/HtmlBlock";
 import { Callout } from "./blocks/Callout";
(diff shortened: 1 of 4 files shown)


backnotprop:
thanks!

backnotprop:

PR #839, opened 2026-06-02
feat(setup-goal): opt-in "grill first" interview step in Phase 1

dgrissen2:
## What

Adds an **opt-in "grill first"** step to Phase 1 (Rearticulate) of the
`plannotator-setup-goal` skill. Before the compact interview bundle, the skill can run a
relentless, one-question-at-a-time interview that walks the decision tree and resolves
dependent choices, then folds what it learns forward into a sharper bundle (Phase 2) — or
straight into the fact sheet (Phase 3) when grilling fully resolves scope.

- **Suggest-first** when the goal is vague or has many interdependent decisions.
- **On-demand** any time the user says "grill me first".
- **Opt-in:** clear, well-scoped goals skip it, so it never fights the bundle's
  "fewer, higher-leverage questions" rule.

## Why we want it

The downstream pipeline (facts → plan → goal) is only as sound as the shared understanding
fixed in Phase 1. The interview bundle is deliberately low-friction and breadth-limited — its
questions are frozen upfront, so it structurally can't branch when one answer reshapes the
next. Grilling is the complementary depth tool: it resolves interdependent unknowns
conversationally before any facts are locked. Keeping it opt-in preserves the fast path for
clear goals while giving vague or high-stakes goals the rigor they need.

## Attribution

The grilling protocol is injected **verbatim** from the `/grill-me` skill by **Matt Pocock**
(`mattpocock/skills`, MIT-licensed), with a source-link comment kept inline in the skill:
https://github.com/mattpocock/skills/blob/main/skills/productivity/grill-me/SKILL.md

## Scope

- One file changed: `apps/skills/plannotator-setup-goal/SKILL.md` (+11 lines).
- Prompt-only — no binary/server changes, no new dependencies.

### DIFF
--- a/apps/skills/plannotator-setup-goal/SKILL.md
+++ b/apps/skills/plannotator-setup-goal/SKILL.md
@@ -23,6 +23,17 @@ Use `goals/<slug>/` for both working JSON files and final docs. The JSON files a
 
 **Browser session patience rule:** Plannotator goal setup is a user-driven browser session. After launching an interview or facts command, be absolutely patient and keep waiting on the user until they submit, dismiss, or explicitly ask you to stop. Do not close, kill, restart, refresh, or open a second copy just because the UI is idle or the user is taking time. Never close and reopen the session as a way to update state; if a rerun is needed after the prior session ends, update the working JSON file and launch a new command from that file.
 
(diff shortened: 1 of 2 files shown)


backnotprop:
@dgrissen2 do you intend for the grill session to be a chat-based, one-question-at-a-time prepass, not the bundled Plannotator Q&A UI.

(regarding the failing test)

dgrissen2:
one question at a time pre-pass, i think it would be too difficult with the bundled Q&A ui (which is nice, but if it's a shaky goal, I think requires more back and forth)

backnotprop:

PR #862, opened 2026-06-05
feat(review): Tripwires — passive alerts when a diff touches sensitive code

backnotprop:
## What

**Tripwires** let you declare slop-free zones — files and symbols that should never be casually touched. When a code-review diff crosses one, the hit surfaces **passively** in the review UI:

- amber glyph on the file in the file tree and file header
- compact amber marker in the diff gutter at the hit location (no comment card — code stays readable)
- entry in the annotation sidebar with the rule's note, click-to-jump

No banners, no modals, no approve-gating. Tripwires are **informational only**: excluded from feedback sent to the agent and from annotation count badges.

## Where rules live (two layers, global-first)

1. **Global (default)** — `~/.plannotator/tripwires/<project-key>.json`, private to your machine, auto-created empty on first review. Keyed by normalized remote URL so your clone, every worktree, and PR checkouts share one file (git-common-dir fallback for remote-less repos, incl. git <2.31). **Nothing is ever written into your repo.**
2. **Repo (team opt-in)** — a committed `.plannotator/tripwires.json` is honored if present and merged after global rules. Plannotator only creates it via the explicit `tripwires add --repo`.

```json
{
  "rules": [
    { "globs": ["src/billing/**"], "symbols": ["chargeCustomer"], "note": "Money path" }
  ]
}
```

**Any touch trips**: added/edited/deleted lines, renames, file deletions, binary changes, chmod-only changes. Symbol rules fire when the symbol appears in any changed line or its enclosing function context.

## Managing rules

```bash
plannotator tripwires list        # merged rules grouped by origin + live tripped/armed status
plannotator tripwires add --glob "src/billing/**" --note "money path"   # direct write (global)
plannotator tripwires add --glob "migrations/**" --repo                 # team file (explicit)
plannotator tripwires add protect the auth core    # free text → prints agent instructions
plannotator tripwires validate    # diagnostics fail-open swallows at review time
plannotator tripwires path        # where each layer's file lives
```

In-agent: `/plannotator-review --tripwires` (`-t`) prints a live scan instead of opening the UI; `--add-tripwire <description...>` prints rule-authoring instructions for the agent. Both short-circuit before any server/browser startup, in all three runtimes (Claude Code/OpenCode/Pi); PR scans short-circuit before paying for a checkout.

## How it works

- `packages/shared/tripwires.ts` — pure evaluation engine (vendored to Pi). Glob matching on a linear backtracking-free NFA (ReDoS-immune; relevant since fork-PR mode can evaluate hostile configs) with a 256-char glob cap. Diff parser handles both sides, renames, `/dev/null`, C-quoted paths, binary stanzas, mode-only changes.
- Hits inject as external annotations (`source: "tripwire"`), cleared + re-evaluated on every patch change in both runtimes. Fail-open everywhere — both layers fail independently; a malformed config never breaks a review.
- Docs guide + launch blog post included. This repo commits a repo-layer config as the team-opt-in example (and dogfood).

## Verification

- 1,354 tests / 0 fail (102 tripwire-specific incl. ReDoS timing guards, rename/deletion/binary/chmod, key-derivation fallbacks, append-rule round-trips)
- Multi-agent adversarial review across both phases: 20 findings raised, 13 confirmed & fixed (critical ReDoS, rename-keying, PR-checkout short-circuit ordering, git<2.31 key collision, …), rest refuted with evidence
- E2E smokes: auto-instantiation, add/add `--repo`/list/validate, both review flags, no-browser short-circuit


backnotprop:
Tripwires never landed and nothing equivalent is on main (the only `tripwire` matches are incidental words in two test files). The branch is 41 files and +3,929/-36, has never been reviewed, has no issue behind it, and is 555 commits behind — including a full replacement of the review sidebar, the annotation scope model, the decision control, and the external-annotation validators it would have to integrate with. Rebasing it is not a merge, it is a re-implementation against a different codebase.

Recommendation: close the PR and open an issue carrying the design — the two-layer global-first rule file keyed on normalized remote URL, "any touch trips", the passive/no-gating posture, and the deliberate exclusion from feedback and annotation counts. Those decisions are the valuable part and they are currently only recorded in a PR description that will fall off the radar. Nothing here is wrong; it is just a large unreviewed feature branch that the tree has moved past.

PR #868, opened 2026-06-08
refactor(agent-mode): Unify agent mode engine selection

codythatsme:
## Summary

Hey! Feel free to reject this PR if this is not the direction you want to take. The inconsistency between Code Review and Code Tour in the Review Agents panel was bothering me: Code Tour was one mode with an engine picker inside it, while Code Review was split into separate `Code Review · Claude` and `Code Review · Codex` entries.

This unifies that interaction. The top-level choice is now just the thing you want to run - Code Review or Code Tour - and both modes use the same Engine dropdown to pick Claude or Codex.

## What Changed

- Replaced the split Code Review provider choices with a single `Code Review` mode.
- Added a shared `Engine` select for Code Review and Code Tour.
- Replaced the Code Tour radio buttons with the same engine select used by Code Review.
- Kept backend launch behavior intact: Code Review still maps Claude/Codex to the existing providers, and Code Tour still launches through the tour provider with an engine.
- Updated saved agent settings to persist `selectedMode`, `reviewEngine`, and `tourEngine`.

### Code Review

| Before | After |
| --- | --- |
| Before: Code Review exposed as Code Review Codex mode | After: Code Review with Codex selected in the Engine dropdown |

### Code Tour

| Before | After |
| --- | --- |
| Before: Code Tour used radio buttons for engine selection | After: Code Tour uses the same Engine dropdown |

## Validation

- `bun test packages/ui/hooks/useAgentSettings.test.ts`
- `git diff --check`
- Browser smoke test against the review UI confirmed both modes use selects and no radio inputs remain in the Agents panel.


backnotprop:
Ah, that is better. I like it.

PR #878, opened 2026-06-09
feat(ui): render markdown math

### LINKED ISSUE #831: FR: Support Markdown math formula rendering in Plannotator
## Feature request

It would be helpful if Plannotator supported rendering Markdown math formulas, including inline math and block/display math.

Currently, formulas appear to be rendered as plain text instead of being interpreted as math.

For example, this Markdown:

$$
\int_0^1 x^2 dx = \frac{1}{3}
$$

is currently rendered as something like:

```text
[ \int_0^1 x^2 dx = \frac{1}{3} ]
```

This feature would make technical Markdown documents easier to read and annotate. Thanks!

ishowman:
## Summary

Close #831, and extend Plannotator math rendering to support the common KaTeX/LLM-friendly delimiters `\(...\)` and `\[...\]`, in addition to the existing `$...$` and `$$...$$` forms.

## What changed

- Render inline math with both `$...$` and `\(...\)`.
- Render display math with both `$$...$$` and `\[...\]`.
- Keep escaped dollars and non-math text literal.
- Preserve accurate line tracking for display-math blocks so annotation exports stay correct.

## Scope

This stays within the UI/parser layer and does not require server API changes.

## Validation

- `bun test packages/ui/components/InlineMarkdown.test.ts packages/ui/utils/parser.test.ts`
- `PATH=/Users/xxq/Documents/codes/plannotator/packages/ui/node_modules/.bin:$PATH bun run typecheck`
- `git diff --check`

## Notes

This update is aimed at markdown generated by LLMs and math rendered by KaTeX-style delimiters.


XxxXMil:
Would it be possible to also support `\(...\)` and `\[...\]` delimiters? They’re common in LLM-generated Markdown and KaTeX auto-render supports them by default.

ishowman:
> Would it be possible to also support `\(...\)` and `\[...\]` delimiters? They’re common in LLM-generated Markdown and KaTeX auto-render supports them by default.

I've submitted new commit for it.Could you please verify if the result matches your expectations when you have spare time?

XxxXMil:
Thanks for working on this! I tested the PR locally in Pi by installing the PR branch.

Some parts work well:
- Inline math with `\(...\)` renders correctly.
- Block math with `\[...\]` and `$$...$$` renders correctly.

I also found a few issues during manual testing:

1. Inline math using `$...$` is not very stable in my testing. Some `$...$` expressions render, but others remain as raw text.
2. Highlighting anchoring doesn't shown on rendered math.

I attached screenshots showing these cases:

image

image

Raw text for the last screenshot:
```text
In a linear Gaussian model, the prediction \( \hat{y} = \mathbf{x}^\top \boldsymbol{\beta} \) is compared against the target $y$ using the loss \( \mathcal{L}(\boldsymbol{\beta}) = \frac{1}{2N}\sum_{i=1}^{N}(y_i - \mathbf{x}_i^\top\boldsymbol{\beta})^2 \), while the noise term satisfies $\varepsilon \sim \mathcal{N}(0,\sigma^2)$.
```
In this case, `$...$` works well.

XxxXMil:
Small follow-up: I found the cause of the inconsistent `$...$` behavior.

```text
Testing inline `$...$` with spaces inside the delimiters:

$ \varepsilon \sim \mathcal{N}(0,\sigma^2) $

Testing inline `$...$` without spaces inside the delimiters:

$\varepsilon \sim \mathcal{N}(0,\sigma^2)$
```

image

`$\varepsilon \sim \mathcal{N}(0,\sigma^2)$` renders correctly, but `$ \varepsilon \sim \mathcal{N}(0,\sigma^2) $` stays as raw text because of the spaces just inside the delimiters.

This might be avoidable by prompting LLMs to emit `$...$` without inner spaces?

Other test:
image

backnotprop:
Are these worth fixing before shipping this?

backnotprop:
@XxxXMil do you see these things as blocking?

I can jump in

XxxXMil:
My concern is that highlighting isn't working for formulas.

ishowman:
@XxxXMil  KaTeX with space and without space should render as expected now, highlight is still in progress  @backnotprop .

backnotprop:
Thanks @ishowman and @XxxXMil 

shipped this:
https://x.com/plannotator/status/2072390291177554332?s=20

PR #884, opened 2026-06-10
fix(annotate): add /api/save-notes POST endpoint to annotate server

### LINKED ISSUE #844: Bug: Failed to Save to Obsidian
I have Plannotator v0.19.26 and I can't save/export to Obsidian. When I try to save to Obsidian I get the error: `Save failed`.

Image

LMK what other diagnostics I can provide.

titosemi:
Fixes #844

## Problem

Saving annotations to Obsidian (`Save to Obsidian` button) fails silently in annotation mode. The button is visible and clickable but nothing happens because the annotate server lacks the `/api/save-notes` POST route.

The plan review server has this route and works correctly — only the annotate server was missing it.

## Root Cause

`packages/server/annotate.ts` (Bun source) and `apps/pi-extension/server/serverAnnotate.ts` (Pi extension copy) did not implement the `/api/save-notes` handler. When the frontend sends the save request in annotation mode, the server falls through to the HTML catch-all route, returning the SPA HTML instead of JSON.

## Changes

- **`packages/server/annotate.ts`** — Added `/api/save-notes` POST route handler with the same integration logic as `serverPlan.ts`: accepts `obsidian`, `bear`, and `octarine` configs, runs configured saves in parallel, returns `{ ok: true, results }`.
- **`apps/pi-extension/server/serverAnnotate.ts`** — Same route added to the Pi extension copy of the annotate server.
- **`packages/server/integrations.test.ts`** — Added `saveToObsidian` unit tests (success and missing vault cases). Consolidated duplicate imports from `./integrations` into a single statement.
- **`packages/server/annotate.test.ts`** — HTTP endpoint tests for the new route (success, empty integrations, integration-level error).

## Testing

```
bun test packages/server/annotate.test.ts
  ✓ POST saves to Obsidian vault and returns success
  ✓ POST returns 200 with empty results when no integrations
  ✓ POST with missing vault returns integration error, not server error

bun test packages/server/integrations.test.ts
  ✓ saveToObsidian writes plan file to temp vault
  ✓ saveToObsidian fails when vault path does not exist

bun test tests/parity/route-parity.test.ts
  ✓ All Bun ↔ Pi routes match (including save-notes)
```

All 205 server tests pass, 0 fail.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -258,6 +258,7 @@ During normal plan review, an Archive sidebar tab provides the same browsing via
 | `/api/done`           | POST   | Close archive browser (archive mode only)  |
 | `/api/approve`        | POST   | Approve plan (body: planSave, agentSwitch, obsidian, bear, feedback) |
 | `/api/deny`           | POST   | Deny plan (body: feedback, planSave)       |
+| `/api/save-notes`     | POST   | Save to external note apps (Obsidian, Bear, Octarine) |
 | `/api/image`          | GET    | Serve image by path query param            |
 | `/api/upload`         | POST   | Upload image, returns `{ path, originalName }` |
 | `/api/obsidian/vaults`| GET    | Detect available Obsidian vaults           |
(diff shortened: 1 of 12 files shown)


backnotprop:
thank you!

backnotprop:
Thanks!! I cleaned up conflicts, extracted the save-notes handler into the shared handler modules, and fixed the catch-block bug. Ready for merge.

PR #912, opened 2026-06-13
feat(review): add Cursor CLI as an experimental review agent provider

backnotprop:
## What

Adds **Cursor CLI as an experimental third background review provider** (`cursor`), alongside Claude, Codex, and Code Tour. It plugs into the existing `/api/agents/jobs` job system — **not** Ask AI — so it reuses the same job cards, live logs, findings, and external-annotation pipeline as the other review agents.

## Design principle

**Match the existing providers; don't out-engineer them.** Cursor's only real difference is that its output is prose, not schema-validated JSON (it has no `--json-schema` flag). That earns strict *parsing* — and nothing else. Deliberately **not** included: per-job nonce, diff-grounding gate, isolated `CURSOR_CONFIG_DIR`, env scrubbing. Cursor inherits the environment exactly like Claude/Codex.

## Backend

- **`packages/server/cursor-review.ts`** (new) — command builder (read-only flags, prompt on stdin), a **line-buffered NDJSON stream reducer** (pipe chunks ≠ record boundaries), static-marker **last-block** JSON extraction, schema validation, and a finding→annotation transform mirroring `transformClaudeFindings` (author `Cursor`).
- **`agent-jobs.ts`** — capability detection (the Cursor binary is literally `agent`, not `cursor`); a **fail-closed server-built-provider guard** so client-supplied argv is never spawned for built-in providers; cursor stdout-log dispatch.
- **`review.ts` `onJobComplete`** — **fails closed by mutating the job** (`job.status = "failed"`), never by throwing (the runner swallows throws and would otherwise leave an exit-0 job green); PR/diff-scope attribution; workspace path normalization.
- **Full Pi parity** — `apps/pi-extension/server/{agent-jobs,serverReview}.ts`, vendored via `vendor.sh`.

## UI

Capability-gated and **invisible unless the Cursor CLI is installed**. In Code Review mode, Cursor appears as a review-only engine with an **experimental** label and an `Auto` model default. Claude / Codex / Code Tour are untouched. *(This part is intentionally minimal — flagged for refinement during review.)*

## Safety

Read-only posture is **flags-only**: `--mode ask` + `--sandbox enabled`, no `--force`/`--yolo`.

## Verification

- `bun run typecheck` (vendor.sh + tsc across shared/ai/server/ui/pi-extension): **green**.
- `cursor-review.test.ts`: **22/22** (high-value only — chunk-split NDJSON, partial-output dedup, last-block-wins, every failure path → fail, transform shape).
- Affected existing review-agent test families: pass.

This was built and then put through an adversarial multi-agent review, which caught and fixed a real blocker (a Bun-only call in dead model-discovery code that broke the Pi Node typecheck — removed).

## ⚠️ Outstanding (must verify before trusting in production)

The **read-only safety gate is human-only**: confirming `agent -p --mode ask` is actually non-mutating requires an **authenticated** Cursor account (the dev shell was unauthenticated). The code ships the right flags and is labeled experimental, but that behavioral guarantee is **not yet verified**.

## Notes for reviewers

- Design rationale docs (`cursor-cli-review-agent-approach.md` + handoffs) were kept **out** of this commit to keep the diff focused on code. They can be added if useful.
- UI is type-clean but **not runtime-exercised** (built/reviewed headlessly).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -311,7 +311,7 @@ During normal plan review, an Archive sidebar tab provides the same browsing via
 | `/api/external-annotations` | POST | Add external annotations (single or batch `{ annotations: [...] }`) |
 | `/api/external-annotations` | PATCH | Update fields on a single annotation (`?id=`) |
 | `/api/external-annotations` | DELETE | Remove by `?id=`, `?source=`, or clear all |
-| `/api/agents/capabilities` | GET | Check available agent providers (claude, codex, tour) |
+| `/api/agents/capabilities` | GET | Check available agent providers (claude, codex, tour, cursor) |
 | `/api/agents/jobs/stream` | GET | SSE stream for real-time agent job status updates |
 | `/api/agents/jobs` | GET | Snapshot of agent jobs (polling fallback, `?since=N` for version gating) |
 | `/api/agents/jobs` | POST | Launch an agent job (body: `{ provider, command, label }`) |
--- a/apps/pi-extension/server/agent-jobs.ts
+++ b/apps/pi-extension/server/agent-jobs.ts
@@ -95,28 +95,6 @@ export interface AgentJobHandlerOptions {
 }
 
 /**
- * `agent` is a very generic binary name — other tools ship binaries called
- * `agent` too, so finding one on PATH does not prove it is Cursor. Verify its
- * identity before offering Cursor as a provider. Cheap one-time probe: `agent
- * about --format json` is fast, works unauthenticated, and only the Cursor CLI
- * emits a JSON object with a `cliVersion` field. Any failure → not Cursor.
- */
-function isCursorAgentAvailable(): boolean {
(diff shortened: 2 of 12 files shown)


backnotprop:
Superseded by [#959](https://github.com/backnotprop/plannotator/pull/959) ("Cursor + OpenCode review engines — unified marker-review"), merged 2026-06-24 and shipped in **v0.21.2**. `cursor` is a first-class agent-job provider on main today (`packages/server/agent-jobs.ts:75`, with the `agent`-binary detection note at `:210`), and it later grew `PLANNOTATOR_CURSOR_SANDBOX` ([#1095](https://github.com/backnotprop/plannotator/pull/1095)) for hosts where Cursor's sandbox cannot start. The unified marker-review path replaced this branch's standalone `packages/server/cursor-review.ts`, which does not exist on main. Closing; the branch is 503 commits behind and nothing in it is still needed.

PR #921, opened 2026-06-16
feat(config): support share toggle via config.json

yonihorn:
## Summary

Adds support for toggling URL sharing via `~/.plannotator/config.json`, not just the `PLANNOTATOR_SHARE` environment variable. Teams who want sharing off persistently can now set it in config instead of exporting an env var in every shell.

```json
{ "share": "disabled" }
```

## Details

- New `share?: "enabled" | "disabled"` field on `PlannotatorConfig`, mirroring the env var's value so the two stay aligned.
- New `resolveSharingEnabled(config)` helper following the existing `resolveUseJina`/`resolveUseGlimpse` pattern. Precedence: `PLANNOTATOR_SHARE` env var → `config.share` → default enabled. The env var keeps its existing `!== "disabled"` semantics and remains highest priority, so current behavior is unchanged.
- All inline `process.env.PLANNOTATOR_SHARE !== "disabled"` checks now route through the helper: hook server, OpenCode plugin (config-aware fallback), and the Pi extension (browser + server fallbacks).
- Docs updated (configuration, environment-variables reference, sharing guide, AGENTS.md).

`typecheck` and `bun test` both pass.


yonihorn:
Hi @backnotprop ! will appreciate your review on this PR 
My team would benefit from having this config value rather than using the env variable 🙏

backnotprop:
Thanks!!

backnotprop:
Only fixed the merge conflicts, thanks for this, release will be tonight.

PR #939, opened 2026-06-18
fix(ai): resolve Bedrock/Vertex model ids for Ask AI

### LINKED ISSUE #938: Ask AI fails on Bedrock/Vertex: hardcoded bare model alias rejected with 400
## Summary

The Ask AI feature can't be used with Amazon Bedrock (or Vertex AI). Every query returns:

> API Error (claude-sonnet-4-6): 400 The provided model identifier is invalid.

## Environment

- Agent: Claude Code on Amazon Bedrock (`CLAUDE_CODE_USE_BEDROCK=1`)
- Model configured via env as a full inference-profile ARN (`ANTHROPIC_MODEL`, `ANTHROPIC_DEFAULT_*_MODEL`), as Bedrock requires.

## Root cause

`packages/ai/providers/claude-agent-sdk.ts` hardcodes `DEFAULT_MODEL = "claude-sonnet-4-6"` and passes it (or the picker's bare alias) to the spawned `claude` binary as `--model`. Bare aliases work on the first-party Anthropic API, but Bedrock/Vertex require a full inference-profile id / ARN (e.g. `arn:aws:bedrock:...:inference-profile/global.anthropic.claude-sonnet-4-6`), so the request 400s.

A normal Claude Code session works because it reads the ARN from the env (`ANTHROPIC_MODEL` / `ANTHROPIC_DEFAULT_*_MODEL`); Ask AI overrides that with the bare alias, so Bedrock rejects it. The model picker is affected too — it only offers bare aliases, so every option 400s on Bedrock.

## Proposed fix

When running on Bedrock/Vertex, map the picker's family aliases (opus / sonnet / haiku) to the matching `ANTHROPIC_DEFAULT_*_MODEL` ARN, pass through identifiers that are already ARNs / inference profiles, and otherwise inherit the env default. Off-cloud behavior unchanged.

masterpavan:
Closes #938

## Problem

The Ask AI feature is unusable on Amazon Bedrock (and, by the same mechanism, Vertex AI). Every query fails with:

> API Error (claude-sonnet-4-6): 400 The provided model identifier is invalid.

### Root cause

`ClaudeAgentSDKProvider` hardcodes a bare model alias and force-passes it to the spawned `claude` binary as `--model`:

- `packages/ai/providers/claude-agent-sdk.ts` — `DEFAULT_MODEL = "claude-sonnet-4-6"`, used whenever no model is explicitly selected (`baseConfig()`), and the model picker only offers bare aliases too.

Bare aliases like `claude-sonnet-4-6` resolve fine against the first-party Anthropic API, but Bedrock/Vertex require a full inference-profile id / ARN (e.g. `arn:aws:bedrock:...:inference-profile/global.anthropic.claude-sonnet-4-6`). A normal Claude Code session works because it reads that ARN from `ANTHROPIC_MODEL` / `ANTHROPIC_DEFAULT_*_MODEL`; Ask AI overrides it with the bare alias, so Bedrock 400s.

## Fix

Add `resolveSDKModel()` in the Claude provider. When `CLAUDE_CODE_USE_BEDROCK` or `CLAUDE_CODE_USE_VERTEX` is set:

- pass through identifiers that are already ARNs / `anthropic.` inference profiles
- map the picker's family aliases (opus / sonnet / haiku) to the matching `ANTHROPIC_DEFAULT_*_MODEL` env var
- otherwise fall back to `ANTHROPIC_MODEL`, or omit `--model` so the SDK inherits the environment's default

Off Bedrock/Vertex, behavior is unchanged (alias returned as-is).

## Verification

- `bun test packages/ai/ai.test.ts` → 97 pass (9 new `resolveSDKModel` cases: alias→ARN mapping, ARN/profile pass-through, `[1m]` variant, env fallback, Vertex).
- Drove the real provider via the production factory on a live Bedrock account with **no model set** (the failing path): it resolved to the configured ARN and returned a successful reply — no 400.

Happy to adjust the approach (env-var name, mapping strategy, etc.) if you'd prefer something different.

### DIFF
--- a/packages/ai/ai.test.ts
+++ b/packages/ai/ai.test.ts
@@ -802,6 +802,79 @@ describe("AI endpoints", () => {
 });
 
 // ---------------------------------------------------------------------------
+// Claude Agent SDK — Bedrock/Vertex model resolution
+// ---------------------------------------------------------------------------
+
+import { resolveSDKModel } from "./providers/claude-agent-sdk.ts";
+
+describe("resolveSDKModel", () => {
+  const SONNET_ARN =
+    "arn:aws:bedrock:us-east-1:123456789012:inference-profile/global.anthropic.claude-sonnet-4-6";
+  const OPUS_ARN =
+    "arn:aws:bedrock:us-east-1:123456789012:inference-profile/global.anthropic.claude-opus-4-8[1m]";
+  const HAIKU_ARN =
+    "arn:aws:bedrock:us-east-1:123456789012:inference-profile/global.anthropic.claude-haiku-4-5-20251001-v1:0";
+
+  const bedrockEnv = {
+    CLAUDE_CODE_USE_BEDROCK: "1",
+    ANTHROPIC_MODEL: OPUS_ARN,
+    ANTHROPIC_DEFAULT_SONNET_MODEL: SONNET_ARN,
+    ANTHROPIC_DEFAULT_OPUS_MODEL: OPUS_ARN,
+    ANTHROPIC_DEFAULT_HAIKU_MODEL: HAIKU_ARN,
+  };
+
+  test("off Bedrock/Vertex: returns the requested alias unchanged", () => {
+    expect(resolveSDKModel("claude-sonnet-4-6", {})).toBe("claude-sonnet-4-6");
+  });
+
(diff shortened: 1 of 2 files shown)


backnotprop:
Thank you!

PR #384, opened 2026-03-24
feat(gemini): Gemini CLI plan review integration

backnotprop:
## Summary

- Adds `apps/gemini-hook/` — a thin adapter that enables Plannotator plan review for Gemini CLI via the `BeforeTool` hook system
- Reads plan file from disk (Gemini provides `plan_path`, not inline content), delegates to shared `@plannotator/server`, translates decision to Gemini's hook output format
- Install script updated to download Gemini binary, install policy TOML to `~/.gemini/policies/`, and configure `~/.gemini/settings.json`

## How it works

1. User policy (`plannotator.toml`) grants `allow` for `exit_plan_mode` → skips TUI dialog
2. `BeforeTool` hook fires → `plannotator-gemini` binary starts
3. Adapter reads plan from disk, opens Plannotator review UI in browser
4. User approves/denies with annotations → decision returned to Gemini as hook output
5. On deny: model receives structured feedback via `planDenyFeedback()`, revises plan

Depends on upstream fix: google-gemini/gemini-cli#21802 (merged)

## Test plan

- [ ] `echo '<BeforeTool JSON>' | bun run apps/gemini-hook/server/index.ts` opens browser with plan
- [ ] Approve → stdout is `{}`
- [ ] Deny with feedback → stdout is `{"decision":"deny","reason":"YOUR PLAN WAS NOT APPROVED..."}`
- [ ] End-to-end with patched Gemini CLI: `/plan` → task → plan appears in browser → approve/deny works

PR #946, opened 2026-06-19
feat(antigravity): add native plan review and installer support

ykalani:
Adds native Antigravity CLI plan-file review and installer support.

When Antigravity is detected, the installer writes `~/.gemini/config/plugins/plannotator/{plugin.json,hooks.json}` and installs review/annotation skills. `~/.gemini/antigravity-cli` is only a detection signal, never an installation destination.

The named `PreToolUse` hook intercepts writes to the current conversation's `implementation_plan.md`. The runtime reads Antigravity's camelCase `toolCall` payload, opens Plannotator with the `antigravity` origin, and returns native allow/deny decisions with feedback. Incremental plan edits request a full-file resubmission; unrelated writes abstain with `{}`.

Installation safeguards:
- CLI, environment, and config opt-outs; detected/skipped/not-detected reporting; uninstall cleanup, including paths left by the original integration.
- Probe the downloaded binary before writing Antigravity configuration, so old Plannotator releases are skipped. Hooks reference that exact binary by absolute path.
- Antigravity-specific skills set `PLANNOTATOR_ORIGIN=antigravity` for review and annotation.
- Antigravity-only profiles do not receive Gemini configuration merely because they share `~/.gemini`. Profiles with Gemini-specific signals retain both integrations.

Scope and compatibility:
- This gates conventional plan-file writes, not a planning-to-implementation transition. Shell writes, custom plan filenames, and Antigravity's own artifact-review settings remain outside this hook.
- Antigravity CLI 1.2.7+ is the documented support baseline. A live session acceptance check and maintainer agreement on the plan-write gate remain outstanding; this PR does not claim end-to-end verification.
- This is a minor-release feature; no release version is bumped here.

Validation covers temporary-home installs in Bash, PowerShell, and CMD; old-binary preservation; Gemini coexistence; repeat installs; quoted binary paths; opt-outs; uninstall; and native payload/decision handling. Installed plugin shape is also checked with `agy plugin validate` when available. See `apps/antigravity/README.md` for the live acceptance checklist.

### DIFF
--- a/scripts/install.cmd
+++ b/scripts/install.cmd
@@ -714,6 +714,16 @@ if !ERRORLEVEL! equ 0 (
         echo Installed Gemini commands to !GEMINI_COMMANDS_DIR!\
     )
 
+    REM Antigravity CLI plugin commands -- ponytail: resolve AGY_BASE once
+    set "AGY_BASE="
+    if exist "%USERPROFILE%\.gemini\config" (set "AGY_BASE=%USERPROFILE%\.gemini\config") else if exist "%USERPROFILE%\.gemini\antigravity-cli" set "AGY_BASE=%USERPROFILE%\.gemini\antigravity-cli"
+    if defined AGY_BASE if exist "apps\gemini\commands" (
+        set "AGY_PLUGIN_COMMANDS_DIR=!AGY_BASE!\plugins\plannotator\commands"
+        if not exist "!AGY_PLUGIN_COMMANDS_DIR!" mkdir "!AGY_PLUGIN_COMMANDS_DIR!"
+        xcopy /y /q "apps\gemini\commands\*.toml" "!AGY_PLUGIN_COMMANDS_DIR!\" >nul 2>&1
+        echo Installed Antigravity commands to !AGY_PLUGIN_COMMANDS_DIR!\
+    )
+
     REM Kiro -> hand-maintained kiro skills (3) + 2 extras, only when detected.
     if "!KIRO_AVAILABLE!"=="1" if exist "apps\kiro-cli\skills" (
         if not exist "!KIRO_SKILLS_DIR!" mkdir "!KIRO_SKILLS_DIR!"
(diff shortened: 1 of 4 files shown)


backnotprop:
Thanks!

backnotprop:
Thanks @ykalani, and sorry this stalled after a one-word reply. Checked it against main: it is **not** superseded. The shipped Gemini leg writes `~/.gemini/policies/plannotator.toml` plus a `BeforeTool` hook into `~/.gemini/settings.json` (`scripts/install.sh:1913-1974`, from [#384](https://github.com/backnotprop/plannotator/pull/384), v0.16.7), while your branch targets Antigravity's distinct *plugin* layout — `~/.gemini/{config|antigravity-cli}/plugins/plannotator/{plugin.json,hooks.json}` and its own `policies/` dir. That is genuinely additive.

The branch is 507 commits behind and `scripts/install.sh` has been restructured substantially since June, so it needs a rebase. Three conventions landed after you opened it that a new host leg is now expected to follow:

1. **An opt-out:** `--skip-antigravity` + `PLANNOTATOR_SKIP_ANTIGRAVITY_INSTALL` + `skipInstall.antigravity`, matching the codex/gemini/kiro/opencode pattern (`scripts/install.sh:296-311`, `packages/shared/config.ts:163-168`).
2. **Honest reporting:** the installer now prints `detected / skipped (reason) / not detected` per host rather than an unconditional banner — the unconditional `ANTIGRAVITY CLI USERS` block in the current diff would print even when nothing was detected.
3. **Uninstall:** host integrations are removed by `packages/server/uninstall.ts`; a new leg should add its paths there (see #1179 for the remaining gaps).

Are you up for a rebase? If we don't hear back in 14 days I'll close this to keep the queue honest — it can be reopened any time, and the design above is the whole review.

ykalani (commented):
Rebased onto current main and pushed the update to this PR.

Added all three requested conventions:
- Opt-outs via `--skip-antigravity` (`-SkipAntigravity` in PowerShell), `PLANNOTATOR_SKIP_ANTIGRAVITY_INSTALL`, and `skipInstall.antigravity`.
- Detection reporting that distinguishes installed, skipped with reason, and not detected.
- Uninstall cleanup for both Antigravity layouts, preserving unrelated host files.

Installer and uninstall tests: 197 passed, 9 platform skips. Coverage includes Bash, PowerShell, and CMD, opt-out precedence, repeat installs, and preservation of existing files when skipped.

Ready for another review. Thanks for the guidance!

backnotprop:
Thanks for the rebase, the opt-out, reporting, and uninstall work is exactly right and should stay. But checked against the Antigravity CLI itself, the plugin it writes can't fire:

- Antigravity's `hooks.json` has no `"hooks"` wrapper and uses `PreToolUse`, not `BeforeTool` (the binary has zero `BeforeTool` strings).
- There is no `exit_plan_mode` tool in Antigravity, so the matcher matches nothing. Which event should Plannotator gate on is the open question.
- `policies/` is Gemini CLI's mechanism; Antigravity has none, so `policies/plannotator.toml` is a dead file, and `plugins/<name>/commands/` isn't part of its plugin layout either.
- The only global config root it scans is `~/.gemini/config/`; `~/.gemini/antigravity-cli/` is its private data dir (its own changelog calls writing there a bug).
- There's no runtime side: `apps/hook/server/index.ts` only understands Claude Code and Gemini payloads, so even a firing hook would exit with "No plan content".

So this needs the runtime adapter (payload branch + an `antigravity` origin) and the corrected plugin shape before it can do anything. Happy to keep it open if you want to take that on; it's a minor-release feature either way.

PR #961, opened 2026-06-24
feat(annotate): per-file version diff for .md and .html (rendered HTML highlights)

### LINKED ISSUE #960: Annotate mode: show the version diff (highlighted) for .md and .html files
## Summary

Plan mode shows a highlighted **version diff** (the `+N/-M` badge → `PlanDiffViewer`) when a plan is revised, powered by `previousPlan`/`versionInfo` that the plan server saves to `~/.plannotator/history/...`. **Annotate mode has no equivalent** — `plannotator annotate <file>` never tracks history, so there's no way to see *what changed* between edits of a standalone Markdown note or an HTML spec, even though the diff UI and block-level commenting already exist in the editor.

This is especially useful for an iterative loop: annotate a doc → the agent edits it → reopen and immediately see the highlighted changes before the next round.

## Request

Bring the version diff to **annotate mode** for both:

- **`.md` / `.txt`** — the existing rendered Markdown diff (clean/classic/raw).
- **`.html` with `--render-html`** — render the diff as the **real page with inline `<ins>`/`<del>` highlights**, so you see *where* it changed (not a source dump), with commenting still working on the diff view.

## Notes / constraints

- History should be keyed by **file path** (stable across edits) rather than the plan flow's first-heading+date slug — re-opening the same file should group its versions even when headings change.
- The diff UI (`PlanDiffViewer`, Version Browser, block annotations) is already mode-agnostic and consumes `previousPlan`/`versionInfo`; the annotate server just needs to send them.
- Per `CLAUDE.md`, the change must land in **both** the Bun (`packages/server/annotate.ts`) and Pi (`apps/pi-extension/server/serverAnnotate.ts`) runtimes.

Happy to contribute this — PR incoming.

egouilliard-leyton:
Closes #960

## What

Brings the **version diff** (the highlighted "what changed since the last version" view that plan mode has) to **annotate mode**, for both Markdown and HTML files. Annotate already has the diff UI (`PlanDiffViewer`, Version Browser, block-level comments) — it just never received `previousPlan`/`versionInfo`, because the annotate server didn't track history. This wires that up.

- **`.md` / `.txt`** → the existing rendered Markdown diff (clean / classic / raw), with block comments.
- **`.html` with `--render-html`** → the diff is rendered as the **real page with inline `<ins>`/`<del>` highlights** (not a source dump), so you see *where* it changed, and commenting still works because it's the same `HtmlViewer` iframe.

## How

- **History keyed by file path.** Unlike the plan flow (`generateSlug` = first-heading + date), annotate keys history by file path — `annotate-{sanitized-basename}-{hash8}` — so re-opening the same file groups its versions across edits even when headings change. Dedup + sequential numbering reuse the existing `saveToHistory` storage.
- **Server (`packages/server/annotate.ts` + Pi `apps/pi-extension/server/serverAnnotate.ts`).** On open, save the markdown (or raw HTML source) to history, then expose `previousPlan` / `versionInfo` / `diffCurrent` on `/api/plan` and add `/api/plan/version` + `/api/plan/versions`. Both runtimes updated per `AGENTS.md`.
- **HTML rendered diff.** New `packages/shared/html-diff.ts`: a small, dependency-light, tag-aware `htmlDiff(old, new)` that wraps changed text in `<ins>`/`<del>` while keeping tags balanced (`<script>`/`<style>` treated as opaque). The server sends `diffHtml = rewriteHtml(htmlDiff(prev, current))`; `HtmlViewer` injects `ins`/`del` highlight CSS and shows a **"Show changes"** toggle; `App` swaps the iframe to the diff page when toggled and suppresses the Markdown block-diff path on the HTML surface (where `markdown` is empty).

## Testing

- `packages/shared/html-diff.test.ts` — 9 unit tests (text change, word replace, added element, script/style opacity, identical-input no-op, structural balance).
- `tsc --noEmit` clean for `packages/shared`, `packages/server`, `packages/ui`; frontend build clean.
- Manually verified end to end on a real 58 KB HTML spec: the diff renders as the page with inline highlights, the toggle works, and commenting works on the diff view. `/api/plan` returns `renderAs: html` + `diffHtml`, and `/api/plan/versions` lists the stored versions.

## Notes

- HTML diff is a linear token diff (delete + insert for moves; attribute-only changes re-emit the tag without a visible marker). Good enough for the prose/spec-review use case; a tree-aware pass could be a follow-up.


backnotprop:
Great feature and genuinely well-built — the reuse of the plan-version machinery, the dual-runtime mirror, and the htmlDiff test cases all show real care. Thank you!

We reviewed in depth and, to save you a round-trip, pushed the fixes directly to your branch (hope that's OK — happy to walk through any of it):

1. **`diff` dependency declared in `apps/pi-extension/package.json`** — `html-diff` imports it at module load, so a standalone Pi install couldn't resolve it (the monorepo masked this via hoisting) and every annotate session broke.
2. **Tokenizer made attribute-aware** — `<[^>]*>` split tags at a `>` inside quoted attributes (e.g. `title="a > b"`), corrupting the diff output. Now quoted values are consumed whole; 3 regression tests added.
3. **History persistence disclosed + opt-out** — saving a copy of every annotated file into the data dir is new behavior users should be able to see and disable: `PLANNOTATOR_ANNOTATE_HISTORY=0` / `{ "annotateHistory": false }`, default on, documented in AGENTS.md.
4. **Versions tab hidden on the HTML surface** — the base-version picker had nothing to drive there (the HTML diff is fixed to current-vs-previous), so it was a dead control; the viewer's "Show changes" toggle is untouched.

All gates green after the changes (typecheck, 1,698 tests, both app builds). Non-blocking notes for possible follow-ups, no action needed: a size cap on `htmlDiff` for very large documents, and `<textarea>` contents rendering literal `<ins>` markers when changed.

Merging once CI confirms — thanks again for the contribution!

backnotprop:

Issue #949, opened 2026-06-21
Regression for `/plannotator-last` command on pi


emmaneugene:
As of the latest v0.21.0 release, I believe there's a regression with the plannotator pi extension, where the `/plannotator-last` command renders a blank page when the session has more than one recent assistant message

I did some debugging with agents and I believe this was caused by commit 1fa60522, which converted `messageAnnotationEntries` from a render-time `useMemo` into an on-demand `useCallback`. I'm not super familiar with React, but it seems that root cause is that for conversations with more than 1 assistant response, `setState()` is being called during render and causing an infinite render loop.

I created a fork with a fix and more detailed bug report [here](https://github.com/emmaneugene/plannotator/tree/fix/pi-plannotator-last-render-loop), but it's AI-generated, so it probably deserves a review. 

Thanks for making an awesome tool!

backnotprop:
Fixed in [v0.21.1](https://github.com/backnotprop/plannotator/releases/tag/v0.21.1) via #950. Your diagnosis was spot on — the render-path call to `saveCurrentMessageState()` was writing state during render; it now reads via `getMessageStatesWithCurrent()`. Thanks for the thorough report and the fix, @emmaneugene.

Update with `pi install npm:@plannotator/pi-extension` to pick it up.

PR #962, opened 2026-06-24
fix(opencode): honor cli runtime for command routing

### LINKED ISSUE #947: plannotator last renders blank page in manual workflow mode (OpenCode)
## Description

When the OpenCode plugin is configured with `workflow: manual`, running the `/plannotator-last` slash command opens the browser but renders a blank page.

## Environment

- **Tool:** OpenCode (not Claude Code)
- **Plugin version:** `@plannotator/opencode@latest` (0.21.0)
- **Plannotator CLI version:** 0.21.0 (darwin-arm64)
- **Workflow mode:** `manual`

## Steps to Reproduce

1. Configure the plugin with `workflow: manual` in `opencode.jsonc`
2. Have an active OpenCode session with conversation history
3. Run the `/plannotator-last` slash command (which triggers the skill that calls `plannotator last`)
4. Browser opens at `http://localhost:<port>` → blank dark page, no content rendered

## What Works

Running `plannotator annotate-last --stdin` directly from bash with piped content works correctly and renders the UI.

## Root Cause (suspected)

In `manual` mode, the plugin does not actively intercept or track assistant messages. When `plannotator last` is called, it appears unable to locate the current session's last message — the server starts and the browser opens, but no content is served to the frontend.

In `plan-agent` mode the plugin actively tracks messages, and `/plannotator-last` works correctly. So the issue is specific to `manual` mode.

## Expected Behavior

`plannotator last` (and `/plannotator-last`) should work regardless of the plugin's workflow mode, since the CLI should be able to find the last message from the active session independently.

## Workaround

Switch to `plan-agent` workflow mode. The downside is that `submit_plan` becomes available to the plan agent, which may not be desired for users who want fully manual control.

snvtac:
## Summary
- route OpenCode runtime selection through a shared helper so `runtime: "cli"` is an explicit first-class CLI path
- avoid reading bundled HTML for CLI `submit_plan`, keeping source checkouts and CLI-only workflows independent of embedded assets
- add OpenCode plugin hook coverage for `/plannotator-last`, `/plannotator-annotate`, `/plannotator-review`, and CLI `submit_plan` without bundled HTML

## Issue
Fixes #947

## Verification
- `bun test apps/opencode-plugin/*.test.ts`
- `PATH="$PWD/apps/hook/node_modules/.bin:$PWD/packages/ui/node_modules/.bin:$PATH" BUN_INSTALL_CACHE_DIR="$PWD/../bun-cache" bun run typecheck`
- `bun test`
- `git diff --check`
- `bun build embedded.ts --outfile dist/embedded.js --target bun --external @opencode-ai/plugin && bun build index.ts --outfile dist/index.js --target node` (run from `apps/opencode-plugin` with the system Bun binary)

### DIFF
--- a/apps/opencode-plugin/index.test.ts
+++ b/apps/opencode-plugin/index.test.ts
@@ -1,0 +1,254 @@
+import { afterEach, describe, expect, mock, test } from "bun:test";
+import { chmodSync, existsSync, mkdtempSync, readFileSync, renameSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+import PlannotatorPlugin from "./index";
+
+const tempDirs: string[] = [];
+const originalPlannotatorBin = process.env.PLANNOTATOR_BIN;
+const originalShareUrl = process.env.PLANNOTATOR_SHARE_URL;
+const originalPasteUrl = process.env.PLANNOTATOR_PASTE_URL;
+
+function makeTempDir(): string {
+  const dir = mkdtempSync(path.join(tmpdir(), "plannotator-opencode-index-"));
+  tempDirs.push(dir);
+  return dir;
+}
+
+function installFakeCli(tempDir: string): {
+  argsFile: string;
+  stdinFile: string;
+} {
+  const argsFile = path.join(tempDir, "args.txt");
+  const stdinFile = path.join(tempDir, "stdin.json");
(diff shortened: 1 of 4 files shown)


backnotprop:
Closing this. #947 wasn't a routing bug, it was a frontend crash (same as #949), already fixed in 0.21.1. The runtime change doesn't touch it. Both paths served the same broken page.

Also, `runtime: "cli"` is a debug switch, not something users should set. Happy to take the `resolveRuntimeExecutionMode` cleanup as its own refactor if you want. Thanks.

PR #968, opened 2026-06-25
fix(vscode): enable clipboard in webview iframe

### LINKED ISSUE #864: [Bug] Content is not pasted from the clipboard in the VS Code extension
When I try to paste text from the clipboard into a Plannotator comment in the VS Code extension, nothing happens.

Copying text from the Plannotator tab also doesn't work.

#### VS Code Info:
- Version: 1.123.0 (Universal)
- Commit: 6a44c352bd24569c417e530095901b649960f9f8
- Date: 2026-06-03T11:29:03+02:00
- Electron: 42.2.0
- ElectronBuildId: 14159160
- Chromium: 148.0.7778.97
- Node.js: 24.15.0
- V8: 14.8.178.14-electron.0
- OS: Darwin arm64 25.1.0

#### Plannotator Info:
- Extension Version: 0.16.6
- Bin Version: 0.19.27

rushelex:
## Problem

Closes #864.

In the VS Code extension, copying text from the Plannotator tab and pasting into a comment both silently fail.

## Root cause

`PanelManager` renders the Plannotator app inside an `<iframe>` within the VS Code webview (`apps/vscode-extension/src/panel-manager.ts`). A VS Code webview is itself a sandboxed iframe, so a *nested* iframe is gated by Permissions Policy. The iframe had no `allow` attribute, which blocks `navigator.clipboard` reads/writes and `paste` events in the embedded app.

The app relies on `navigator.clipboard.writeText` for every copy action (code blocks, tables, export, annotations, …) and on `clipboardData` for paste — all of these are blocked under the default permissions policy of a nested iframe.

## Fix

Grant clipboard access to the iframe:

```html
<iframe id="pn-frame" src="${url}" allow="clipboard-read; clipboard-write"></iframe>
```

## Testing

- Added a `panel-manager.test.ts` case asserting the iframe HTML carries `allow="clipboard-read; clipboard-write"`.
- `bun test src/panel-manager.test.ts` → 3 pass.
- `bun run build:vscode` succeeds.

### DIFF
--- a/apps/vscode-extension/src/panel-manager.test.ts
+++ b/apps/vscode-extension/src/panel-manager.test.ts
@@ -42,6 +42,31 @@ describe("PanelManager", () => {
     );
   });
 
+  it("grants clipboard permission to the iframe", async () => {
+    let capturedHtml = "";
+    const spy = spyOn(vscode.window, "createWebviewPanel");
+    spy.mockImplementation((() => {
+      let disposeListener: (() => void) | null = null;
+      return {
+        webview: {
+          get html() { return capturedHtml; },
+          set html(v: string) { capturedHtml = v; },
+        },
+        reveal() {},
+        dispose() { disposeListener?.(); },
+        onDidDispose(listener: () => void) {
+          disposeListener = listener;
+          return { dispose() {} };
+        },
+      } as unknown as vscode.WebviewPanel;
+    }) as typeof vscode.window.createWebviewPanel);
+    spies.push(spy);
+
+    await manager.open("http://127.0.0.1:9999/review?id=42");
+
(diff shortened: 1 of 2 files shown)


rushelex:
Closing in favor of #970.

This PR added `allow="clipboard-read; clipboard-write"` to the iframe, but
testing in a live VS Code webview showed that's insufficient: the clipboard
failure is `NotAllowedError: Document is not focused`, not a Permissions
Policy block — the nested iframe document never holds focus, so the async
Clipboard API is rejected and native copy/paste events don't fire at all.

The real fix routes the clipboard through the extension host
(`vscode.env.clipboard`) and drives copy/paste off keydown. See #970,
which closes both #864 and #969.

PR #970, opened 2026-06-25
fix(vscode): bridge clipboard and forward keystrokes in webview

### LINKED ISSUE #864: [Bug] Content is not pasted from the clipboard in the VS Code extension

### LINKED ISSUE #969: [Bug] VS Code keybindings (Cmd+P, etc.) don't work while a Plannotator tab is focused
When a Plannotator tab is focused in the VS Code extension, native VS Code
keybindings stop working — e.g. `Cmd+P` (Quick Open), `Cmd+Shift+P`
(Command Palette), and similar global shortcuts do nothing until focus
returns to a regular editor.

#### Root cause

The extension renders the Plannotator app inside a **nested** iframe:

```
VS Code webview (outer sandboxed iframe)
  └─ wrapper HTML (panel-manager.ts)
       └─ <iframe> with the Plannotator app  ← focus lives here
```

VS Code resolves its keybindings by intercepting `keydown` events that
bubble up to the webview. When focus is inside the nested app iframe
(a different origin via `asExternalUri`), `keydown` events don't cross the
iframe boundary, so the webview — and therefore VS Code — never sees them.

There is currently no keyboard forwarding in the extension (no `keydown`
handling in `apps/vscode-extension/src/`).

#### Suggested fix

Adopt the standard VS Code webview pattern: forward keyboard events from the
nested iframe up to the wrapper and re-dispatch a synthetic `KeyboardEvent`
that the webview can pick up. Two small additions to existing injection points:

1. Inside the iframe (`cookie-proxy.ts`): listen for `keydown` and
   `postMessage` the serialized event to the parent.
2. In the wrapper (`panel-manager.ts`): on that message, reconstruct a
   `KeyboardEvent` and `window.dispatchEvent(...)`.

#### VS Code Info:
- Version: 1.125.1 (Universal)
- OS: Darwin arm64 25.1.0

#### Plannotator Info:
- Extension Version: 0.16.6
- Bin Version: 0.21.2

rushelex:
Closes #864. Closes #969.

The Plannotator app runs inside a **nested cross-origin iframe** within the VS
Code webview. Two separate input paths break there, both rooted in the same
constraint — the inner document never holds focus and events don't cross the
iframe boundary:

| Symptom | Issue |
|---|---|
| Copy / cut / paste silently fail (keyboard and app buttons) | #864 |
| VS Code keybindings (Cmd+P, Cmd+Shift+P, …) dead while a Plannotator tab is focused | #969 |

## Root cause

- **Clipboard (#864):** the iframe document is never focused, so the async
  Clipboard API is rejected (`NotAllowedError: Document is not focused`) and
  native `copy`/`cut`/`paste` events never fire. Granting `allow="clipboard-*"`
  on the iframe does **not** help — the rejection is about focus, not
  Permissions Policy (this is why the earlier #968 approach was insufficient).
- **Keybindings (#969):** `keydown` events don't bubble past the nested iframe
  boundary, so VS Code's keybinding service never sees them.

The one input signal the iframe still receives is `keydown`. Both fixes hang off it.

## Fix

- **Clipboard bridge:** route reads/writes through the extension host, which
  owns the system clipboard via `vscode.env.clipboard`. The injected script
  overrides `navigator.clipboard` (for app buttons) and handles Cmd+C/X/V on
  keydown (for native shortcuts), relaying through the wrapper to an
  `onDidReceiveMessage` handler and back.
- **Keystroke forwarding:** all other keys are posted to the wrapper, which
  re-dispatches a synthetic `KeyboardEvent` on the webview document so VS Code
  can resolve its shortcuts. Undo/redo/select-all stay native.

## Testing

- `bun test apps/vscode-extension/src/` → 23 pass.
- `bun run build:vscode` succeeds; `tsc` clean.
- Manually verified in VS Code (1.125.1): Cmd+C / Cmd+V / Cmd+X, app copy
  buttons, and Cmd+P all work in a focused Plannotator tab.

### DIFF
--- a/apps/vscode-extension/mocks/vscode.ts
+++ b/apps/vscode-extension/mocks/vscode.ts
@@ -74,8 +74,14 @@ export const commands = {
   async executeCommand(_command: string, ..._args: unknown[]) {},
 };
 
+export interface Webview {
+  html: string;
+  onDidReceiveMessage(listener: (message: unknown) => void): { dispose(): void };
+  postMessage(message: unknown): Thenable<boolean>;
+}
+
 export interface WebviewPanel {
-  webview: { html: string };
+  webview: Webview;
   iconPath?: Uri;
   reveal(viewColumn?: number): void;
   dispose(): void;
@@ -105,11 +111,15 @@ export const window = {
     _viewType: string,
     _title: string,
     _showOptions: number,
-    _options?: { enableScripts?: boolean },
+    _options?: { enableScripts?: boolean; retainContextWhenHidden?: boolean },
   ): WebviewPanel {
     let disposeListener: (() => void) | null = null;
     return {
-      webview: { html: "" },
+      webview: {
+        html: "",
+        onDidReceiveMessage() { return { dispose() {} }; },
+        postMessage() { return Promise.resolve(true); },
+      },
       reveal() {},
       dispose() {
         disposeListener?.();
@@ -137,6 +147,12 @@ export const window = {
 export const env = {
   async asExternalUri(uri: Uri): Promise<Uri> {
     return uri;
+  },
+  clipboard: {
+    async readText(): Promise<string> {
+      return "";
+    },
+    async writeText(_value: string): Promise<void> {},
   },
 };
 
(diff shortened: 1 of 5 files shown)


backnotprop:
thanks!

PR #729, opened 2026-05-15
migrate pi extension from mariozechner to earendil-works

nulladdict:
mario joined earendil and repo and npm packages have been recently moved

old repo links and npm packages, albeit deprecated, still work for now, so this is not urgent
this is also a breaking change for pi users, since it requires pi 0.74 and up

announcement: https://mariozechner.at/posts/2026-04-08-ive-sold-out/
related pi release: https://github.com/earendil-works/pi/releases/tag/v0.74.0

#726 is also effectively the same, but is missing a repo link change, includes some drive-by package updates and a suspicious `minimumReleaseAgeExcludes` change
feel free to close this and make a combination change instead

PR #984, opened 2026-07-01
fix(pi-extension): hoist @earendil-works/pi-* runtime deps

viadezo1er:
Move `@earendil-works/pi-ai`, `pi-agent-core`, and `pi-tui` from devDependencies to dependencies so they are installed and hoisted alongside @`plannotator/pi-extension`.

These packages are imported at runtime by `index.ts` (e.g. `import { Type } from "@earendil-works/pi-ai"`), but were only listed as devDependencies/peerDependencies. pi installs extensions with peer-dep resolution disabled (--omit=peer / --legacy-peer-deps), so on a clean install or `pi update` the modules never land in the extension's node_modules and pi fails to open with:

```
  Failed to load extension: Cannot find module '@earendil-works/pi-ai'
  Require stack:
  - @plannotator/pi-extension/index.ts
```

The namespace was migrated from `@mariozechner/*` to `@earendil-works/*` in #729, which swapped the imports and devDependency pins but kept them as devDependencies, leaving no guaranteed install path for the runtime imports.

MIT & Apache 2.0 dual-licensing is fine.

### DIFF
--- a/apps/pi-extension/package.json
+++ b/apps/pi-extension/package.json
@@ -14,9 +14,17 @@
   "bugs": {
     "url": "https://github.com/backnotprop/plannotator/issues"
   },
-  "keywords": ["pi-package", "plannotator", "plan-review", "ai-agent", "coding-agent"],
+  "keywords": [
+    "pi-package",
+    "plannotator",
+    "plan-review",
+    "ai-agent",
+    "coding-agent"
+  ],
   "pi": {
-    "extensions": ["./"]
+    "extensions": [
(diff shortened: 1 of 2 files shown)


backnotprop:
Closing this PR because supported Pi clean installs resolve these host APIs through Pi's loader aliases, while moving them into production dependencies duplicates a large provider stack and permits host/extension version skew. There is no linked issue. Please file a fresh issue with an exact supported-version reproduction if this still occurs.

PR #986, opened 2026-07-01
feat(annotate): add sidebar shortcuts

leoreisdias:
## Context

Annotate mode has a left sidebar with useful destinations, but it did not expose keyboard shortcuts for opening those panels. This adds shortcut coverage for the annotate sidebar and makes the shortcuts visible from Settings when the app is in annotate mode.

## What Changed

- Adds an annotate sidebar shortcut scope for toggling the Contents and Files sidebar panels.
- Adds a double-tap Shift shortcut for toggling the Agent control when the annotate Agent TUI is available.
- Passes annotate mode into Settings so the shortcut reference shows annotate-specific actions instead of the plan-review action list.
- Splits the hardcoded shortcut help data into shared plan/annotate sections plus mode-specific action/sidebar sections.

## Review Focus

- Please look hardest at the shortcut choices and event guards in `packages/editor/App.tsx`.
- The Agent TUI shortcut intentionally uses `Shift Shift` after avoiding conflicts with browser shortcuts, Pinpoint/Selection's Option behavior, bare-letter annotation shortcuts, and built-in browser/AI shortcuts.
- I did not keep an Escape shortcut for closing the left sidebar. The shortcuts that open the sidebar already toggle it closed, and Escape created awkward interactions with annotation UI dismissal.

## Design Note

There is some duplication between annotate and plan shortcut definitions in the Settings shortcut display. I leaned slightly away from over-DRYing this for now so each mode can freely evolve its own wording and available actions later. Happy to fold this into a more shared definition model if reviewers prefer that direction.

Relatedly, the new annotate sidebar scope currently lives under `packages/ui/shortcuts/plan-review` because the existing shortcut taxonomy only has plan-review and code-review folders, and annotate mode already reuses much of the plan-review shortcut surface. I am happy to move this into a dedicated annotate shortcut folder if that feels cleaner.

## Test Plan

- [x] `bun test packages/ui/shortcuts.test.ts`
- [x] `PATH="$PWD/packages/ui/node_modules/.bin:$PATH" bun run typecheck`
- [x] `bun run build:review && bun run build:hook`

### DIFF
--- a/packages/editor/App.tsx
+++ b/packages/editor/App.tsx
@@ -102,6 +102,11 @@ import type { AgentTerminalCapability } from '@plannotator/shared/agent-terminal
 import { DEMO_PLAN_CONTENT as DEFAULT_DEMO_PLAN_CONTENT } from './demoPlan';
 import { DIFF_DEMO_PLAN_CONTENT } from './demoPlanDiffDemo';
 import { canUseAnnotateWideMode, resolveWideModeExitLayout, type WideModeLayoutSnapshot, type WideModeType } from './wideMode';
+import {
+  annotateSidebarShortcuts,
+  useAnnotateSidebarShortcuts,
+  useDoubleTapShortcuts,
+} from '@plannotator/ui/shortcuts';
 const USE_DIFF_DEMO =
   import.meta.env.VITE_DIFF_DEMO === '1' ||
   import.meta.env.VITE_DIFF_DEMO === 'true';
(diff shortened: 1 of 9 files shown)


backnotprop:
Nice work — the mode-aware Settings help and the conflict analysis in the PR description were appreciated. One fix pushed to your branch (hope that's OK): the double-tap engine counted *any* two Shift keyups within the window, so the Agent-TUI toggle fired during ordinary Shift usage — two capitalized words, two Shift+Clicks extending a selection, or pressing this PR's own `Mod+Shift+B` twice. A tap now only counts when the key goes down and comes up alone (any other keydown, a pointer interaction mid-press, or a modifier held at release breaks the sequence), and the binding's `preventDefault` flag is now honored in the double-tap path. Full suite + typecheck green. Thanks for the contribution!

leoreisdias:
@backnotprop good catch! I appreciate that!

backnotprop:

PR #996, opened 2026-07-04
fix(annotate): recognize .jsx/.ts/.tsx as html-asset script extensions

Antheurus:
## Problem

`CONTENT_TYPES_BY_EXT` in `packages/shared/html-assets.ts` only recognizes `.js`/`.mjs`, so `rewriteHtmlAssetReferences()` silently leaves `<script src="component.jsx">` (or `.ts`/`.tsx`) untouched instead of rewriting it to `/api/html-assets/...` like every other supported asset type.

Under render-html mode the page is shown in an iframe via `srcDoc` (`HtmlViewer.tsx`), which has no base URL of its own, so any unrewritten relative `<script src>` fails to resolve and the referenced file never loads — with no visible error, since `sandbox="allow-scripts"` happily executes whatever JS *did* load, it just silently has fewer globals defined than the app expects.

This breaks any multi-file browser prototype that loads React/TS components via `<script type="text/babel" src="foo.jsx">` instead of a single bundle — a common pattern for design mockups and Babel-standalone demos (e.g. multiple components split across files, each script tag loading one). The app renders blank with no console-visible signal pointing at the actual cause.

## Fix

Add `.jsx`, `.ts`, `.tsx` to `CONTENT_TYPES_BY_EXT` alongside the existing `.js`/`.mjs` entries, so `<script src>` references to these extensions get the same asset-rewrite treatment. Content-type only — same `application/javascript` type already used for `.js`/`.mjs`, since these files are always loaded with `type="text/babel"` and transpiled client-side regardless of the served content-type.

Added test coverage in `html-assets.test.ts`:
- A dedicated test rewriting `.jsx`/`.ts`/`.tsx` script references
- Extended the content-type coverage test with the three new extensions

## Safety

Purely additive — three new keys in a extension→MIME-type lookup map. For every file extension already recognized (`.js`, `.mjs`, `.css`, images, fonts, etc.), behavior is byte-for-byte unchanged. For any script reference that isn't `.jsx`/`.ts`/`.tsx`/`.js`/`.mjs` (e.g. inline scripts, or `src` with no matching extension), behavior is also unchanged — it still falls through to "leave untouched" exactly as before, per the existing `does not rewrite traversal or unknown extension asset refs` test.

## Verification

- `bun test packages/shared/html-assets.test.ts` — 8/8 pass (2 new)
- `bun test` (full suite) — 1845 pass, 38 skip, 0 fail
- Full typecheck chain (`packages/shared`, `packages/ai`, `packages/server`, `packages/ui`, `apps/pi-extension`) — clean

## Changed files
- `packages/shared/html-assets.ts` — 3 insertions
- `packages/shared/html-assets.test.ts` — 17 insertions


Antheurus:
Closing this — withdrawing for now, not planning to follow up. Thanks for the great tool regardless.

PR #997, opened 2026-07-05
feat: guide per-file summaries + GitHub Copilot CLI agent engine

backnotprop:
## What

Two additions to the agent-job stack that shipped in #990/#993/#994:

### 1. Per-file summaries in Guided Review

Each `diffs[]` entry in the guide schema now carries a required `summary`: 1–2 sentences describing the semantic change in that file, written **from the diff hunks alone** — the prompt explicitly forbids per-file investigation (no opening files, no searches), keeping guide generation fast. Rendered as a muted line above each diff in the guide screen.

- Schema-enforced for Claude/Codex (`--json-schema` / `--output-schema`); prompt-enforced via the marker contract for Cursor/OpenCode/Pi/Copilot.
- Tolerant ingestion: a missing/blank/mistyped summary is dropped silently — never drops the file ref, never fails the guide. Old guides and summary-less marker output stay valid everywhere (automatic ingestion, manual repair paste, mechanical JSON repair).
- Repair prompts are instructed to fill a missing required field with an empty string rather than fabricating captions for files they can't see.

### 2. GitHub Copilot CLI as an agent-job engine

`copilot` joins Cursor/OpenCode/Pi as the fourth **marker engine**, launchable for both code review and guided review when the binary is on PATH:

- `copilot --output-format json` JSONL stream; assembled text read from `assistant.message` events (deltas skipped, same shape as Pi's `message_end`).
- Non-interactive posture: `--no-ask-user` (unallowed tools auto-deny cleanly), `--deny-tool=write`, shell allowlist limited to `git`/`gh`/`glab`/`jj`/`wc`, builtin MCP server + auto-update disabled. Sits between Claude (full subcommand granularity) and the other markers (unrestricted Bash) on the trust spectrum.
- Model catalog discovered from `copilot help config` (no models subcommand exists); fails closed to a Default-only picker.
- UI: engine row + per-surface model settings in the Agents tab and the guide launcher, official Copilot mark as a single `currentColor` icon.
- Refactor: exported `MarkerEngineId` and replaced every hand-repeated `"cursor" | "opencode" | "pi"` cast across both server runtimes — the next engine is a two-edit change.

## Verification

- 1887 tests, 0 failures; typecheck green across all packages incl. the Pi vendor step.
- Copilot verified **live end-to-end**: composed review prompt through the real `copilot` binary with the exact production argv → marker block parsed → seeded bug found with correct file/line/severity.
- Permission posture probed against the real CLI: denied tools return a clean `denied` result (no hang), `gh` network calls work under the shell allowlist.

## Notes

- Copilot's `--effort` knob is deliberately not surfaced yet; easy follow-up if wanted.
- Tour remains Claude/Codex only (unchanged).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -327,7 +327,7 @@ During normal plan review, an Archive sidebar tab provides the same browsing via
 | `/api/external-annotations` | POST | Add external annotations (single or batch `{ annotations: [...] }`) |
 | `/api/external-annotations` | PATCH | Update fields on a single annotation (`?id=`) |
 | `/api/external-annotations` | DELETE | Remove by `?id=`, `?source=`, or clear all |
-| `/api/agents/capabilities` | GET | Check available agent providers (claude, codex, tour, guide, cursor, opencode, pi) |
+| `/api/agents/capabilities` | GET | Check available agent providers (claude, codex, tour, guide, cursor, opencode, pi, copilot) |
 | `/api/agents/review-profiles` | GET | List launchable review profiles (enabled skills + builtin default) |
 | `/api/agents/skills` | GET | List all discovered skills for the add-a-review picker (each flagged `enabled`) |
 | `/api/agents/review-skills` | POST | Enable a skill as a review (body: `{ name }`); writes `review-skills.json` |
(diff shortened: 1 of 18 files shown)


vazquezjosue182001-arch:
Csm

vazquezjosue182001-arch:
Neta csm

PR #957, opened 2026-06-23
Make the document UI reusable as published building blocks

backnotprop:
## Make the document UI reusable as published building blocks

Extracts Plannotator's document UI (`@plannotator/ui`) into installable, host-overridable building blocks so a separate app (the commercial Workspaces/Enterprise app) can reuse the rendering, theme, editor, settings, comments, and layout — **without changing Plannotator's own behavior at all.**

> Supersedes the reverted from-scratch attempt (ADRs 002/003). The corrected approach is ADR 004: **share by publishing, move + decouple, never rewrite, never delete working code until a human confirms parity.**

### THE LAW (how to read this diff)
Every change is **move + decouple**, never rewrite. For each spot where a shared component called a hard-coded `/api/*` route or a Plannotator-only global, the I/O was lifted to an **optional seam** (a module-level `setX()` setter or an optional prop) whose **default reproduces today's behavior exactly**. Plannotator passes nothing and stays byte-for-byte identical; a host injects its own backend. Nothing Plannotator-specific was deleted.

### What's in scope
- **`@plannotator/core` (new):** a small, browser-safe, zero-dependency package of pure utilities + types, carved out of `@plannotator/shared` so `@plannotator/ui` can be published without dragging the Node/git/server kitchen sink. `@plannotator/shared` re-exports each moved module via one-line shims, so all ~99 internal import sites are untouched.
- **`@plannotator/ui`:** ~13 host-override seams across Phases 2–6 (image, storage, doc-preview, file-tree, identity, drafts, live comments, versions/diff, settings sync, save-to-notes, Ask AI), a single `configurePlannotatorUI()` front door over the global seams, a `loadFromBackend()` settings-rehydration hook, and a precompiled `styles.css` build.
- Two override-path bug fixes + 16 per-seam override tests.

### Phases (each verify-gated, defaults = today's behavior)
| Phase | What became host-overridable |
|---|---|
| 1 | Packaging unblock (peer deps, files allowlist) — no runtime change |
| 2 | Image resolver, settings storage backend |
| 3 | Markdown mode, code-path validation, doc-preview fetch, scroll provider |
| 4 | File-tree backend |
| 5 | Identity, draft transport, live-comment transport |
| 6 | Version/diff fetchers, vscode-diff, settings sync, save-to-notes, Ask-AI transport (+ diff/annotation CSS relocated into the package) |
| 7 | Carve `@plannotator/core`, complete the settings provider, `configure()` front door, precompiled CSS, publish-prep |

### Verification
[...]

PR #1007, opened 2026-07-06
fix(review): render tables + inline video in PR comments, descriptions, and commit messages

backnotprop:
Found while dogfooding: reviewing PR #957 inside Plannotator, a PR comment containing a markdown table rendered as one collapsed line of pipes, and links to test-recording videos were plain links.

**Root cause:** the lightweight `MarkdownBody` renderer (used by the PR-comments panel, PR description tab, and commit-description header — NOT the main document viewer, which has a full `TableBlock`) had no `table` case, so table blocks fell through to the paragraph branch.

**Changes:**
- `table` case reusing `parseTableContent` from `@plannotator/ui`'s TableBlock (single pipe parser), compact styling, inline markdown in cells
- A paragraph that is exactly one link/bare URL to a video file now renders an inline `<video controls>` player with a fallback link
- `MarkdownBody` extracted to its own module (it served three components from inside `PRSummaryTab`); importers re-pointed
- New `markdownBody.test.tsx` (5 tests, renderToString — runs in plain CI)

**Verification:** typecheck clean · full suite 1897 pass / 0 fail · `apps/review` builds.

### DIFF
--- a/packages/review-editor/components/CommitDescriptionHeader.tsx
+++ b/packages/review-editor/components/CommitDescriptionHeader.tsx
@@ -1,7 +1,7 @@
 import React, { useState } from 'react';
 import type { CommitDiffInfo } from '@plannotator/shared/types';
 import { Avatar } from './Avatar';
-import { MarkdownBody } from './PRSummaryTab';
+import { MarkdownBody } from './MarkdownBody';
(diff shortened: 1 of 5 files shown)


backnotprop:

PR #1008, opened 2026-07-07
fix(hook): echo tool_input as updatedInput so plan approval survives Claude Code 2.1.199+

### LINKED ISSUE #995: Plan Approve silently ignored on Claude Code >= 2.1.200: PermissionRequest allow without updatedInput is discarded for ExitPlanMode
## Summary

Since Claude Code **2.1.200**, clicking **Approve** in the Plannotator plan review UI no longer returns control to the CLI. The page closes as usual, but the CLI falls back to the built-in interactive plan-approval dialog ("Would you like to proceed? 1. Yes, and bypass permissions…"), so the user has to approve a second time in the terminal. **Deny / Send Feedback still works.**

## Environment

- Plannotator 0.21.4 (plugin + binary, versions consistent)
- Claude Code 2.1.200 / 2.1.201, Windows 11
- Hook: `PermissionRequest` / matcher `ExitPlanMode` (stock plugin `hooks/hooks.json`)

## Evidence (transcript timeline)

Scanning `~/.claude/projects/*/**.jsonl` for `hook_permission_decision` attachments:

- Last `allow` delivered by the hook: **Jul 2, 18:54** — session on Claude Code **2.1.198**.
- After upgrading to **2.1.200** (Jul 3): zero `allow` decisions delivered, while `deny` decisions continue to land normally (multiple confirmed on 2.1.200 and 2.1.201).
- Every "approval" after the upgrade shows no `hook_permission_decision` attachment at all — the plan gets approved via the CLI's own fallback dialog instead.

## Root cause

Claude Code 2.1.200 added a guard in its PermissionRequest hook result handling (from the bundled CLI, prettified):

```js
if (l.behavior === "allow") {
  let c = l.updatedInput || n;                    // n = original tool_input
  if (l.updatedInput) { /* re-check permission rules with updated input */ }
  if (!mir(e, l.updatedInput) && e.requiresUserInteraction?.()) return;  // <-- discards the allow
  ...
  return { behavior: "allow", updatedInput: c, ... };
}
```

with

```js
function mir(e, t) { return t !== void 0 && !Dw(e); }   // Dw(e) = tool is MCP
```

`ExitPlanMode.requiresUserInteraction()` returns `true`, and it is not an MCP tool. So for `ExitPlanMode`, a hook `allow` **without `updatedInput` is silently dropped** (`return;` → treated as "hook gave no decision" → interactive dialog). The `deny` branch is untouched, which matches the observed behavior exactly.

Plannotator's approve output (`apps/hook/server/index.ts`, plan flow) currently emits:

```json
{"hookSpecificOutput":{"hookEventName":"PermissionRequest","decision":{"behavior":"allow"}}}
```

— no `updatedInput`, so Claude Code ≥ 2.1.200 ignores it. (Note: the `updatedPermissions` from the approve are dropped too, since the guard returns before they are applied.)

## Suggested fix

Echo the original `tool_input` back as `updatedInput` in the allow decision:

```js
console.log(JSON.stringify({
  hookSpecificOutput: {
    hookEventName: "PermissionRequest",
    decision: {
      behavior: "allow",
      updatedInput: hookInput.tool_input,   // <-- echo the plan input from stdin
      ...(updatedPermissions.length > 0 && { updatedPermissions }),
    },
  },
}));
```

## Workaround (verified)

A wrapper hook script that pipes stdin to `plannotator`, parses its stdout, and injects `updatedInput` (= `tool_input` from stdin) into the allow decision restores Approve on 2.1.200/2.1.201 without touching the binary.

## Related UX note

When the decision is dropped, the browser UI still shows "approved" and auto-closes (`handleApprove` in `packages/editor/App.tsx` treats any fetch completion as success and its `catch` is silent), so the user gets no signal that the approval never reached the CLI. A toast on non-OK delivery would make this class of failure visible.

flex-yj-kim:
## Problem

Since **Claude Code 2.1.199**, clicking **Approve** in the plan review UI no longer returns control to the CLI — the page closes but the built-in plan-approval dialog reappears in the terminal. **Deny is unaffected.** Reproduced on 2.1.199–2.1.202.

## Cause

2.1.199 added a guard that **discards a `PermissionRequest` `allow` for `ExitPlanMode` when `updatedInput` is absent** (ExitPlanMode requires user interaction and isn't an MCP tool). The hook emitted a bare `allow`, so it was dropped — along with `updatedPermissions`. The deny path skips this guard, matching the symptom.

## Fix

Echo the original `tool_input` (in scope as `event.tool_input`) back as `decision.updatedInput`. Backward compatible — a verified no-op on 2.1.198, so safe to ship unconditionally.

## Verification

Built the hook binary with this change, ran it against a mock `ExitPlanMode` event, and drove `/api/approve`; the decision now carries `updatedInput` plus `updatedPermissions`. Independently confirmed on 2.1.201 in #995 and anthropics/claude-code#74256.

Fixes #995

### DIFF
--- a/apps/hook/server/index.ts
+++ b/apps/hook/server/index.ts
@@ -1981,6 +1981,11 @@ if (args[0] === "sessions") {
             hookEventName: "PermissionRequest",
             decision: {
               behavior: "allow",
+              // Echo the original tool_input as updatedInput. Claude Code
+              // >= 2.1.199 silently drops an allow decision for ExitPlanMode
+              // (a tool requiring user interaction) when updatedInput is
+              // absent, falling back to the built-in approval dialog.
+              updatedInput: event.tool_input,
               ...(updatedPermissions.length > 0 && { updatedPermissions }),
             },
           },


backnotprop:

PR #1019, opened 2026-07-08
feat(annotate): keep sidebar shortcuts visible with Agent TUI

leoreisdias:
## Context

When the annotate Agent TUI is open and the Content/Files sidebar is closed, the collapsed sidebar shortcuts should still give people a quick way back to those tabs. The previous behavior hid the shortcuts while the Agent TUI was open, which forced people to collapse the terminal first.

## What Changed

- Keep the collapsed left-sidebar shortcut rail visible whenever the left sidebar is closed.
- When the Agent TUI is open, position that rail on the Agent TUI divider using the same `--agent-terminal-w` width variable that drives the resizable terminal.
- Allow `SidebarTabs` to receive an inline style so placement can be controlled by the parent layout without duplicating the tab UI.
- Preserve the document-side inset while the collapsed rail is visible, so the shortcuts do not cover the document edge.

## Screenshot

image

All closed:
image

## Review Focus

The important behavior to check is the collapsed shortcut rail placement in annotate mode:

- Agent TUI closed: shortcuts remain on the far-left document edge.
- Agent TUI open: shortcuts sit on the Agent TUI border/divider and move with terminal resizing.
- Left sidebar open: the collapsed rail is hidden as before.

## Test Plan

- [x] `PATH="$PWD/node_modules/.bun/typescript@5.8.3/node_modules/.bin:$PATH" bun run typecheck`
- [x] `bun run build:review && bun run build:hook`
- [x] Plannotator code review completed with no changes requested


leoreisdias:
@backnotprop ,I saw that u did some fixes on this one. Let me know if you need anything I could help! Thx!

PR #1022, opened 2026-07-08
Fix annotate terminal startup in large folders

backnotprop:
## Summary
- exclude `.claude/` and `.agents/` from annotate folder file-browser scans
- cap folder file-browser scans to avoid monopolizing the Bun server on huge trees
- make the annotate WebTUI panel cancel cleanly before `onReady` and use the DOM renderer

## Verification
- `bun test packages/server/agent-terminal.test.ts packages/shared/reference-common.test.ts packages/server/reference-handlers.test.ts`
- `bun run typecheck`
- live smoke: `plannotator annotate .` on this repo; `/api/reference/files` returned 189 files quickly and Pi spawned through WebTUI


backnotprop:

PR #1024, opened 2026-07-08
Disambiguate OpenCode model dropdown labels by provider

### LINKED ISSUE #988: OpenCode provider: model dropdown shows duplicate labels when multiple providers share model names
When using Ask AI with the OpenCode provider, the model dropdown only shows model names without any provider context. OpenCode supports multiple model providers (e.g. DeepSeek official, OpenRouter, etc.), and these providers can offer models with **identical names** — for example both `deepseek/deepseek-v4-pro` and `openrouter/deepseek-v4-pro` display as `deepseek-v4-pro` in the dropdown, making them impossible to tell apart.

In [packages/ai/providers/opencode-sdk.ts](https://github.com/backnotprop/plannotator/blob/9bded3432437c18f33b464eb40e1ed050382d84f/packages/ai/providers/opencode-sdk.ts#L225), `fetchModels()` constructs the model entry as:

```ts
{
  id: `${provider.id}/${model.id}`,   // correct, e.g. "deepseek/deepseek-v4-pro"
  label: model.name ?? model.id        // only model name, loses provider context
}
```

The model `id` contains `provider.id/model.id`, but the `label` shown in the UI dropdown is just the bare model name.

yusufemreboyraz:
## Summary
- `fetchModels()` in `packages/ai/providers/opencode-sdk.ts` built each model's dropdown `label` from `model.name ?? model.id` alone.
- OpenCode lets multiple providers expose models with the same name (e.g. `deepseek-v4-pro` via both DeepSeek and OpenRouter). The `id` field already disambiguates them (`${provider.id}/${model.id}`), but the `label` shown in the dropdown did not, so identical entries appeared side by side with no way to tell them apart.
- Fix: append the provider's display name to the label, e.g. `deepseek-v4-pro (DeepSeek)` vs. `deepseek-v4-pro (OpenRouter)`.

## Test plan
- [x] `bun x tsc --noEmit -p packages/ai/tsconfig.json` — clean.
- [x] `bun test packages/ai/ai.test.ts` — 102/102 pass.
- [x] `bun test` (full suite) — same pass/fail counts as `main` (5 pre-existing failures unrelated to this change, caused by missing generated codegen files, reproduced identically on `main` without this diff).

Fixes #988

### DIFF
--- a/packages/ai/providers/opencode-sdk.ts
+++ b/packages/ai/providers/opencode-sdk.ts
@@ -228,7 +228,11 @@ export class OpenCodeProvider implements AIProvider {
 				for (const model of Object.values(provider.models)) {
 					models.push({
 						id: `${provider.id}/${model.id}`,
-						label: model.name ?? model.id,
+						// include the provider name: different providers can expose
+						// models with the same name (e.g. deepseek-v4-pro via
+						// DeepSeek and via OpenRouter), which would otherwise be
+						// indistinguishable in the dropdown
+						label: `${model.name ?? model.id} (${provider.name})`,
 					});
 				}
 			}


backnotprop:

PR #1027, opened 2026-07-09
Add file browser filtering

backnotprop:
## Summary
- Add a compact filter control to the Files sidebar
- Filter the loaded file tree client-side while preserving folder context and active file behavior
- Add focused tests for file-tree filtering semantics

## Tests
- bun test packages/ui/components/sidebar/FileBrowser.test.ts
- bun run --cwd packages/ui typecheck
- git diff --check


backnotprop:

PR #1028, opened 2026-07-09
feat(ui): handle-wide click-to-collapse + suppressible hover on resize handles

backnotprop:
## What

Reworks the sidebar/panel resize-handle interaction across the plan editor and review editor, and adds host-facing seams to `@plannotator/ui` so the commercial Workspaces app can drive its own edge UX.

### New UX (Plannotator, all handles)
- **No hover highlight** on the track bar.
- **Cursor-following tooltip**: "Click to close · Drag to resize".
- **Single click anywhere on the handle collapses** the panel (drag still resizes).
- **Double-click reset** is now unreachable (the single click collapses first).
- **Collapse chevron retained** on hover.

### New seams in `@plannotator/ui` (default-off, Plannotator opt-in)
- `useResizablePanel({ onClick, clickThreshold })` — the hook owns the pointer-capture drag state machine, so this is the only reliable way to distinguish a click from a drag-start. `onClick` fires on pointer-up only when travel stayed under `clickThreshold` (default 4px). Not fired on snap-close.
- `ResizeHandle` props: `hideHoverTrack`, `tooltip`, `trackClassName`, plus a `[data-resize-track]` attribute (same host-CSS pattern as `[data-collapse]`).

## Why
Workspaces is reworking sidebar-edge interactions (no hover reveal, tooltip, click-to-collapse). Two capabilities couldn't be done cleanly from the host side — targeting/suppressing the hardcoded hover track, and telling a click apart from a drag. Both now have seams; Workspaces builds its own tooltip and double-click removal on top.

## Compatibility
- New options/props are all optional and default to today's behavior. When nothing is passed, existing consumers are unchanged (the click-vs-commit branch is gated on `onClick` being present).
- No version bump — that happens at release time.

## Test
- Plan editor + review editor dev servers: hover a divider (no highlight, tooltip follows cursor), single-click (collapses), drag (resizes), chevron on hover.
- `packages/ui` typecheck + strict-consumer typecheck pass; ui test suite green (349 pass / 0 fail).

### DIFF
--- a/packages/editor/App.tsx
+++ b/packages/editor/App.tsx
@@ -250,6 +250,9 @@ const feedbackLossDescription = (annotationCount: number, hasDirectEdits: boolea
 
 type SourceFileEditWarningAction = 'send-feedback' | 'approve' | 'close';
 
+/** Hint shown following the cursor while hovering a sidebar/panel resize handle. */
+const RESIZE_HANDLE_TOOLTIP = 'Click to close · Drag to resize';
+
 const App: React.FC = () => {
   const [markdown, setMarkdown] = useState(DEMO_PLAN_CONTENT);
   const [annotations, setAnnotations] = useState<Annotation[]>([]);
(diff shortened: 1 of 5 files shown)


backnotprop:

PR #1034, opened 2026-07-10
fix(opencode): preserve planning handoff

franktronics:
## Summary

- Prevent the OpenCode `plan` agent and configured planning agents from receiving the `Proceed with implementation` handoff after plan approval.
- Keep the implementation handoff unchanged for `build` and other non-planning agents.
- Emit a visible status message when approval targets a planning agent:

  `Plan approved. Plan mode remains active; no implementation has been requested.`

## Related issues

Regression from #648: the agent picker correctly removed the implementation directive for `No switch`, but selecting `plan` still injected `Proceed with implementation`.

#648 also addressed #159, which requested approval without automatically starting implementation. This change preserves that behavior when the selected target is a planning agent.

## Changes

| File | Change |
| --- | --- |
| `apps/opencode-plugin/workflow.ts` | Adds a planning-target predicate for implementation handoffs. |
| `apps/opencode-plugin/index.ts` | Sends the implementation prompt only to non-planning agents; planning agents receive a non-executing status message instead. |
| `apps/opencode-plugin/workflow.test.ts` | Covers `build`, `plan`, custom planning agents, and regular non-planning agents. |

## Test plan

- [x] `bun test apps/opencode-plugin`
- [x] `bun run build:opencode`
- [x] Manual: selecting `plan` keeps plan mode active and displays the status message.
- [x] Manual: selecting `build` still sends `Proceed with implementation`.

### DIFF
--- a/apps/opencode-plugin/index.ts
+++ b/apps/opencode-plugin/index.ts
@@ -44,6 +44,7 @@ import {
   shouldModifyPrompts,
   shouldRegisterSubmitPlan,
   shouldRejectSubmitPlanForAgent,
+  shouldStartImplementationForAgent,
   type RuntimeMode,
   type PlannotatorOpenCodeOptions,
 } from "./workflow";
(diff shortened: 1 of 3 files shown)


backnotprop:
thanks, in review

PR #1035, opened 2026-07-10
fix(ui): keep Ask AI input visible after responses

dmmulroy:
## Summary

- keep the Ask AI follow-up composer inside the sidebar after a long response completes
- replace the chat root's `h-full` sizing with a bounded `flex-1 min-h-0` layout

## Root cause

The chat root was a shrinking flex child with `h-full` and an automatic minimum height. While the short streaming state could shrink beneath the sidebar header, a sufficiently long completed response made the automatic minimum height retain the full panel height. The header plus that full-height child pushed the provider bar and textarea below the viewport.

## Verification

- `bun run --cwd apps/review build && bun run build:hook`

A 10-second headed-Chrome proof was recorded locally as `artifacts/ask-ai-input-fix-proof.mp4` (kept out of git to avoid committing a generated 6.3 MB binary).


packages/ui/components/ai/DocumentAIChatPanel.tsx
```diff
@@ -84,7 +84,7 @@ export const DocumentAIChatPanel: React.FC<DocumentAIChatPanelProps> = ({
   }, [generalInput, onAskGeneral]);
 
   return (
-    <div className="flex flex-col h-full">
+    <div className="flex flex-col flex-1 min-h-0">
```
backnotprop, line 87: [praise] testing this review from something new

backnotprop (approved):

PR #1040, opened 2026-07-11
fix(ai): swallow async EPIPE on codex app-server stdin (crashes host)

### LINKED ISSUE #1039: Plan-review crashes pi with unhandled EPIPE from Codex app-server stdin
## Summary

Opening the plan-review page crashes the entire `pi` process with an **unhandled `write EPIPE`** originating from the Codex app-server provider's `stdin`. The pi session dies mid-tool-call (the `plannotator_submit_plan` tool call is written to the session log with **no** tool result), so it reads as a hard crash rather than a caught error.

## Environment

- `@plannotator/pi-extension` **0.23.0**
- `codex-cli` **0.22.0** (`/opt/homebrew/bin/codex`, on `PATH`)
- Node **v22.19.0**
- macOS (arm64)
- pi agent 0.80.6, model glm-5.2 (zai)

## Stack trace

```
### UNCAUGHT EXCEPTION ###
Error: write EPIPE
    at afterWriteDispatched (node:internal/stream_base_commons:159:15)
    at writeGeneric (node:internal/stream_base_commons:150:3)
    at Socket._writeGeneric (node:net:966:11)
    at Socket._write (node:net:978:8)
    at writeOrBuffer (node:internal/streams/writable:572:12)
    at Writable.write (node:internal/streams/writable:510:10)
    at CodexAppServerProcess.send (generated/ai/providers/codex-app-server.ts:442:21)
    at CodexAppServerProcess.sendAndWait (codex-app-server.ts:448:12)
    at CodexAppServerProcess.doStart (codex-app-server.ts:366:18)
    at CodexAppServerProvider.fetchModels (codex-app-server.ts:550:7)
```

## Root cause

`startPlanReviewServer` calls `createPiAIRuntime()`, which registers the Codex provider whenever `codex` is found on `PATH` and eagerly calls `fetchModels()`. That spawns the codex app-server and writes an `initialize` message to its `stdin`.

With an incompatible codex version (here codex-cli 0.22.0), the child closes its end of the pipe, so the write fails **asynchronously** with `EPIPE`. The spawned process' `stdin` has **no `error` listener**, so Node escalates the stream error to `uncaughtException` and kills the whole pi process.

The existing `.catch(() => {})` on `fetchModels()` in `server/ai-runtime.ts` does **not** cover this, because the failure surfaces as an `'error'` event on the stdin stream, not as a promise rejection.

`send()` only guards against a `destroyed` stream, which does not prevent the async EPIPE:

```ts
send(message: RpcMessage): void {
  if (!this.proc?.stdin || this.proc.stdin.destroyed) return;
  this.proc.stdin.write(`${JSON.stringify(message)}\n`);
}
```

## Proposed fix

Attach a no-op `error` listener to the child's `stdin` right after spawn (single choke point for every `send()` caller). In `generated/ai/providers/codex-app-server.ts`, after `this.proc = proc;`:

```ts
this.proc = proc;
// Swallow async EPIPE on stdin — codex may close the pipe mid-write
// (e.g. incompatible app-server version). Without this listener the
// 'error' event escalates to uncaughtException and kills the host process.
proc.stdin?.on("error", () => {});
proc.once("exit", () => {
  this.handleProcessEnd(new Error("Codex app-server exited unexpectedly"));
});
```

With the listener, the write error is handled, the process still ends cleanly via the `exit`/timeout paths, `fetchModels()` rejects and is swallowed by the existing `.catch()`, and pi keeps running. Verified locally: the same repro that produced the uncaught EPIPE now completes cleanly (all plan-review endpoints return 200).

## Repro

1. Have an incompatible `codex` on `PATH` (codex-cli 0.22.0 reproduces it reliably).
2. In a pi session with the extension, write a markdown plan and call `plannotator_submit_plan`.
3. The plan-review page opens, `createPiAIRuntime` probes codex, and pi dies with the EPIPE above.

jash90:
Fixes #1039.

## Problem

Opening the plan-review page crashes the entire host process (`pi`) with an unhandled `write EPIPE` from the Codex app-server provider's `stdin`. The session dies mid tool-call.

```
Error: write EPIPE
    at CodexAppServerProcess.send (packages/ai/providers/codex-app-server.ts)
    at CodexAppServerProcess.sendAndWait
    at CodexAppServerProcess.doStart
    at CodexAppServerProvider.fetchModels
```

## Root cause

`createPiAIRuntime()` registers the Codex provider whenever `codex` is on `PATH` and eagerly calls `fetchModels()`, which spawns the app-server and writes `initialize` to its stdin. With an incompatible codex version (reproduced with codex-cli 0.22.0) the child closes its pipe end, so the write fails **asynchronously** with `EPIPE`. The spawned `stdin` has no `error` listener, so Node escalates the stream error to `uncaughtException` and kills the host.

The existing `.catch(() => {})` on `fetchModels()` does not cover this — the failure is an `'error'` event on the stdin stream, not a promise rejection.

## Fix

Attach a no-op `error` listener to the child's stdin right after spawn — a single choke point for every `send()` caller. The process still ends via the existing `exit`/timeout paths, `fetchModels()` rejects into the existing `.catch()`, and the host keeps running.

Verified locally: the same repro that produced the uncaught EPIPE now completes cleanly (all plan-review endpoints return 200).

## Note

`generated/` copies are build artifacts; this changes the `packages/` source only.

### DIFF
--- a/packages/ai/providers/codex-app-server.ts
+++ b/packages/ai/providers/codex-app-server.ts
@@ -332,6 +332,10 @@ class CodexAppServerProcess {
     }
 
     this.proc = proc;
+    // Swallow async EPIPE on stdin — codex may close the pipe mid-write
+    // (e.g. incompatible app-server version). Without this listener the
+    // 'error' event escalates to uncaughtException and kills the host process.
+    proc.stdin?.on("error", () => {});
     proc.once("exit", () => {
       this.handleProcessEnd(new Error("Codex app-server exited unexpectedly"));
     });


jash90:
Closing — root cause turned out to be a stale local codex 0.22.0 (no `app-server` subcommand) shadowing a current codex on PATH, not a defect in this repo. Resolved by fixing the environment. Sorry for the noise.

Issue #899, opened 2026-06-13
Remote and multi-session session router: URL surfacing, decision receipt, and host launch behavior


backnotprop:
Provenance stamp: This issue was triaged on June 13, 2026 as part of the older-than-two-month GitHub issue triage exercise.

Source issues: #102, #154, #158, #192, #518

## Summary

Several old issues point at the same underlying problem: Plannotator still needs a dependable session router for users who run multiple agent sessions, remote sessions, non-interactive commands, or host environments that cannot open a normal browser tab.

The common requirement is not only "open a URL." The requirement is that the right plan or review session is visible, reachable, and connected back to the agent process that is waiting for the user's decision.

## What the source issues requested

- #102 requested support for concurrent Claude Code sessions without fixed-port conflicts. It suggested either a port list/range or a single UI that can multiplex active plans.
- #154 reported that `plannotator review` launched from a Claude Code skill could start the server but fail to open the browser in a non-TTY environment. It asked for browser opening to be reliable or for the URL to be printed in a way the caller can use.
- #158 requested a multi-instance daemon that can collect plans from many Claude sessions, show which sessions are waiting, replace repeated plans from the same session, and avoid disruptive popups.
- #192 requested a usable remote workflow for SSH and cloud sessions, including a local-browser URL and a safe return path from the browser decision to the original waiting session.
- #518 reported host-specific behavior with cmux, where Plannotator opened a new small pane instead of a useful tab and did not auto-close correctly.

## Triage determination

These are one product area. Treating them separately would lead to partial fixes that help one host but leave the core session-lifecycle problem intact. The replacement issue should track a router-level design that covers local, remote, multi-session, and host-specific launch behavior.

## Acceptance criteria

- One long-lived Plannotator runtime can host multiple active plan, review, and annotate sessions.
- Each session has a stable identity that includes enough context to prevent approving the wrong waiting agent process.
- Remote sessions expose a usable URL through the best available host channel.
- A browser decision returns a clear receipt such as `applied to waiting session`, `stale plan`, or `session no longer waiting`.
- Non-TTY and skill-launched commands expose a useful URL instead of silently relying on browser launch.
- Repeated plans from the same agent session update or replace the prior waiting session instead of opening confusing duplicate views.
- Host launch policy is explicit for browser, Glimpse, VS Code, cmux, SSH, headless, and no-browser environments.
- The old source issues can be closed after this issue is created and their specific requirements are represented here.

## Notes

[...]

PR #1042, opened 2026-07-13
feat(server): support bounded port ranges

iurysza:
## Summary

- accept inclusive `PLANNOTATOR_PORT` ranges such as `19432-19463`
- try each port in order and bind the first available one in Bun and Pi runtimes
- preserve existing fixed-port retries and random-port defaults

## Validation

- focused port tests: 42 passed, 0 failed
- syntax bundle of all changed runtime files
- `git diff --check`

Related to #899


backnotprop:
thanks, in review

PR #1044, opened 2026-07-13
fix visual-explainer Mermaid theme colors

### LINKED ISSUE #1043: plannotator-visual-explainer emits unsupported OKLCH Mermaid themeVariables
## Summary

The `plannotator-visual-explainer` skill explicitly tells agents to pass `oklch(...)` color strings into `mermaid.initialize({ themeVariables })`.

Mermaid 11's theme processing does not accept those values. As a result, otherwise-valid Mermaid diagrams render as the generic error SVG:

> Syntax error in text

This is **not** a Mermaid diagram-syntax problem, and OKLCH is not generally invalid in the generated HTML. Modern browsers accept the existing OKLCH values in page CSS. The failure is specifically caused by passing OKLCH into Mermaid's own theme/color processing.

## Faulty instruction

The current upstream reference contains this exact example:

https://github.com/backnotprop/plannotator/blob/main/apps/skills/extra/plannotator-visual-explainer/references/theme-override.md#L82-L100

```js
mermaid.initialize({
  theme: "base",
  themeVariables: {
    primaryColor: "oklch(0.50 0.25 280)",
    primaryTextColor: "oklch(1 0 0)",
    primaryBorderColor: "oklch(0.88 0.01 260)",
    lineColor: "oklch(0.40 0.02 260)",
    secondaryColor: "oklch(0.92 0.01 260)",
    tertiaryColor: "oklch(0.92 0.01 260)"
  }
});
```

The wrapper skill requires agents to read that reference and says the Plannotator color layer overrides the delegated `nicobailon/visual-explainer` palette. The delegated visual-explainer's Mermaid template already uses compatible hex values, but this Plannotator override replaces them with the failing OKLCH values.

## Reproduction

1. Install or update `plannotator-visual-explainer` from `backnotprop/plannotator`.
2. Ask it to generate any architecture explainer containing Mermaid.
3. Use the instructed `themeVariables` above with Mermaid 11.
4. Render even a trivial diagram:

```mermaid
flowchart TD
  A["Provider"] --> B["Database"]
```

### Actual

Mermaid returns an error SVG with `aria-roledescription="error"` and `Syntax error in text`.

### Control

Replace only the OKLCH `themeVariables` with hex colors. The same Mermaid source renders successfully.

### Expected

The skill should preserve OKLCH for normal page CSS, but supply Mermaid-compatible literal colors (for example hex) to `themeVariables`.

Mermaid's own theming documentation demonstrates literal hex colors for these fields:
https://mermaid.js.org/config/theming.html#customizing-themes-with-themevariables

## Why this keeps recurring

This is in the upstream skill source, not merely in one generated artifact. Every agent following the skill is explicitly instructed to reproduce the invalid configuration, and refreshing the installed skill from `backnotprop/plannotator` restores the same reference. The example was introduced with the standalone skill/HTML annotation work in commit `1eb561551ca1` and remains present on `main`.

Both repo-local and global installations therefore reproduce the same behavior.

## Suggested acceptance criteria

- Keep Plannotator's OKLCH semantic tokens for page CSS.
- Use a separate light/dark hex (or otherwise Mermaid-compatible) palette inside `themeVariables`.
- State explicitly that `oklch(...)`, `var(...)`, and `color-mix(...)` must not be passed to Mermaid theme variables unless Mermaid adds support.
- Add a render check that rejects Mermaid error SVGs (`aria-roledescription="error"` / `Syntax error in text`) before an explainer is delivered.

FNDEVVE:
Fixes #1043

## Root cause

The delegated `nicobailon/visual-explainer` Mermaid template uses compatible literal hex colors, but Plannotator's authoritative theme override replaced those values with `oklch(...)` strings.

OKLCH remains valid for normal page CSS in modern browsers. Mermaid is a separate color-processing boundary: with `theme: "base"`, it parses the supplied `themeVariables` and derives additional colors itself. Mermaid 11 rejects the OKLCH strings during that calculation.

## Minimal reproduction

The diagram source is unchanged between both runs:

```mermaid
flowchart TD
  A["Provider"] --> B["Database"]
```

With the six original OKLCH values, Mermaid 11 fails during theme initialization with `Unsupported color format`. Replacing only the Mermaid theme colors with the literal palette in this PR renders a normal `aria-roledescription="flowchart-v2"` SVG containing `Provider` and `Database`.

## What changed

- Replaced only the Mermaid-specific palette guidance with literal light and dark hex palettes that preserve Plannotator's purple/neutral identity.
- Explicitly prohibited OKLCH functions, CSS custom-property references such as `var()`, and `color-mix()` values inside Mermaid `themeVariables` while retaining OKLCH for ordinary page CSS.
- Added a reusable light/dark example that follows Plannotator's synchronized `color-scheme`.
- Made Mermaid 11 light/dark rendering a hard pre-delivery gate for explainers, rejecting exceptions, empty SVGs, `aria-roledescription="error"`, and `Syntax error in text`.
- Added a colocated Bun regression test that executes the exact documented example, captures both palettes, and uses the repository's existing Mermaid 11 + Happy DOM toolchain to render two representative flowcharts per palette. It verifies non-empty flowchart SVGs, labels, applied palette colors, and the absence of error signatures without adding a dependency.

The light and dark foreground/background pairs were browser-checked at contrast ratios from 6.70:1 to 12.09:1. Ordinary Plannotator CSS custom properties, including the existing OKLCH light and dark tokens, remain unchanged.

## Verification

- `bun install --frozen-lockfile` — passed; no lockfile change.
- `bun test apps/skills/extra/plannotator-visual-explainer/SKILL.test.ts` — 6 passed, including four real Mermaid 11.15.0 SVG renders.
- `bun test apps/skills` — 9 passed.
- `bun test scripts/install.test.ts` — 85 passed; includes skill/frontmatter and installer package contracts.
- `bun test` — 1,950 passed, 84 expected skips, 0 failed.
- `bun run typecheck` — passed.
- `bun run build:review && bun run build` — passed with no generated artifact drift.
- `git diff --check` — passed.
- `git merge-tree --write-tree origin/main HEAD` — passed against current `main` (`4f1ae8ee`).

### DIFF
--- a/apps/skills/extra/plannotator-visual-explainer/SKILL.md
+++ b/apps/skills/extra/plannotator-visual-explainer/SKILL.md
@@ -24,6 +24,12 @@ Three paths depending on content type. Each has its own references and structure
 
 Always deliver via Plannotator's annotation UI. Do NOT use `open` or `xdg-open`.
 
+For any deliverable that uses Mermaid, render every diagram with Mermaid 11 in both the light
+and dark palettes before opening the annotation UI. Rendering is a hard gate: an exception,
+empty SVG, or error output such as `aria-roledescription="error"` or `Syntax error in text`
+means the explainer is not deliverable. Fix the diagram or theme configuration and rerun both
+palettes until every SVG passes.
+
 **Plans/proposals** (user should approve/deny):
 ```bash
 plannotator annotate <file> --gate
--- a/apps/skills/extra/plannotator-visual-explainer/SKILL.test.ts
+++ b/apps/skills/extra/plannotator-visual-explainer/SKILL.test.ts
@@ -1,0 +1,45 @@
+/// <reference types="bun-types" />
+/// <reference types="node" />
+
+import { describe, expect, test } from "bun:test";
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+
+const themeOverride = readFileSync(
+  join(import.meta.dir, "references/theme-override.md"),
+  "utf-8",
+);
+const mermaidStart = themeOverride.indexOf("## Mermaid theming");
+const mermaidEnd = themeOverride.indexOf("\n## ", mermaidStart + 1);
+const mermaidSection = themeOverride.slice(mermaidStart, mermaidEnd);
+const exampleStart = mermaidSection.indexOf("```javascript");
+const exampleEnd = mermaidSection.indexOf("```", exampleStart + 3);
+const mermaidExample = mermaidSection.slice(exampleStart, exampleEnd);
+
+describe("plannotator-visual-explainer Mermaid theming", () => {
+  test("keeps the Mermaid theming section", () => {
+    expect(mermaidStart).toBeGreaterThan(-1);
+    expect(mermaidEnd).toBeGreaterThan(mermaidStart);
+    expect(exampleStart).toBeGreaterThan(-1);
+    expect(exampleEnd).toBeGreaterThan(exampleStart);
+  });
+
+  test("uses Mermaid-compatible literal colors", () => {
+    const literalColors = mermaidExample.match(/#[0-9a-f]{6}\b/gi) ?? [];
+    expect(mermaidExample).toContain("themeVariables");
+    expect(literalColors.length).toBeGreaterThanOrEqual(10);
+  });
+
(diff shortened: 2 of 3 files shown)


backnotprop:
thanks, in review

PR #1015, opened 2026-07-07
fix(review): show System in theme menu via shared mode list

gwynnnplaine:
Small QoL fix. I've sent a couple of little patches before (#523); while using code review I hit a papercut worth fixing.

### The papercut

The code-review **Options → Theme** menu only offers Light/Dark, but plan review, annotate, and Settings all offer **System** too — so "follow OS theme" silently wasn't available in review.

### Root cause

Each surface hand-listed its own theme modes (`PlanHeaderMenu`, `ReviewHeaderMenu`, `ThemeTab`, `ModeToggle`). The review picker had dropped `system`, and its `activeTheme` memo collapsed `system` → the resolved light/dark, so a System button couldn't have shown as active anyway.

### Changes

- Added one shared list, `packages/ui/components/themeModes.ts` (`THEME_MODES`), carrying each mode's id + label + icon.
- `Mode` is now derived from that list, so a mode can't exist without a render entry and no surface can omit one (no separate completeness guard needed).
- All four pickers map the shared list; active state = the selected mode everywhere. Dropped `ReviewHeaderMenu`'s `activeTheme` memo and its local Sun/Moon icons.

### Verification

typecheck clean · `apps/review` build + `build:hook` green · manual: `plannotator review` now shows Light/Dark/System (System selectable, follows the OS); plan/annotate/Settings/landing unchanged.

### Not in this PR

`ThemeProvider` still exposes redundant `theme`/`setTheme` and `mode`/`setMode` aliases — happy to collapse those separately if you'd like.

PR #1047, opened 2026-07-14
feat(review): add System option to the theme mode toggle

alexanderkreidich:
## Summary

The code review editor's Options menu only offered **Light/Dark**, while `ThemeProvider` already supports a `system` mode and the plan editor menu + Settings theme tab already expose it. This PR brings the review menu in line and fixes two issues surfaced along the way.

## Changes

- **`ReviewHeaderMenu.tsx`** — theme picker now maps `light / dark / system`, using the shared `themeIcons` (removed local duplicate Sun/Moon SVGs). Highlighting uses the stored mode directly instead of collapsing `system` to its resolved mode, so the System button highlights correctly.
- **`ActionMenu.tsx`** — widened the default dropdown panel `w-56` → `w-64`. The three-button segmented control overflowed the 224px panel (the "System" label clipped past the container edge). `PlanHeaderMenu` and `ReviewHeaderMenu` are the only `ActionMenu` consumers, and both render the three-button row.
- **`themeRegistry.ts` + both header menus** — added `getUnsupportedMode()` and disabled the incompatible mode button for dark-only/light-only color themes, instead of letting the toggle silently no-op. The highlight follows the mode actually rendered.

## Validation

- `bun run typecheck` — clean
- `bun test` — 1619 pass, 0 fail
- Built in documented order (`apps/review` build → `build:hook`) and manually verified in the browser: System option renders inside the container, follows macOS appearance changes live, and persists across reloads.

## UI verification

System theme toggle in the review editor Options menu

### DIFF
--- a/packages/review-editor/components/ReviewHeaderMenu.tsx
+++ b/packages/review-editor/components/ReviewHeaderMenu.tsx
@@ -7,6 +7,7 @@ import {
 } from '@plannotator/ui/components/ActionMenu';
 import { useTheme } from '@plannotator/ui/components/ThemeProvider';
 import { SunIcon, MoonIcon, SystemIcon } from '@plannotator/ui/components/icons/themeIcons';
+import { getUnsupportedMode } from '@plannotator/ui/utils/themeRegistry';
 import { MenuVersionSection } from '@plannotator/ui/components/MenuVersionSection';
 import { TextShimmer } from '@plannotator/ui/components/TextShimmer';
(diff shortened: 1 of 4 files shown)


backnotprop:
Closing as superseded by #1015, now merged. Its constrained-palette handling and menu-geometry work were incorporated into the shared-mode implementation, and @alexanderkreidich is credited as a co-author in the final merge commit.

Issue #642, opened 2026-05-01
Website annotation like in pi-annotate


JulianS-Uni:
Can you implement a website annotation similar to what [pi-annotate](https://github.com/nicobailon/pi-annotate) is doing? I like pi-annotate but having the chrome extension is clunky and the extension does not work when you are coding on a remote machine over ssh. Plannotator just works out of the box.

I thought about loading the website (url) to a new frame on the annotation overview. From there the annotation overview should recognice DOM-Elements and enable to comment on them. The report should include the DOM elements, id and class for the agent to recognize it.

backnotprop:
This is definitely a cool idea. Though, a lot of big websites will block DOM access through iframe.
- For your local apps, no problem, and can work today. If this is your main use case. It's easy
- Otherwise, to get around this, you either need like a Chrome extension or to launch a separate Chrome browser with debug mode enabled or have a desktop app.

Questions:
1. Is your main use case any website, or your local apps.

2. Is a desktop app a non-starter? I've been testing one of the works that keeps planning and review sessions alive long term.It would also enable this feature.

3. Also, random: I'm just curious. Can you let me know what your remote SSH setup is for Plannotator?

JulianS-Uni:
1. My use case would be my local apps. I wasn't aware of the problem you described and just thought it would be a nice extra to be able to use any website as reference or mockup.
2. I would be fine with a desktop app I think.
3. I am just using vscode with the [remote explorer extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.remote-explorer) which seems to tunnel the localhost automatically. Vscode is installed on my local machine and my code and plannotator alongside with pi is installed on the remote.
But I guess something like `ssh -L 3000:localhost:3000 user@192.168.1.100` would also work, although I haven't tried it for plannotator.

backnotprop:
Shipped in #1352, now on main and in the next release.

`plannotator annotate http://localhost:5173` proxies the running app through a loopback reverse proxy and opens it in the annotate UI: click any element to comment (placed markers, shift-click multi-select), Esc to interact with the app normally, and text selection commenting works in both modes. HMR and SPA navigation are supported.

Scope note: this covers local apps (loopback only) by design. Arbitrary remote URLs still use the existing static conversion path (Jina or fetch+Turndown). If remote-URL live annotation matters to you beyond that, open a fresh issue and we will scope it separately.

Thanks @JulianS-Uni for the original ask.

PR #1049, opened 2026-07-14
feat(annotate): annotate live localhost design previews

notxcain:
## Summary

Adds `plannotator annotate http://localhost:PORT/path` — renders a **live** frontend dev-server page (as an HTML artifact) for annotation, instead of converting it to markdown.

Partially addresses #642 (annotate a live website/URL in a frame, no extension, works over SSH) — scoped here to **localhost/loopback** dev previews; arbitrary public URLs still convert to markdown.

- **Per-session reverse proxy** forwards to the dev server, proxies its HMR websocket (no reload loop), and injects the annotation bridge into HTML responses. Loopback-only (`isLoopbackUrl` guard), bound to `127.0.0.1`. Bun impl (`packages/server/preview-proxy.ts`) + `node:http` mirror for Pi (`apps/pi-extension/server/previewProxy.ts`).
- Editor renders an `<iframe src>` at the proxy origin (additive optional `src` prop on `HtmlViewer`) so the real ES-module graph loads with **no null-origin CORS problem**; annotation works via the existing `postMessage` bridge. The `srcDoc` path is byte-for-byte unchanged when `src` is absent (`@plannotator/ui` seam rule).
- Threads `livePreviewUrl` through the annotate `/api/plan` (Bun, OpenCode, Pi) and CLI dispatch, carrying the target path onto the iframe URL.
- Live-preview sessions disable URL sharing, version history, and host-theme token injection.

### Why a proxy (not `<base>` / srcdoc)
A `srcDoc` iframe is a null origin; a modern Vite dev server returns no `Access-Control-Allow-Origin` to a null origin, so its module scripts are CORS-blocked → blank page. And its URLs are root-absolute (`/@vite/client`, `/src/main.tsx`), which a `<base>` can't redirect. A proxy that owns the origin root solves both.

### Caveat for reviewers — Pi runtime not fully verified
The Pi runtime's live-preview path (`apps/pi-extension/server/previewProxy.ts`, a `node:http` mirror of the Bun proxy) is **not end-to-end verified**. HTTP forwarding + bridge injection were smoke-tested against a real dev server, but the **HMR-websocket passthrough was never exercised against a live HMR client** — no Pi runtime was available this session. It is a dependency-free raw-socket implementation mirroring the Bun version. One known edge case: the WS passthrough hooks the socket `connect` event, which is correct for plain TCP but wrong for TLS — an https-loopback dev server with HMR would need `secureConnect` (plain-http loopback, the common case, is unaffected). The Claude Code path is verified end-to-end in a browser; OpenCode shares the same server/UI code but was only build-checked.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -254,6 +254,12 @@ Input type detected:
                deliberately excluded — it commonly holds secrets and annotate history
                copies file contents; source-code extensions stay with code review)
   .html/.htm → file read, rendered as raw HTML by default (or converted to markdown with --markdown)
+  http://localhost, 127.*, [::1]  → LIVE design preview: a per-session reverse proxy
+             (packages/server/preview-proxy.ts) forwards to the dev server, proxies its HMR
+             websocket, and injects the annotation bridge; the editor renders an <iframe src>
+             at the proxy origin so the real page (Vite ES-module graph and all) loads and is
+             annotated as HTML. Loopback-only. Non-goals: URL sharing, version history, and
+             host-theme token injection are disabled for live sessions.
   https://   → fetched via Jina Reader (default) or fetch+Turndown (--no-jina)
   folder/    → file browser opened, files converted on demand
         ↓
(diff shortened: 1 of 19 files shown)


backnotprop:
Sorry we missed this. Your PR sat unreviewed while we built the same idea from scratch on an internal spike branch, and that work landed today in #1352 without ever pulling yours in. That is on us, not on the quality of this PR: you saw the right feature early and had a working proxy approach for it.

We will credit you as a contributor on this feature in the release notes for the version that ships it. Closing as superseded by #1352, and we would genuinely welcome more from you.

notxcain:
@backnotprop awesome news! Thank you.

PR #1068, opened 2026-07-17
fix(review): Collapse viewed files in guided review

josdirksen:
I've started using the guided review a lot. Which works great with large PRs. What I was missing though, was the option to collapse the viewed files in the guided review. This is available in the normal review setup, but not in the guided one.

So this PR introduces the same collapse/viewed setup as is present in the normal view:

- Marking an unviewed file as Viewed collapses it.
- Unmarking it does not auto-expand it.
- A separate header control expands it.


backnotprop (approved):
Reviewed in depth — this is built the right way: it reuses the App-level viewed state (so GitHub sync, drafts, and tree badges stay coherent) and reproduces the all-files view's collapse semantics exactly (collapse on marking viewed, never auto-expand on unmark).

One gap: the guide's reveal channel (sidebar annotation clicks / AI line citations) expands the target *section* but the new per-file collapse unmounts the diff body, so a jump to a collapsed file scrolled to the header and then silently no-op'd. I pushed a small commit to your branch mirroring GuideSectionCard's token-keyed reveal effect at the file level, so a targeted file reopens before the scroll lands.

Thanks for the contribution — the collapse-on-viewed flow feels great in the guide.

PR #1076, opened 2026-07-19
fix(review): use canonical GPT-5.6 Sol model ID

### LINKED ISSUE #1070: GPT-5.6 Sol not supported in agents tab properly
This comment in the plannotator source code seems to be incorrect, it should be gpt-5.6-sol:
Image
(Source: packages/ui/components/AgentsTab.tsx)

Model picker in codex 0.144.4:
Image

Here are the errors that occur when a run is started with gpt-5.6 started.
Image

rNoz:
Use the Codex catalog's canonical `gpt-5.6-sol` identifier for GPT-5.6 Sol and migrate persisted `gpt-5.6` settings so existing users do not keep sending the stale model ID.

Fixes #1070

## Root cause

The review settings catalog exposed GPT-5.6 Sol with the model value `gpt-5.6`, while OpenAI Codex 0.144.4 publishes the model as `gpt-5.6-sol` in its authoritative `models.json`. AgentsTab feeds that catalog value into persisted Codex settings, so both new selections and previously saved review, tour, or guide selections could retain the stale identifier.

## Changes

- Change the AgentsTab model catalog entry to `gpt-5.6-sol` with the display label `GPT-5.6 Sol`.
- Add `migrateCodexSection` to migrate stale review, tour, and guide model selections.
- Migrate per-model reasoning-effort and fast-mode preferences from `gpt-5.6` to `gpt-5.6-sol`.
- Preserve an existing canonical preference when both stale and canonical keys are present, remove stale keys, and leave unrelated model settings unchanged.
- Add focused catalog and settings migration coverage.

This does not change provider defaults, reasoning options, fast-mode behavior, or settings for unrelated models.

## Validation

- `bun test packages/ui/components/AgentsTab.test.ts packages/ui/hooks/useAgentSettings.test.ts`: 13 passed, 0 failed, 16 assertions.
- `bun run typecheck` with TypeScript 5.9.3 available on `PATH`: passed.
- `bun test`: 2,132 passed, 99 skipped, 0 failed.
- `bun run --cwd apps/review build`: passed.
- `bun run build:hook`: passed after the review build.

## Manual testing gap

A live launch against Codex 0.144.4 was not run because the locally installed Codex CLI was older during validation. The canonical identifier was verified against Codex 0.144.4 `models.json`, and the focused tests cover catalog selection plus persisted migration behavior.

Please, let me know if you want that I do some further E2E test/proof. Otherwise, ready.

### DIFF
--- a/packages/ui/components/AgentsTab.test.ts
+++ b/packages/ui/components/AgentsTab.test.ts
@@ -1,0 +1,10 @@
+import { expect, test } from "bun:test";
+import { CODEX_MODELS } from "./AgentsTab";
+
+test("uses the canonical GPT-5.6 Sol model ID", () => {
+  expect(CODEX_MODELS).toContainEqual({
+    value: "gpt-5.6-sol",
+    label: "GPT-5.6 Sol",
+  });
+  expect(CODEX_MODELS.some(({ value }) => value === "gpt-5.6")).toBe(false);
+});
--- a/packages/ui/components/AgentsTab.tsx
+++ b/packages/ui/components/AgentsTab.tsx
@@ -50,10 +50,9 @@ export const CLAUDE_EFFORT: Array<{ value: string; label: string }> = [
 ];
 
 export const CODEX_MODELS: Array<{ value: string; label: string }> = [
-  // GPT-5.6 naming scheme: the bare `gpt-5.6` alias routes to `gpt-5.6-sol`
-  // (flagship); `-terra` is the mid price/performance tier and `-luna` the
-  // efficient high-volume tier.
-  { value: 'gpt-5.6', label: 'GPT-5.6' },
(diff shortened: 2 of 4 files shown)


rNoz:
Author of reported bug https://github.com/backnotprop/plannotator/issues/1070#issuecomment-5017110060 has confirmed this PR fixes the issue.

backnotprop (approved):
Verified against #1070: Codex 0.144.4's models.json publishes gpt-5.6-sol and rejects the bare gpt-5.6 id. The persisted-settings migration (review/tour/guide sections plus per-model preferences, canonical key winning over stale) is exactly right — without it existing users would keep sending the broken id. Thanks for the thorough fix and tests.

PR #1078, opened 2026-07-19
fix(install): pass --global to npx skills add so extras do not leak into cwd

### LINKED ISSUE #1077: Installer extra skills can be written into the current project
## Summary

Plannotator's macOS/Linux, PowerShell, and Command Prompt installers invoke `npx skills add backnotprop/plannotator/apps/skills/extra` without an explicit scope. When an installer or its printed install-later command runs from inside a Git repository, the Skills CLI can select project scope and write Plannotator's extra skills into that repository's `.agents/skills/` directory instead of the user's global skills directory.

## Affected surfaces

- `scripts/install.sh`
- `scripts/install.ps1`
- `scripts/install.cmd`
- The install-later commands printed by those installers
- Public installation examples that repeat the same unscoped command

## Reproduction

1. Enter a Git repository that does not intentionally vendor Plannotator skills.
2. Run `npx --yes skills@1.5.14 add backnotprop/plannotator/apps/skills/extra --help` from that directory.
3. Inspect the repository's `.agents/skills/plannotator-*` directories and `skills-lock.json`.

## Observed behavior

Despite the trailing `--help`, Skills CLI 1.5.14 did not show add-command help. With `--yes`, it auto-detected GitHub Copilot and installed all three extra skills under the current repository's `.agents/skills/` directory. It also created a project-local `skills-lock.json`. This created unrelated project files and left a dirty working tree.

## Expected behavior

Plannotator's user-level installer should install its optional extra skills in the user's global skills location regardless of the directory from which installation starts.

## Root cause

The Skills CLI distinguishes project and user-global installation. Its `--global` option selects the user-level directory, while an unscoped invocation may select project scope from the current working directory. Plannotator's three installers and public examples omit that option, so their result depends on where the user runs them.

## Security and hygiene impact

This is primarily a repository hygiene and trust-boundary problem. A user running a global application installer can unexpectedly modify an unrelated source repository, add files that may later be committed by mistake, and make installation behavior vary by current directory.

## Proposed fix

Add `--global` to every supported installer invocation and printed example for `npx skills add backnotprop/plannotator/apps/skills/extra`. Add a deterministic parity check that enumerates every occurrence across the three installers and requires all occurrences to use explicit global scope.

## Prior art and overlap

No issue or pull request was found that already fixes this cross-platform scope leak. Merged PR #850 introduced an unscoped extra-skills command. Issues #853 and #998 concern adjacent skill installation behavior but do not cover current-directory project writes or explicit global scope.

## Environment

The live reproduction was tested with Skills CLI 1.5.14. The four resulting project-local artifacts, three skill directories and `skills-lock.json`, were removed after the observation.

Skills CLI 1.5.19 was not run live for this proof. Its tagged source retains the same contract: the main add-command case parses options and calls the add operation without intercepting `-h` or `--help`, while the add option parser does not recognize help and still supports `--global` for user-level scope. The behavior applies when installation starts inside a Git repository and the Skills CLI resolves an unscoped install as project-local. The affected Plannotator surfaces are platform-independent because the shell, PowerShell, and Command Prompt installers all emit the same unscoped operation.

## Contribution

I am willing to contribute the cross-platform installer, documentation, and regression-test update. I worked in a fix a couple of weeks ago and it worked great. However, I have spent the morning polishing it and considering other systems/platforms for a more robust fix. I'll open a PR soon => Done!

rNoz:
Plannotator's optional extra-skills installation was unscoped. When an installer or a copied install-later command ran inside a Git repository, the Skills CLI could use project scope and write Plannotator skills into that repository's `.agents/skills/` directory.

This change makes the intended user-level behavior explicit by adding `--global` everywhere Plannotator installs or documents its extra skills.

Closes https://github.com/backnotprop/plannotator/issues/1077

## Why

A global application installer should not modify an unrelated repository based on the caller's current directory. Explicit scope makes the result consistent across supported platforms and avoids dirtying projects with accidental skill files.

## Changes

- Add `--global` to the extra-skills commands in `scripts/install.sh`, `scripts/install.ps1`, and `scripts/install.cmd`.
- Update `AGENTS.md` and the installation and Claude Code marketing guides so copied commands match installer behavior.
- Update the existing shell assertion and add a deterministic parity test that inspects every extra-skills command across the installers and checked documentation, requires at least one occurrence in every checked file, and rejects any occurrence without `--global`.

The installers retain their existing prompts, quoting, and platform-specific control flow.

## Validation

- `bun test scripts/install.test.ts`: 86 passed, 0 failed, 501 assertions.
- `bash -n scripts/install.sh`: passed.
- `shellcheck -S error scripts/install.sh`: passed.
- `bun run build:marketing`: passed, 40 pages built.
- `bun run typecheck` with TypeScript 5.9.3 available on `PATH`: passed.
- `bun test`: 2,129 passed, 99 skipped, 0 failed.
- Pre-fix live reproduction with Skills CLI 1.5.14: the unscoped command created three project-local skill directories and `skills-lock.json`; the generated artifacts were removed.
- Post-fix isolated-home E2E with Skills CLI 1.5.19: the emitted `--global` command installed exactly three skills under the temporary home, left the temporary Git project clean, created no project lockfile, and left the user's real `~/.agents` state unchanged. The temporary home was removed afterward.

The deterministic regression test establishes source parity for every checked command, while the isolated-home E2E proves the external CLI's global installation behavior without changing the user's real global state.

## Manual testing gaps

No live PowerShell or Command Prompt installer run was performed on Windows (I don't have that system).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -48,7 +48,7 @@ plannotator/
 │       │   ├── plannotator-review/    # Lightweight: opens review UI
 │       │   ├── plannotator-annotate/  # Lightweight: opens annotate UI
 │       │   └── plannotator-last/      # Lightweight: annotates last message
-│       └── extra/                 # EXTRA skills — NOT default-installed (except Kiro); add via `npx skills add backnotprop/plannotator/apps/skills/extra`
+│       └── extra/                 # EXTRA skills — NOT default-installed (except Kiro); add via `npx skills add backnotprop/plannotator/apps/skills/extra --global`
 │           ├── plannotator-compound/        # Research analysis agent (map-reduce over denied plans)
 │           ├── plannotator-setup-goal/      # Goal package scaffolder for /goal workflows
 │           └── plannotator-visual-explainer/ # Visual HTML generator (plans, diagrams, PR explainers) with Plannotator theming
--- a/apps/marketing/src/content/docs/getting-started/installation.md
+++ b/apps/marketing/src/content/docs/getting-started/installation.md
@@ -138,7 +138,7 @@ Upgrading from an older version? The installer removes the legacy `~/.claude/com
 Optional extra skills (compound planning, setup-goal, visual explainer) are not installed by default. Add them with:
 
 ```bash
-npx skills add backnotprop/plannotator/apps/skills/extra
+npx skills add backnotprop/plannotator/apps/skills/extra --global
 ```
 
 ## OpenCode
@@ -243,7 +243,7 @@ Notes:
 The installer also copies Plannotator's core skills (`plannotator-review`, `plannotator-annotate`, `plannotator-last`) into `~/.agents/skills` — the official OpenAI agent skills path. Optional extra skills (compound planning, setup-goal, visual explainer) are not installed by default; add them with:
 
 ```bash
-npx skills add backnotprop/plannotator/apps/skills/extra
+npx skills add backnotprop/plannotator/apps/skills/extra --global
 ```
 
 You can still use the direct commands at any time:
--- a/apps/marketing/src/content/docs/guides/claude-code.md
+++ b/apps/marketing/src/content/docs/guides/claude-code.md
@@ -92,7 +92,7 @@ Annotates the agent's most recent message. See the [annotate last docs](/docs/co
 Optional extra skills (compound planning, setup-goal, visual explainer) are not installed by default. Add them with:
 
 ```bash
-npx skills add backnotprop/plannotator/apps/skills/extra
+npx skills add backnotprop/plannotator/apps/skills/extra --global
 ```
 
 ## Plugin installation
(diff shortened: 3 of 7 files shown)


backnotprop (approved):
Thanks @rNoz, this is a model external PR. Verified the premise against skills CLI 1.5.19 (the add command's --yes scope note is literally "project if in a project, else global", so the leak is real), confirmed every extras command site in the repo is covered (the Kiro path uses direct file copies and is correctly out of scope, and CLAUDE.md is covered via the AGENTS.md symlink), and mutation-tested the new parity test: removing a single --global from any checked file fails with a precise per-file message. As a bonus, this makes the installer's own extras_present re-run detection self-consistent, since that check only ever looked at user-level skill dirs. All installers soft-fail the extras step, so there was no abort risk either way. One follow-up on our side: the two marketing docs pages edited here no longer ship, so the same fix needs to land in the external docs repo. Merging.

PR #1095, opened 2026-07-20
fix: Cursor review engine sandbox can be disabled for systems where it cannot start (NixOS/AppArmor)

### LINKED ISSUE #1094: Guided Review forces Cursor sandbox on Linux, cannot create one + ignores user config
## Summary

Guided Review with the Cursor engine fails on Linux systems where Cursor Agent sandboxing is unavailable.

Plannotator launches Cursor Agent with `--sandbox enabled`, which overrides the user's Cursor Agent configuration—even when sandboxing has been explicitly disabled with:

```bash
agent sandbox disable
```

Environment
OS: NixOS / Linux
Plannotator: v0.23.1
Cursor Agent: 2026.07.09-a3815c0
Cursor configuration:
```json
{
  "approvalMode": "allowlist",
  "sandbox": {
    "mode": "disabled"
  }
}
```
### Steps to reproduce

1. Configure Cursor Agent to use allowlist mode: `agent sandbox disable`
2. Open a Plannotator code review.
3. Select Guide / Start a guided review.
4. Choose Cursor as the engine.
5. Generate the guide.

### Actual behavior
The guided review fails with:
```
Error: Sandbox mode is enabled but not available on this system.
Sandbox failed to start, possibly due to AppArmor configuration.
Visit https://cursor.com/docs/agent/terminal for troubleshooting steps.
Run 'agent sandbox disable' to switch to allowlist mode.
```

The error persists after running agent sandbox disable because Plannotator invokes Cursor with the sandbox explicitly enabled.
In v0.23.1, packages/server/marker-review.ts builds the Cursor command with:
"--sandbox", "enabled".

I fixed by doing an ugly workaround in my nix setup that removes the `--sandbox`:
https://github.com/Vincent-HD/.nixfiles/commit/ff73defa6f06dfdf8b342dc1abe6fe712b5cd929#diff-acbe854aae5415b9efb96e917df336b3f684c4b1f61812adee710e889d116fa6

backnotprop:
## Problem

The Cursor review engine (`cursorBuildArgv` in `packages/server/marker-review.ts`) hardcodes `--sandbox enabled` when launching Cursor's `agent` CLI. On systems where Cursor's sandbox cannot start — NixOS, AppArmor-restricted Linux — every Guided Review / review job hard-fails:

> Sandbox mode is enabled but not available on this system

The hardcoded flag also overrides the user's own `agent sandbox disable` configuration, so there was no way to opt out.

## Fix

**Default unchanged**: review jobs still pass `--sandbox enabled` as part of their read-only posture (`--mode ask` + `--sandbox enabled`, no `--force`/`--yolo`).

New escape hatch, following the repo's env-plus-config pattern (`resolveCursorSandbox()` in `packages/shared/config.ts`, next to `resolveAnnotateHistory`/`resolveUseJina`): when the user opts out, the `--sandbox enabled` pair is **omitted entirely** — never flipped to `--sandbox disabled` — so the user's own Cursor Agent configuration governs (e.g. the reporter's allowlist mode).

Wired at every `buildMarkerCommand` call site in **both runtimes**: Bun (`packages/server/review.ts`, `packages/server/guide/guide-review.ts` — normal and repair launches) and Pi (`apps/pi-extension/server/serverReview.ts`; `guide-review.ts` is vendored, with the new `../config` import rewrite added to `vendor.sh`). `marker-review.ts` itself stays env-free — the resolved value flows in via `MarkerBuildOptions.cursorSandbox`, same style as the existing `thinking` knob.

## Usage

Env var (takes precedence):

```bash
export PLANNOTATOR_CURSOR_SANDBOX=0   # or false / disabled
```

Or persistently in `~/.plannotator/config.json`:

```json
{ "cursorSandbox": false }
```

`1` / `true` / `enabled` (and unset) keep today's behavior.

## Tradeoff

Stated honestly: with the sandbox flag omitted, the review job's write protection relies on `--mode ask` plus whatever sandboxing the user's own Cursor configuration provides. That is why the default stays on and opting out is explicit. The doc comment above `cursorBuildArgv`, `AGENTS.md`, and the marketing environment-variables reference all document the escape hatch and this tradeoff.

## Tests

- Kept the existing default-behavior assertion (`--sandbox enabled` present).
- New: `cursorSandbox: false` omits the pair (and never emits `disabled`); explicit `true`/`{}` keep it.
- New `packages/shared/config.test.ts` covering `resolveCursorSandbox`: default, config key, env off-values (`0`/`false`/`disabled`, case-insensitive), env-wins-over-config both directions.
- `bun test packages/server packages/shared`: 1017 pass. Pi extension server tests: 32 pass. `bun run typecheck` (includes vendored Pi build): clean.

Closes #1094

Credit to @Vincent-HD for the report and for confirming the workaround diff (removing the flag) on NixOS.

https://claude.ai/code/session_01YXkgsNucxDwAL4GdR4XYRk


Vincent-HD:
Thank you so much for the quick response, it's much appreciated 🙏

PR #1098, opened 2026-07-20
fix(pi): keep PR review progress out of composer

alexanderkreidich:
## Summary

- render PR fetch and checkout progress through Pi's managed footer status
- clear temporary review progress after startup succeeds or fails
- surface local-checkout fallback warnings through Pi notifications instead of raw terminal output

## Why

`/plannotator-review <PR URL>` wrote startup progress with `console.error()`. Those writes bypass Pi's TUI renderer and can land inside the active composer, corrupting its layout.

### Before — progress output collides with the composer

PR review progress written inside the Pi composer before the fix

### After — progress renders in Pi's managed footer

PR review progress rendered below the Pi composer after the fix

## Validation

- `bun run build:pi`
- `bun test apps/pi-extension` — 71 passed, 1 skipped
- `bun run --cwd apps/pi-extension tsc --noEmit`
- real PR startup through a PTY using `2bb-dev/openorange#606`; every progress message was present in `PI_TUI_WRITE_LOG`, and no legacy raw output reached the composer

### DIFF
--- a/apps/pi-extension/plannotator-browser.ts
+++ b/apps/pi-extension/plannotator-browser.ts
@@ -67,6 +67,32 @@ export interface BrowserDecisionSession<T> {
 	stop: () => void;
 }
 
+type CodeReviewOptions = {
+	cwd?: string;
+	defaultBranch?: string;
+	diffType?: DiffType;
+	prUrl?: string;
+	vcsType?: VcsSelection;
+	useLocal?: boolean;
+};
+
+type CodeReviewDecision = {
+	approved: boolean;
+	feedback?: string;
+	annotations?: unknown[];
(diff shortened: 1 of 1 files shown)


backnotprop (approved):
Thanks @alexanderkreidich, this is a well-aimed fix. Routing startup progress through ctx.ui.setStatus is exactly Pi's intended channel for ambient extension status (it renders in the managed footer and never touches the transcript or the input editor), and moving the checkout warnings to ctx.ui.notify gives them a durable trace in chat. The try/finally clear means no stale footer text either way. Verified against the Pi runtime source; full Pi test suite and typecheck pass. Appreciate the before/after screenshots and the PTY validation.

PR #1099, opened 2026-07-20
feat: annotate accepts YAML, JSON, TOML and other plain-text files

### LINKED ISSUE #1029: Annotate command rejects .yaml and other text-based files
## Description

The `/annotate` command rejects `.yaml`, `.yml`, `.json`, `.toml`, and many other plain-text file formats with the error:

> Only .md, .mdx, .txt, .html, .htm files are supported.

These are all plain text files that the annotation engine can handle just fine — the restriction is purely an extension whitelist in the code, not a technical limitation.

### Root cause

Two places in `index.ts` restrict which files can be annotated:

**1. File type guard (line ~587):**
```ts
} else {
    if (!/\.(mdx?|txt)$/i.test(absolutePath)) {
        ctx.ui.notify("Only .md, .mdx, .txt, .html, .htm files are supported.", "error");
        return;
    }
    markdown = readFileSync(absolutePath, "utf-8");
}
```
This is the catch-all `else` block — after checking for folders and HTML files, it rejects anything that is not `.md`, `.mdx`, or `.txt`.

**2. Folder scanner (line ~566):**
```ts
if (!hasMarkdownFiles(absolutePath, FILE_BROWSER_EXCLUDED, /\.(mdx?|txt|html?)$/i)) {
    ctx.ui.notify(`No markdown, text, or HTML files found in ${absolutePath}`, "error");
```
Folder annotation also limits discovery to the same narrow set.

### Affected file types

Text-based files that are **unnecessarily rejected**:
- `.yaml`, `.yml` — YAML configs, CI/CD, OpenAPI specs
- `.json`, `.jsonc`, `.json5` — JSON configs, package.json, tsconfig
- `.toml` — pyproject.toml, Cargo.toml, ruff.toml
- `.xml`, `.svg` — XML configs, Maven POMs
- `.ini`, `.cfg`, `.conf` — INI-style configs
- `.env`, `.env.example` — environment files
- `.css`, `.scss`, `.less` — stylesheets
- `.js`, `.jsx`, `.ts`, `.tsx` — source code (comments as annotations)
- `.py`, `.rs`, `.go`, `.java`, `.rb`, `.php`, `.sh`, `.bash` — any source code
- `Dockerfile`, `Makefile`, `.gitignore`, etc.

### Why it matters

The annotation system uses `startMarkdownAnnotationSession` with plain text content read via `readFileSync(absolutePath, "utf-8")`. It works with any UTF-8 text — the annotator renders text and lets the user place annotations on it. There is no markdown-specific processing done on the content; it is sent as-is to the annotation UI.

### Suggested fix

**Option A (conservative):** Expand the regex to include common text-based config and code formats:
```ts
/\.(mdx?|txt|ya?ml|jsonc?|json5|toml|xml|cfg|ini|conf|env|css|s[ac]ss|less|[jt]sx?|py|rs|go|java|rb|php|sh|bash|zsh)$/i
```

**Option B (permissive):** Allow any file that is readable as UTF-8 text and not explicitly excluded (e.g., not binary). This would handle `Dockerfile`, `Makefile`, `.gitignore`, and any future text format without manual whitelist updates:
```ts
const isTextFile = (path: string): boolean => {
    const buf = readFileSync(path);
    // Skip binary files (null bytes in first 8KB)
    return !buf.subarray(0, Math.min(buf.length, 8192)).includes(0);
};
```

### Steps to reproduce

1. Open a Pi session with plannotator extension active
2. Try `/annotate some-file.yaml`
3. See error: "Only .md, .mdx, .txt, .html, .htm files are supported."

## Diagnostics

- **plannotator version:** 0.22.0
- **File location:** `index.ts` lines ~566 and ~587
- **Package:** `@plannotator/pi-extension`
- **Repository:** `backnotprop/plannotator`, path `apps/pi-extension`

backnotprop:
## What

`/plannotator-annotate` (and `plannotator annotate`) now accepts common plain-text config and data formats instead of rejecting everything that isn't `.md`/`.mdx`/`.txt`/`.html`/`.htm`. The content pipeline already reads files as UTF-8 text, so the restriction was purely an extension whitelist — as reported by @Serhioromano in #1029.

**New accepted extensions:** `.yaml` `.yml` `.json` `.jsonc` `.json5` `.toml` `.ini` `.cfg` `.conf` `.properties` `.csv` `.tsv` `.log` `.xml` `.env.example`

These render exactly the way `.txt` renders today — plain text through the markdown pipeline, selectable and annotatable. Syntax-highlighted / code-fence rendering for structured formats is a possible follow-up, deliberately out of scope here.

**Deliberate exclusions:**
- `.env` — it commonly holds secrets, and annotate's per-file version history copies file contents into `~/.plannotator/history/`. The secret-free `.env.example` convention is accepted instead.
- Source-code extensions (`.ts`, `.py`, …) — they belong to the code-file link/popout system and code review, and would flood the annotate folder file browser.

## How

New single source of truth: `packages/core/annotatable.ts` exporting `ANNOTATABLE_TEXT_REGEX` (plain-text set) and `ANNOTATABLE_DOC_REGEX` (plain-text + HTML), re-exported through `@plannotator/shared/resolve-file` and vendored into the Pi extension via `vendor.sh`. Per-call-site decisions:

| Site | Decision |
|------|----------|
| `resolveMarkdownFile` accept + fuzzy walk (`packages/shared/resolve-file.ts`) | Widened to `ANNOTATABLE_TEXT_REGEX` — single-file annotate and bare-filename in-root search now find config files. HTML keeps its own branch at call sites. |
| Folder discovery (hook CLI, OpenCode, Pi) | Widened to `ANNOTATABLE_DOC_REGEX`; error messages updated. |
| File-browser listing (Bun `reference-handlers.ts`, Pi `reference.ts`) | Widened to `ANNOTATABLE_DOC_REGEX` in both runtimes. |
| `/api/doc` serve guards (both runtimes) | New `doc=1` query param, set only by the file browser. Extensions that overlap `CODE_FILE_REGEX` (`.yaml`/`.json`/`.toml`/`.ini`/`.xml`) render as annotatable documents **only** with `doc=1`; without it they keep the `codeFile: true` popout response, so code-file links inside plans/docs are unaffected. Non-overlapping new extensions (`.csv`, `.log`, …) serve as documents unconditionally. |
| Linked-doc link regexes (`InlineMarkdown`, `HtmlBlock`) | **Unchanged** — `config.yaml` links inside a document keep the syntax-highlighted code popout, which is the better rendering for a reference link. |
| File-browser display names | Extension stripping stays narrow (`.md`/`.txt`/`.html`) so `config.yaml` and `config.json` siblings stay distinguishable. |
| Mid-edit file-open guard (`packages/editor/App.tsx`) | Gates on `isSourceSaveFilePath` (`.md`/`.mdx`/`.txt`) — config files are view-only (no source-save capability), so they cannot be opened mid-edit; this avoids silently downgrading "Done editing" to feedback-only edits. |
| Obsidian vault walkers | **Unchanged** — vaults stay markdown-only. |

**Frontmatter rule (from review):** stripping a leading `--- … ---` pair is a markdown convention; for non-markdown plain-text sources those delimiters are real content (a k8s multi-document YAML starts with them). `parseMarkdownToBlocks` gained a `{ frontmatter }` option and the editor keys it off the active document's path via `shouldStripFrontmatter()`: strip for `.md`/`.mdx` and pathless/converted sources (plans, agent messages, URLs, `--markdown` HTML); keep raw for every other annotatable text type (including `.txt`, whose leading `---` pair previously vanished too).

**Size cap (from review):** a shared `MAX_ANNOTATABLE_FILE_BYTES` (2MB — the same limit the code-file popout has always enforced, now single-sourced) guards the annotate CLI single-file read in all three runtimes and the `/api/doc` document branches in both servers. A multi-GB `server.log` now gets a clear "File too large to annotate (max 2MB)" error instead of OOMing the server and being copied into annotate history. Note: this also caps `.md`/`.txt` document serves, previously unbounded — a behavior change only for pathological inputs.

Annotate history slug (`annotate-{basename}-{hash8}`) is extension-agnostic and works unchanged for the new types; Bun and Pi servers are updated in lockstep. Agent-facing skill docs (`plannotator-annotate` core + Kiro) now mention the config formats.

## Tests

- `packages/shared/resolve-file.test.ts`: accept cases for every new extension, bare-filename in-root resolution, `.py` rejection, `.env` rejected / `.env.example` accepted, predicate classification.
- `packages/server/reference-handlers.test.ts`: file-browser listing includes config formats and excludes `.env`/source code; `doc=1` serves `.yaml` as markdown; without `doc=1` the code-popout response is preserved; `.csv` serves as markdown unconditionally; 413 size-cap coverage for the `doc=1`, markdown-fallback, and base-relative branches.
- `packages/ui/utils/parser.test.ts`: YAML/JSON content parses without loss; multi-document YAML keeps its **first** document with `frontmatter: false`; default parse still strips markdown frontmatter; `shouldStripFrontmatter` path-rule coverage.

All suites green: `bun test packages/shared packages/server packages/ui packages/editor packages/core apps/opencode-plugin apps/pi-extension` (1,776 pass, 0 fail), `bun run typecheck` (includes Pi vendor + strict-consumer), review+hook bundles build, and CLI smoke tests (`annotate x.py` → helpful error; `annotate smoke.yaml` → resolves and opens; 3MB `big.yaml` → size-cap error).

Closes #1029 — thanks @Serhioromano for the detailed report.

https://claude.ai/code/session_01YXkgsNucxDwAL4GdR4XYRk

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -248,7 +248,11 @@ Claude Code: plannotator annotate subcommand runs
 OpenCode/Pi: event handler intercepts command
         ↓
 Input type detected:
-  .md/.mdx   → file read from disk
+  .md/.mdx/.txt → file read from disk
+  plain-text config/data formats (.yaml .yml .json .jsonc .json5 .toml .ini .cfg .conf .properties .csv .tsv .log .xml .env.example)
+             → read from disk, rendered as plain text exactly like .txt (.env itself is
+               deliberately excluded — it commonly holds secrets and annotate history
+               copies file contents; source-code extensions stay with code review)
   .html/.htm → file read, rendered as raw HTML by default (or converted to markdown with --markdown)
   https://   → fetched via Jina Reader (default) or fetch+Turndown (--no-jina)
   folder/    → file browser opened, files converted on demand
@@ -412,7 +416,7 @@ Every plan is automatically saved to `~/.plannotator/history/{project}/{slug}/` 
 
 This powers the version history API (`/api/plan/version`, `/api/plan/versions`) and the plan diff system.
 
-**Annotate mode** also saves history on open, so the same version diff works when annotating a standalone `.md`/`.txt`/`.html` file. It keys the slug by **file path** — `annotate-{sanitized-basename}-{hash8}` — rather than heading + date, so re-opening the same file groups its versions even as its content (and headings) change. **Note this writes a copy of each annotated file's content** under `~/.plannotator/history/` (or `PLANNOTATOR_DATA_DIR`); disable via `PLANNOTATOR_ANNOTATE_HISTORY=0` or `{ "annotateHistory": false }` in `~/.plannotator/config.json` to keep annotate sessions stateless (the version diff is then unavailable). For `--render-html` files the diff is rendered as the real page with inline `<ins>`/`<del>` highlights via `htmlDiff()` (`packages/shared/html-diff.ts`).
+**Annotate mode** also saves history on open, so the same version diff works when annotating a standalone `.md`/`.txt`/`.html` file (or any other supported plain-text file, e.g. `.yaml`/`.json`/`.toml`). It keys the slug by **file path** — `annotate-{sanitized-basename}-{hash8}` — rather than heading + date, so re-opening the same file groups its versions even as its content (and headings) change. **Note this writes a copy of each annotated file's content** under `~/.plannotator/history/` (or `PLANNOTATOR_DATA_DIR`); disable via `PLANNOTATOR_ANNOTATE_HISTORY=0` or `{ "annotateHistory": false }` in `~/.plannotator/config.json` to keep annotate sessions stateless (the version diff is then unavailable). For `--render-html` files the diff is rendered as the real page with inline `<ins>`/`<del>` highlights via `htmlDiff()` (`packages/shared/html-diff.ts`).
 
 History saves independently of the `planSave` user setting (which controls decision snapshots in `~/.plannotator/plans/`). Storage functions live in `packages/shared/storage.ts` (runtime-agnostic, re-exported by `packages/server/storage.ts`). Pi copies the shared files at build time. Slug format: `{sanitized-heading}-YYYY-MM-DD` (heading first for readability).
 
--- a/apps/hook/server/index.ts
+++ b/apps/hook/server/index.ts
@@ -96,7 +96,7 @@ import { urlToMarkdown, isConvertedSource } from "@plannotator/shared/url-to-mar
 import { createWorktreePool, type WorktreePool, type PoolEntry } from "@plannotator/shared/worktree-pool";
 import { parsePRUrl, checkPRAuth, fetchPR, getCliName, getCliInstallUrl, getMRLabel, getMRNumberLabel, getDisplayRepo } from "@plannotator/server/pr";
 import { writeRemoteShareLink } from "@plannotator/server/share-url";
-import { resolveMarkdownFile, resolveUserPath, hasMarkdownFiles } from "@plannotator/shared/resolve-file";
+import { resolveMarkdownFile, resolveUserPath, hasMarkdownFiles, ANNOTATABLE_DOC_REGEX, ANNOTATABLE_EXTENSIONS_HINT } from "@plannotator/shared/resolve-file";
 import { FILE_BROWSER_EXCLUDED } from "@plannotator/shared/reference-common";
 import { statSync, rmSync, realpathSync, existsSync } from "fs";
 import { parseRemoteUrl } from "@plannotator/shared/repo";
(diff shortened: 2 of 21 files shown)


backnotprop:
Adversarially reviewed, then re-verified after fixes. The doc=1 mechanism keeps code-file links on the syntax-highlighted popout while file-browser selections render as annotatable documents, in both runtimes, with path containment unchanged. Review findings all addressed: multi-document YAML no longer loses its first document to frontmatter stripping (now keyed off file type, and .txt got the same fix), a shared 2MB cap now guards every doc read path (previously only the code popout was capped; a multi-GB log can no longer OOM the server or flood annotate history), and config files are view-only rather than silently downgrading mid-edit saves to feedback-only. .env stays excluded because annotate history copies file contents. 1,776 tests pass across all runtimes. Thanks @Serhioromano for the report.

PR #1101, opened 2026-07-21
fix(ui): parse YAML block scalars in frontmatter

ruaridhw:
## Problem

`extractFrontmatter` (`packages/ui/utils/parser.ts`) is a line-based `key: value` parser with no support for YAML **block scalars**. When a value is a folded (`>`) or literal (`|`) scalar, the parser stores the bare indicator as the value and silently drops the indented body.

A document with:

```yaml
---
name: example
description: >-
  one two
  three four
---
```

parses to `description === ">-"`, so the description renders empty (just `>-`) in the plan/annotate viewer. This shows up in the wild because agent skill `SKILL.md` files routinely fold their long `description` field with `>-`.

## Fix

When the value after the colon matches a block-scalar indicator (`|`/`>` with optional `+`/`-` chomping), consume the following lines indented deeper than the key as the value:

- `>` (folded) — adjacent non-empty lines join with a space; a run of N blank lines folds to N newlines
- `|` (literal) — newlines preserved
- common leading indentation stripped; trailing blank lines dropped

Extracted three small private helpers (`indentWidth`, `dedentLines`, `foldScalarLines`) plus `parseBlockScalar` to keep `extractFrontmatter` readable. Plain single-line values and `-` arrays are unchanged.

Obviously this isn't a comprehensive YAML extractor though so this just patches over this (one) issue (for now)

## Tests

Added an `extractFrontmatter — block scalars` suite covering folded `>-`, folded `>` with a paragraph break, literal `|`, "doesn't swallow the next key", and a plain-value/array regression.

- `bun test packages/ui/utils/parser.test.ts` → 126 pass, 0 fail
- `tsc --noEmit -p packages/ui/tsconfig.json` → clean

No `node:` imports added (keeps `@plannotator/core` browser-safe). UI bundle rebuild left to the maintainer.

### DIFF
--- a/packages/ui/utils/parser.test.ts
+++ b/packages/ui/utils/parser.test.ts
@@ -1078,6 +1078,12 @@ name: bar
     expect(frontmatter?.name).toBe("bar");
   });
 
+  test("CRLF line endings do not leak \\r into the value", () => {
+    const md = "---\r\ndescription: >-\r\n  one two\r\n  three\r\n---\r\n# Hi";
+    const { frontmatter } = extractFrontmatter(md);
+    expect(frontmatter?.description).toBe("one two three");
+  });
+
   test("plain single-line values and arrays still parse", () => {
     const md = `---
 name: foo
--- a/packages/ui/utils/parser.test.ts
+++ b/packages/ui/utils/parser.test.ts
@@ -1029,6 +1029,69 @@ describe("extractFrontmatter — contentStartLine", () => {
   });
 });
 
+describe("extractFrontmatter — block scalars", () => {
+  test("folded (>-) joins wrapped lines with spaces", () => {
+    const md = `---
+description: >-
+  one two
+  three four
+---
+# Hi`;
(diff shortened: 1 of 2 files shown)


backnotprop:
Reviewed this today. Nice work, it held up against everything we threw at it: block scalars at the end of frontmatter, body lines containing colons or leading dashes, empty scalars, arrays following a scalar, the unsupported explicit-indent form falling back to old behavior instead of misparsing, and the realistic SKILL.md shape that motivated the fix.

One gap: CRLF sources. The body lines were collected raw, so a file with Windows line endings left a literal carriage return embedded mid-value ("one two\r three"). Every other path in this parser trims lines before use, so this would have been the first place a \r could survive into parsed data. Since it was a two-line change I pushed it directly to your branch (maintainer edits were enabled) along with a regression test, to save a round trip: f0b8764.

With that, this is ready. It will go out in the next release. Thanks for the contribution and the well-tested PR.

PR #1114, opened 2026-07-23
Fix OpenCode plugin Jinja template corruption with Qwen3.6

atomicflag:
## Issue

The OpenCode plugin's `system.transform` hook was corrupting OpenCode's Jinja template when injecting planning prompts. The error appeared as:

```
raise_exception('System message must be at the beginning.')
```

This occurred when using Qwen3.6 models with their chat template.

## Root Cause

The plugin was pushing planning prompts and improvement context as **separate elements** in the `output.system` array:

```typescript
output.system.push(getPlanningPrompt());
output.system.push(improveContext);
```

OpenCode maps each element of `output.system` to a separate system message. The Qwen3.6 Jinja template only allows **one system message** at position 0:

```jinja
{%- for message in messages %}
    {%- if message.role == "system" %}
        {%- if not loop.first %}
            {{- raise_exception('System message must be at the beginning.') }}
        {%- endif %}
    {%- endif %}
{%- endfor %}
```

When the plugin pushed multiple elements, OpenCode created multiple system messages, causing the template to fail at position > 0.

**Jinja template reference:** https://huggingface.co/Qwen/Qwen3.6-35B-A3B-FP8/raw/main/chat_template.jinja

## Solution

Instead of pushing new elements to `output.system`, the plugin now **appends content to the first element** (`output.system[0]`):

```typescript
output.system[0] += "\n\n" + getPlanningPrompt();
```

This preserves the multi-message `output.system` array structure while keeping all injected content within the **original system message** (element 0). The Jinja template sees only one system message at position 0, containing all the plugin's additions.

## Testing

- Verified the fix works with Qwen3.6 models
- All existing tests pass (10/10 in `plan-mode.test.ts`)
- No breaking changes to other agents or models

### DIFF
--- a/apps/opencode-plugin/index.ts
+++ b/apps/opencode-plugin/index.ts
@@ -24,6 +24,7 @@ import { loadConfig, resolveSharingEnabled } from "@plannotator/shared/config";
 import { readImprovementHook } from "@plannotator/shared/improvement-hooks";
 import { composeImproveContext } from "@plannotator/shared/pfm-reminder";
 import {
+  composeSystemPrompt,
   stripConflictingPlanModeRules,
 } from "./plan-mode";
 import {
@@ -406,32 +407,31 @@ tools (except writing markdown files), or otherwise make changes to the system.
 
       if (shouldInjectFullPlanningPrompt(lastUserAgent, workflowOptions)) {
         const stripped = stripConflictingPlanModeRules(output.system);
-        output.system.length = 0;
-        output.system.push(...stripped);
-        output.system[0] += "\n\n" + getPlanningPrompt();
-
         const hook = readImprovementHook("enterplanmode-improve");
         const pfmEnabled = loadConfig().pfmReminder === true;
         const improveContext = composeImproveContext({
           pfmEnabled,
           improvementHookContent: hook?.content ?? null,
         });
-        if (improveContext) {
-          output.system[0] += "\n\n" + improveContext;
-        }
+        const parts = [...stripped, getPlanningPrompt()];
+        if (improveContext) parts.push(improveContext);
+        output.system.length = 0;
+        output.system.push(...composeSystemPrompt([], parts));
 
         return;
       }
 
       if (!shouldInjectGenericPlanReminder(lastUserAgent, isSubagent, workflowOptions)) return;
 
-      output.system[0] += `\n\n## Plan Submission
+      const planSubmissionReminder = `\n\n## Plan Submission
 
 When you have completed your plan, call the \`submit_plan\` tool to submit it for user review. Pass your full plan as a single edit: \`{ "edits": [{ "start": 1, "content": "..." }] }\`.
 
 The user will review your plan in a visual UI where they can annotate, approve, or request changes. If rejected, the response includes your plan with line numbers; use targeted edits to revise specific sections.
 
 Do NOT proceed with implementation until your plan is approved.`;
+      output.system.length = 0;
+      output.system.push(...composeSystemPrompt(output.system, [planSubmissionReminder]));
     },
 
     // Intercept plannotator commands before the agent sees them.
--- a/apps/opencode-plugin/index.ts
+++ b/apps/opencode-plugin/index.ts
@@ -408,7 +408,7 @@ tools (except writing markdown files), or otherwise make changes to the system.
         const stripped = stripConflictingPlanModeRules(output.system);
         output.system.length = 0;
         output.system.push(...stripped);
-        output.system.push(getPlanningPrompt());
+        output.system[0] += "\n\n" + getPlanningPrompt();
 
         const hook = readImprovementHook("enterplanmode-improve");
         const pfmEnabled = loadConfig().pfmReminder === true;
@@ -417,7 +417,7 @@ tools (except writing markdown files), or otherwise make changes to the system.
           improvementHookContent: hook?.content ?? null,
         });
         if (improveContext) {
-          output.system.push(improveContext);
+          output.system[0] += "\n\n" + improveContext;
         }
 
         return;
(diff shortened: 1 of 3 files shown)


backnotprop:
## Review (at `9d1f57fc`)

Thanks for this, and welcome. The diagnosis is genuinely good work, and we verified it empirically against the real OpenCode binary rather than taking it on faith: `LLMRequestPrep.prepare` maps one `output.system` element to one `{role:"system"}` message exactly as you said, the array arrives pre-joined as a single element so every plugin `push` unavoidably creates a second system message, and OpenCode's own post-hook coalescer can never save the plan path (the strip mutates `system[0]`, so its `l[0] === a` guard fails) and collapses to two messages at best elsewhere. Reproduction confirms main produces 2 to 3 system messages on all three paths and your shape produces 1, which is the only shape Qwen3.6's template accepts. You also covered all three push sites, including the improve-context path, and the PR still merges cleanly onto current main with no rebase needed.

Verdict: **needs changes**, small ones, all in the assignment mechanism rather than the design.

**1. `system[0] +=` can inject the literal string `undefined`.** `stripConflictingPlanModeRules` ends in `.filter(Boolean)` and can return an empty array; then the push leaves `output.system` empty and `output.system[0] +=` evaluates `undefined + "\n\n" + ...`, silently corrupting the prompt. Measured: the resulting first message begins `undefined\n\n## Planning...`. There is no type guard against this (the plugin has no tsconfig and sits outside the repo typecheck).

**2. Index-assignment breaks when another plugin pushed first.** Plugin hooks run sequentially on the same `output` object, so if any plugin ordered before this one pushed an entry, `system[0] +=` splices the planning prompt into the middle of the base prompt and still leaves two elements: the Qwen fix silently fails and the prompt is reordered. `push` at least preserved order.

Both fix with the same collapse-with-order shape, about six lines total across the same hunks:

```ts
// plan path
const parts = [...stripConflictingPlanModeRules(output.system), getPlanningPrompt()];
if (improveContext) parts.push(improveContext);
output.system.length = 0;
output.system.push(parts.filter(Boolean).join("\n\n"));

// generic path
const merged = [...output.system, planSubmissionReminder].filter(Boolean).join("\n\n");
output.system.length = 0;
output.system.push(merged);
```

**3. One ask alongside the fix:** extract the merge into an exported pure helper (for example `composeSystemPrompt(system: string[], additions: string[]): string[]` in `plan-mode.ts`) and add a small table test to the existing `plan-mode.test.ts`: always one element out, the empty-strip case, the multi-element order case, and content order. The transform body is inline in the plugin factory today, which is why nothing covers these lines.

For the record on blast radius, judged acceptable: the change is unconditional across models, content is byte-identical with only the partitioning changed, and the one cache nuance (losing OpenCode's base-prompt/additions split on the generic path) is a fair trade since the appended text is constant within a session.

CI never ran on this fork PR; approving the workflows now. Happy to re-review as soon as the rewrite is pushed.

atomicflag:
@backnotprop The rewrite is ready for review.

atomicflag:
Removed extra `\n\n` in `planSubmissionReminder` since composeSystemPrompt already adds that.

backnotprop:
## Re-review (at `bfc41e4b`)

**TLDR:** The helper, the plan path, and the tests deliver what the first review asked. One blocking defect remains on the generic-reminder path: an evaluation-order bug clears `output.system` before the helper reads it, so the base prompt and other plugins' entries are deleted instead of merged. Two-line fix, plus one whitespace trim and one test row, and this merges.

*AI-assisted review; findings verified by execution, not inspection.*

**Blocker, `index.ts:433-434`:**

```ts
output.system.length = 0;
output.system.push(...composeSystemPrompt(output.system, [planSubmissionReminder]));
```

`length = 0` executes before the argument is evaluated, so `composeSystemPrompt` receives an already-empty array. Verified by running the two lines: input `["OPENCODE BASE PROMPT", "OTHER PLUGIN ENTRY"]` produces output containing only the reminder; both prior entries are gone. That is worse than main, which merely pushed an extra element. Your plan path at `:416-419` is correct precisely because it builds `parts` before clearing; do the same here:

```ts
const composed = composeSystemPrompt(output.system, [planSubmissionReminder]);
output.system.length = 0;
output.system.push(...composed);
```

**Alongside the fix, two small things:**

1. `plan-mode.ts:58`: add `.map(s => s.trim())` before `filter(Boolean)`. The plan path is only clean because `stripConflictingPlanModeRules` trims; real OpenCode entries end in newlines, and the generic path currently composes runs of 3-4 blank lines once the blocker is fixed. Your `bfc41e4b` newline tweak itself is right.
2. One more table row: `[["", "a"], ["b"]] -> ["a\n\nb"]`. We mutation-tested your suite: it catches the original `+=` bug, order reversal, and failure to collapse, but dropping `filter(Boolean)` passes all 15 tests because the empty-strip row uses an empty array rather than an empty string among entries.

Everything else holds: no `undefined` injection anywhere, both call sites route through the exported helper, 112/112 plugin tests pass at your head, no unrelated changes, and the branch sits directly on current main.

One note for later, not on you: `server.ts:132` (the OpenCode 2 adapter) pushes a text part on its generic path. Different message shape, so not the Qwen3.6 bug, but we will confirm it independently before closing the issue as fully fixed.

Happy to re-review immediately when you push.

atomicflag:
Addressed the bug and 2 points.

backnotprop:
Re-verified at `5247d632`. The blocker is fixed and confirmed by execution: both prior entries survive in order with the reminder last, and the empty-string test row now catches the filter regression it was added for. 113/113 plugin tests pass.

One item remains, and it is the last one: the `.trim()` landed in `stripConflictingPlanModeRules`, where it is a no-op (`cleanupSystemEntry` already trims every element there) and only covers the plan path, which was already clean. The generic path still composes `"base\n" + "second\n\n"` into runs of 3-4 blank lines. Move it into the helper itself:

```ts
return [system.concat(additions).map((s) => s.trim()).filter(Boolean).join("\n\n")];
```

drop the now-redundant map at `plan-mode.ts:55`, and add a row like `[["base\n"], ["add\n\n"], ["base\n\nadd"]]` so it stays pinned. Push that and this merges.

*AI-assisted verification, by execution.*

atomicflag:
Moved the trim to the proper location and added a trimming test case.

backnotprop:
Verified at `6bcba39a`: the helper now trims exactly where it matters, the generic-path probe composes cleanly with no blank-line runs, and the full plugin suite passes 114/114 with the new trimming row. All three review rounds are resolved. This is ready to merge; flagging for the maintainer. Thanks for the fast, precise turnarounds @atomicflag.

PR #1122, opened 2026-07-25
fix(ui): move Save to the first tab stop in the comment popover footer

### LINKED ISSUE #1121: Comment popover: Save button is the last tab stop, behind the image-upload icon
In the annotation comment popover (`CommentPopover.tsx`), the footer is laid out `[image icon] ... [Ask AI] [⌘↵ hint] [Save]`. Tab order follows DOM order, so Save — the primary/most-used action — is currently the *last* focusable control, after the image-upload button.

For keyboard-driven review flows (select text → comment → Tab, Tab, Enter), this means a fast tabber is more likely to land on the image-upload button than on Save, which is backwards from intent.

**Proposed fix:** reorder the footer so Save is the first control after the textarea, ahead of Ask AI and the image icon — matching visual left-to-right priority to keyboard priority.

I have a small PR ready that does just this (pure DOM reorder, no behavior change) — will link it here.

balasivagn:
## Summary
- Reorders the comment popover footer (`packages/ui/components/CommentPopover.tsx`) so **Save is the first focusable control** after the textarea, ahead of Ask AI and the image-upload button.
- Tab order previously matched left-to-right DOM order (`image icon → Ask AI → Save`), so a fast keyboard flow (select text → comment → Tab, Tab, Enter) could land on the image-upload button instead of Save — the opposite of what you'd want for the primary action.
- Pure reorder: no handler, class, or behavior changes. `Mod+Enter`/`Escape` are unaffected (they're wired on the textarea's `onKeyDown`, independent of button DOM position). Applied identically to both duplicated footer blocks in the file (dialog/expanded mode and popover mode). Since `CommentPopover` is shared, this fixes both the plan-review and code-review surfaces.

Fixes #1121

## Test plan
- [x] `packages/ui` typechecks clean (`tsc --noEmit` on `packages/ui/tsconfig.json` and `tsconfig.strict-consumer.json`)
- [x] `bun test` — full suite passes (3 pre-existing, unrelated failures in `packages/core/crypto.test.ts` hitting the live paste service over the network, not touched by this change)
- [x] Rebuilt `apps/review` + `apps/hook` bundles and manually verified in the review-editor dev server: Save appears leftmost, and tabbing from the textarea reaches Save before Ask AI / the image icon, in both the inline popover and expanded dialog variants

Closes #1121

### DIFF
--- a/packages/ui/components/CommentPopover.tsx
+++ b/packages/ui/components/CommentPopover.tsx
@@ -322,17 +322,14 @@ export const CommentPopover: React.FC<CommentPopoverProps> = ({
 
           {/* Footer */}
           <div className="flex items-center justify-between px-4 py-3 border-t border-border/50">
-            <div className="flex items-center gap-2">
-              {allowImages && (
-                <AttachmentsButton
-                  images={images}
-                  onAdd={(img) => setImages((prev) => [...prev, img])}
-                  onRemove={(path) => setImages((prev) => prev.filter((i) => i.path !== path))}
-                  variant="inline"
-                />
-              )}
-            </div>
             <div className="flex items-center gap-3">
+              <button
+                onClick={handleSubmit}
+                disabled={!canSubmit}
+                className="px-3 py-1.5 text-xs font-medium rounded-md bg-primary text-primary-foreground hover:opacity-90 disabled:opacity-50 disabled:cursor-not-allowed transition-opacity"
+              >
+                {isGlobal ? 'Add' : 'Save'}
+              </button>
               {onAskAI && (
                 <button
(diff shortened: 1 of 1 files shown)


backnotprop:
Thanks for the PR and for catching this, it was a real accessibility gap.

One thing we want to preserve: Save stays visually bottom-right, which is the convention across the app. So I pushed a commit onto your branch that keeps your goal (Save is the first tab stop after the textarea) but gets there with DOM order plus `flex-row-reverse`, leaving the visual layout unchanged. Screen readers also announce Save first now, since they follow DOM order too.

Tab order from the textarea is Save, then Ask AI, then image upload.

PR #1124, opened 2026-07-25
fix(pi): preserve extension tool selections

CodeByPeete:
## Summary

- preserve tools activated or deactivated by other Pi extensions while Plannotator changes phases
- track and remove only tools added by Plannotator's current phase
- stop enabling fixed discovery tools or advertising a fixed tool inventory

## Validation

- `bun test ./apps/pi-extension/tool-scope.test.ts ./apps/pi-extension/config.test.ts ./apps/pi-extension/phase-tools-runtime.test.ts`

The focused suite passes 18 tests. The full monorepo test run reached 2,201 passing tests. Existing Windows-specific path, executable, and locked temporary-directory failures remain outside the Pi extension changes.


backnotprop:
Reviewed against the Pi source. The approach is right. Two changes before merge:

1. The plan-complete path (`agent_end` handler, index.ts ~1244) transitions to idle without calling `releaseAddedPhaseTools()` or clearing `phaseAddedTools`, unlike `exitToIdle`. Custom phase tools leak into idle. Add the release plus a test for that path.

2. Keep `grep`/`find`/`ls` in planning. They are inactive by default in Pi, so dropping the force-add silently removes native search from planning sessions. Make them the default planning `activeTools` under your tracking model, added on entry and released on exit like any other phase tool.

Nice work on the preserve logic, the tests mirror Pi's real semantics.

backnotprop:
We went ahead and pushed the two fixes from the review so this can move: the agent_end plan-completion path now calls releaseAddedPhaseTools() like exitToIdle does, and grep/find/ls are back as default planning-phase tools, now going through your tracking mechanism (recorded in phaseAddedTools, released on exit, and left alone if another extension already activated them). Tests cover both, including the case where grep was already active before planning. Happy to adjust if you see a problem with either.

backnotprop:
Final review caught one last thing: because planning tools now come purely from the resolved profile and `activeTools` uses replace semantics, any user config setting `phases.planning.activeTools` would silently drop `plannotator_submit_plan` while the planning prompt still told the model to call it. We pushed the union fix (submit tool is always added for the planning phase, deduped, and released on exit like any other tracked addition) plus a test and a README correction, and merged.

Thanks for the clean tracking model, the phaseAddedTools approach made this a two-line fix.

PR #1125, opened 2026-07-25
feat(pi): submit plans from any mode

CodeByPeete:
## Summary

- keep `plannotator_submit_plan` available while Plannotator is idle, planning, or executing
- preserve the existing transition into execution when a planning-phase submission is approved
- treat submissions from idle or executing as review-only, leaving the current phase and tracked plan unchanged

## Validation

- `bun test ./apps/pi-extension/tool-scope.test.ts ./apps/pi-extension/submit-plan-runtime.test.ts`

The focused suite passes 10 tests. The Pi extension suite reaches 58 passing tests. Existing Windows-only semantic-diff executable and locked temporary-directory failures remain outside this change.

### DIFF
--- a/apps/pi-extension/index.ts
+++ b/apps/pi-extension/index.ts
@@ -65,7 +65,6 @@ import {
 	isPlanWritePathAllowed,
 	PLAN_SUBMIT_TOOL,
 	type Phase,
-	stripPlanningOnlyTools,
 } from "./tool-scope.ts";
 import { isRemoteSession } from "./server/network.ts";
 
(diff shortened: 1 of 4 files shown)


backnotprop:
Heads up: this conflicts with #1124, which lands first. This branch keeps the snapshot restore that #1124 removes, so it reintroduces the clobbering bug. After #1124 merges, please rebase and express submit-from-any-mode under its tracking model (keep the submit tool out of phaseAddedTools rather than restoring snapshots).

PR #1126, opened 2026-07-25
feat(pi): add external plan execution handoff

cgngtr:
## Summary

- add an opt-in `executionMode: "external"` setting while preserving automatic execution by default
- restore the pre-planning Pi state and emit a typed `plannotator:plan-approved` event instead of entering the executing phase
- include the plan cwd, path, content, and optional approval feedback in the handoff payload
- document the companion-extension contract and cover config precedence and `--plan` state restoration

## Testing

- `bun test apps/pi-extension` (76 passed, 1 skipped)
- `git diff --check`


backnotprop:
Thanks for the clean contract work here. The typed event, the config plumbing, and the docs all read well.

I merged current main into the branch and pushed a couple of follow-ups:

- #1124 landed phase-added tool tracking, and under that model every idle-exit path has to hand the phase tools back. The handoff path was still missing that call, so approving in `external` mode leaked the planning tool scope (`grep`, `find`, `ls`, `plannotator_submit_plan`) into the idle session. Your own test caught it on the merged tree. I extracted the shared exit sequence into `returnToIdle()` and routed all three idle transitions through it, so a future exit path cannot skip a step.
- Added a default-mode parity test: with `executionMode` unset, approval still enters executing, emits no `plannotator:plan-approved`, appends the `plannotator-execute` entry, and sends the continuation message.
- Two small touches: a warning when `executionMode` has an unrecognized value (it still falls back to automatic), and README notes that `planFilePath` is relative to `cwd` and that companions can subscribe with the raw channel string.

Typecheck and the pi-extension suite are green (113 pass, 0 fail) and CI is passing. Review verdict is positive. This is ready for the maintainer's final merge call.

PR #1131, opened 2026-07-26
fix(opencode): Fix hardcoding default `build` OpenCode agent when sending responses

### LINKED ISSUE #1123: [OpenCode] Allow customizing the review agent name that feedback gets sent back to instead of `build`
## Context
I started using Plannotator to review diffs and send feedback back to my OpenCode session, but I got an error in OpenCode's TUI that it could not find `build` agent. I have the `build` agent disabled in my opencode.json since I have a differently named one as my primary.

## Request
Would you open to supporting a plugin option to define the build agent?

```json
["@plannotator/opencode@latest", {
  "buildAgentName": "MyCustomAgentNameHere",
}]
```

Happy to submit a PR if you're aligned 👍

cb-bradbeebe:
## Summary

- Fixes #1123 by removing OpenCode's implicit fallback to `build` when agent switching is unset/disabled.
- Keeps execution on the current agent unless a valid explicit agent is configured.
- Prevents invalid/unavailable configured agents from being sent to OpenCode.

## What changed

- Updated OpenCode agent-switch behavior so default/disabled mode does **not** set an `agent` parameter.
- Added validation for explicitly configured agents via `client.app.agents()` before request submission.
- If a configured agent is invalid/unavailable, the agent override is omitted and surfaced as a warning (log + toast).
- Updated related docs and tests.

## Testing

- Unit tests
- Marketing docs builds
- Manual testing with local plugin build

#### Before
before-01
before-02

#### After
after-01

### New warning toast if user manually enters a invalid custom agent via Settings
warning-settings-01

warning-toast-01

### DIFF
--- a/apps/marketing/src/content/docs/commands/plan-review.md
+++ b/apps/marketing/src/content/docs/commands/plan-review.md
@@ -106,7 +106,7 @@ Images are stored as temporary files and referenced by name in the feedback sent
 
 ## Agent switching (OpenCode)
 
-OpenCode users can configure which agent receives the approved plan. By default, approved plans stay on the current agent. See [Agent switching settings](/docs/getting-started/ui-settings/#agent-switching).
+OpenCode users can configure which agent receives the approved plan. By default, approved plans are handed off to the build agent. See [Agent switching settings](/docs/getting-started/ui-settings/#agent-switching).
 
 If the configured agent isn't found in the current OpenCode session, Plannotator logs a warning and shows a warning toast, then omits the agent switch so the current/default OpenCode flow continues.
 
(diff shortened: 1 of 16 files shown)


backnotprop:
Thanks for the fix. We kept your validation path and the review-feedback behavior exactly as you built it: agents are validated through client.app.agents(), an unavailable agent is omitted with a log plus toast, and the empty custom-name case no longer leaks "custom" as an agent name.

One change on top: the agent switch cookie is shared by plan approval and code review, so flipping the stored default to disabled also removed the OpenCode plan-approval handoff for everyone who never touched the setting. I pushed a commit that makes the unset default surface-aware instead. Plan approval keeps defaulting to the build handoff, so existing plan-mode users see no change, and review feedback now defaults to staying on the current agent as you intended. An explicit choice in Settings still applies everywhere.

PR #1141, opened 2026-07-27
fix(hook): read annotate-last from Claude Code's transcript tree, not file order

BrandonNoad:
## The bug

Claude Code session logs are append-only and tree-shaped. Every entry records the entry it follows in `parentUuid`, and `/rewind` writes nothing of its own: the next committed message simply re-parents to the rewind target, and every entry that followed that target stays in the file, permanently orphaned.

`extractRecentRenderedMessages` scanned bottom-up in file order, so orphaned messages continued to appear in the `annotate-last` message picker despite no longer being part of the conversation.

## The fix

`resolveActiveBranchIndices` walks `parentUuid` from the newest id-bearing entry back to the root.

- It returns indices rather than a filtered array, so callers continue to report real file line numbers (`lineNumbers` derives from the loop index).
- It returns `null` for any chain that cannot be trusted (no ids present, a dangling parent, a cycle), allowing callers to degrade to the previous file-order read rather than returning nothing.
- The newest entry is not necessarily the last line. `last-prompt`, `ai-title`, `mode` and `file-history-snapshot` carry no ids and are frequently written last.

The behavior is opted into at the Claude Code call site only. Droid's call site, Codex and Copilot (separate parsers), and Pi, OpenCode and Amp (live APIs, no transcript reads) are all unaffected. Both new parameters default to off.

## Scope of the behavior change

Verification against a corpus of real session logs showed every chain resolving cleanly to the root, with no dangling parents or cycles, and the default selection unchanged in every transcript containing a message.

The default not moving is expected rather than evidence that the change is inert. A committed rewind's new branch is always the newest region of the file, so file order and the tree already agree on the most recent message. The user-visible improvement is in the picker, which no longer offers messages that have left the conversation.

## Out of scope: the uncommitted rewind

This does not address the case of rewinding and then immediately running `/plannotator-last`, which cannot be fixed by reading alone. The skill invokes the binary through `` !`plannotator annotate-last $ARGUMENTS` ``, and Claude Code cannot write the invoking message until that command returns, because the message is incomplete until the substitution has been filled. At read time the rewind is therefore absent from disk entirely.

Session logs confirm the ordering: the invocation's transcript entry is timestamped well after the command has run and the annotate window has closed. A rewind that is never followed by a message also does not survive quitting and resuming the session, so no external process can observe one.

## Rationale for the current scope

Switching the skill from the substitution to a Bash-tool invocation would fix the default, since Claude Code writes the message before running tools. That was deliberately left out here: it alters every invocation rather than only rewinds, and it adds a model round-trip for a comparatively rare situation. Reading the tree correctly stands on its own as a bug fix, and it is the prerequisite in either design. With it in place, the rewind case can be covered by a personal skill that invokes `plannotator annotate-last` through the Bash tool, leaving shipped behavior unchanged for everyone else.

## Tests

Fixtures previously assigned random `parentUuid` values, which left every entry an orphan and made branch resolution untestable. `buildLog` now links fixture lines into a single chain, and `buildRewoundLog` models a fork. No existing assertion depends on id values.

New coverage: a linear log, a rewind, a tail carrying no id, a log with no ids at all, a dangling parent, a cycle, line-number preservation after filtering, and the fallback to a file-order read.

### DIFF
--- a/apps/hook/server/index.ts
+++ b/apps/hook/server/index.ts
@@ -1213,7 +1213,12 @@ if (args[0] === "sessions") {
         console.error(`[DEBUG] ${label}: ${paths.length ? paths.join(", ") : "(none)"}`);
       }
       for (const logPath of paths) {
-        const recent = getRecentRenderedMessages(logPath, RECENT_MESSAGES_LIMIT);
+        // Claude Code transcripts are trees: `/rewind` re-parents the next
+        // message rather than truncating, so a file-order read returns
+        // orphaned messages. Follow the id chain instead.
+        const recent = getRecentRenderedMessages(logPath, RECENT_MESSAGES_LIMIT, {
+          activeBranchOnly: true,
+        });
         if (recent.length > 0) {
           recentMessages = recent;
           lastMessage = recent[0];
--- a/apps/hook/server/session-log.test.ts
+++ b/apps/hook/server/session-log.test.ts
@@ -14,6 +14,7 @@ import {
   findAnchorIndex,
   extractLastRenderedMessage,
   extractRecentRenderedMessages,
+  resolveActiveBranchIndices,
   findDroidSessionLogsForCwd,
   resolveDroidSessionLogForCwd,
   projectSlugFromCwd,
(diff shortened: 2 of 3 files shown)


backnotprop:
Great work here, and the write-up made review easy. We tested this against a few thousand real transcripts: linear sessions are identical and the default pick never changes, so this is safe for everyone who never rewinds.

One case needed a guard: a /compact boundary also has parentUuid null, so right after a compaction the walk returns zero messages. The caller treats empty as "wrong log file" and moves on to an older session log. I pushed a small commit to your branch: an empty branch-filtered result falls back to the file-order read, plus tests for the cut, the fallback, and post-compaction recovery. Take a look and if it makes sense to you, this is good to merge.

BrandonNoad:
> Great work here, and the write-up made review easy. We tested this against a few thousand real transcripts: linear sessions are identical and the default pick never changes, so this is safe for everyone who never rewinds.
> 
> One case needed a guard: a /compact boundary also has parentUuid null, so right after a compaction the walk returns zero messages. The caller treats empty as "wrong log file" and moves on to an older session log. I pushed a small commit to your branch: an empty branch-filtered result falls back to the file-order read, plus tests for the cut, the fallback, and post-compaction recovery. Take a look and if it makes sense to you, this is good to merge.

Falling back to file-order read in this case makes sense to me. Better than moving to an older session log. Apparently compact boundaries also carry a `logicalParentUuid` pointing at the pre-compaction parent, so we could try following that when `parentUuid` is `null`. Though I think what we are doing now is fine. A user is unlikely to call `/plannotator-last` after a compaction.

Issue #1144, opened 2026-07-27
Opening any review launches Codex: model discovery runs while the AI runtime is constructed


rNoz:
### What happened?

Opening a plan, annotate, or code review builds the shared AI runtime, and the runtime calls every registered provider's `fetchModels()` while constructing itself. The Codex provider's implementation starts a throwaway `codex app-server` process, so opening a review executes Codex even when Ask AI is never opened and Codex is never selected.

The process is short lived and is cleaned up, so on most machines nothing is visible and this looks like nothing is wrong. It becomes visible when starting the Codex binary itself prompts: on macOS with a quarantined Homebrew payload that is a Gatekeeper confirmation dialog in front of the review that was actually requested. Even when it is invisible it is an unnecessary subprocess for a provider nobody selected.

This is a regression from #971 (2026-06-28), which moved the `codex-sdk` provider onto `codex app-server`. Before that commit the Codex provider had no `fetchModels()` at all, so runtime construction started no process. Anyone who used Plannotator before late June and saw nothing was not missing it: there was nothing to see.

The editor's automatic `/api/ai/capabilities` probe has the same effect, because it awaits the same discovery.

Reproduction, with no real Codex install needed. Any executable named `codex` on `PATH` shows the launch:

1. Create a fake `codex` that records its invocation.

   ```sh
   mkdir -p /tmp/codex-probe
   printf '#!/bin/sh\necho "$@" >> /tmp/codex-probe/invoked.log\nsleep 30\n' > /tmp/codex-probe/codex
   chmod +x /tmp/codex-probe/codex
   ```

2. With `/tmp/codex-probe` first on `PATH`, construct the shared AI runtime the way a review does, and do nothing else.

   ```ts
   const { createAIRuntime } = await import("./packages/server/ai-runtime.ts");
   await createAIRuntime({ cwd: process.cwd() });
   ```

3. `/tmp/codex-probe/invoked.log` exists and contains `app-server`, before any review UI has loaded and before any AI session exists.

Observed on a pristine clone of `main` `a54b46bb` with its own `bun install`: `{"afterConstruct":true,"afterCapabilities":true}`, and the fake recorded the argument `app-server`. Repeating it with the real Homebrew binary (`codex-cli 0.144.6`) and a 60ms process poll caught `/opt/homebrew/bin/codex app-server` running during `createAIRuntime()`. Both `packages/server/ai-runtime.ts` and `apps/pi-extension/server/ai-runtime.ts` behave this way.

[...]

Issue #971, opened 2026-06-26
Document chat / Ask launches Codex with approval_policy=Never, breaking enterprise-managed Codex requirements


crunchy234:
## Summary

Plannotator's document chat / "Ask" flow appears to launch Codex with `approval_policy = Never`, which breaks in enterprise-managed environments where that policy is disallowed.

In my environment, Codex only allows `OnRequest` and `UnlessTrusted`, so the document chat flow shows an approval-policy error and does not behave correctly.

## Environment

- Plannotator: local install
- Codex: desktop/local Codex runtime
- Org setup: enterprise-managed Codex requirements
- Allowed approval policies: `OnRequest`, `UnlessTrusted`

## Steps to reproduce

1. Install and configure Plannotator for Codex.
2. Open a document in Plannotator.
3. Use the document chat / "Ask" feature.
4. Observe the launched Codex session / chat behavior.

## Actual behavior

The document chat flow shows this error:

```text
Configured value for `approval_policy` is disallowed by requirements; falling back to required value OnRequest. Details: invalid value for `approval_policy`: `Never` is not in the allowed set [OnRequest, UnlessTrusted] (set by enterprise-managed requirements ...)
```

In practice, this makes the document chat flow broken or unreliable.

## Expected behavior

Plannotator should launch Codex in a way that is compatible with enterprise-managed requirements.

For example, it should:
- inherit the user's allowed/default Codex approval policy, or
- use `OnRequest`, or
- expose the approval policy for this flow as a configurable option

## Investigation notes

This does not appear to be caused by:
- the normal Plannotator binary install
- Codex hooks setup
- my visible Codex config file

From local debugging, the problematic sessions appear to be launched as a `codex-auto-review` / `guardian`-style flow, and the recorded session config shows:
- `approval_policy = "never"`
- `sandbox_policy = "read-only"`

That suggests this is specifically related to how the document chat / Ask flow launches Codex, rather than a general install issue.

## Related issues

These seem related but do not appear to cover this exact bug:
- #497 Support for OpenAI Codex Hooks
- #903 IDE and runtime integration parity matrix

This looks like a separate compatibility bug affecting enterprise-managed Codex environments.

## Questions

- Is `approval_policy = Never` intentionally forced for this flow?
- Can this flow inherit the user's Codex policy instead?
- Is there a supported workaround for enterprise-managed environments?

PR #1145, opened 2026-07-27
fix(ai): defer Codex model discovery until a Codex session starts

rNoz:
## Summary

Constructing the shared AI runtime called every provider's `fetchModels()`, and the Codex provider's implementation spawns `codex app-server`. Opening any plan, annotate, or code review therefore executed Codex even when the user never opened Ask AI. Codex discovery now runs on explicit activation instead.

The eager launch dates from #971, which moved `codex-sdk` onto `codex app-server`. Before that the provider had no `fetchModels()`, so runtime construction started no process.

Solves https://github.com/backnotprop/plannotator/issues/1144, which is quite annoying to receive the popup about "codex" on every plannotator interaction.

## Why

The spawned process is short lived and cleaned up, so usually nothing is visible. It becomes visible when starting the binary itself prompts: on macOS with a quarantined Homebrew payload, a Gatekeeper confirmation dialog in front of the review the user asked for. Invisible or not, it is a subprocess for a provider nobody selected, started by opening a review. The editor's automatic `/api/ai/capabilities` probe had the same effect, since it awaited the same discovery.

## Changes

- Add `createBestEffortOnce()` in `packages/ai/endpoints.ts`, memoizing an initializer so it runs at most once per runtime and a failure is swallowed, leaving the provider's static fallback metadata in place instead of blocking session creation.
- Add `beforeProviderSession?(providerId)` to `AIEndpointDeps`, invoked by `/api/ai/session` for the resolved provider before the session is created. `/api/ai/capabilities` deliberately does not invoke it: the editor probes capabilities automatically on load, so activating there would reintroduce the same eager launch through another path.
- Register Codex exactly as before in both runtimes, but store the memoized discovery function in a per-provider map instead of calling `fetchModels()` during construction. Every other provider keeps its existing eager discovery, which `beforeCapabilities` still awaits.
- Adopt the post-activation default model when the caller sent no model or sent the pre-activation default, and honor an explicitly chosen model unchanged. Without this a first Codex session would pin the static fallback model that discovery has just replaced.

Codex stays listed by `/api/ai/capabilities` throughout, on its fallback model metadata until the first Codex session.

## Validation

- `bun run typecheck` across core, shared, AI, server, UI, strict consumer, and Pi: passed.
- `packages/ai/ai.test.ts`: 108 passed.
- `packages/server/ai-runtime.test.ts` and `apps/pi-extension/server/ai-runtime.test.ts`: 1 passed each. Each spawns a child process with a fake `codex` on `PATH` and asserts `{ hasCodex: true, afterCapabilities: false, sessionStatus: 200, afterSession: true }`, so Codex is still advertised, not executed by the capabilities probe, and executed by the first Codex session.
- Manual probe on a pristine clone of `main` with its own `bun install`, so no local branch or build artifact is involved. A fake `codex` records its invocation, and the real Homebrew binary (`codex-cli 0.144.6`) was confirmed separately with a 60ms process poll catching `codex app-server` during runtime construction. macOS, Bun 1.3.14:

| Runtime state | `codex` invoked during `createAIRuntime()` | after `/api/ai/capabilities` | after first `codex-sdk` session |
|---|---|---|---|
| `main` `a54b46bb` | yes (`app-server`) | yes | n/a |
| this branch | no | no | yes |

Not covered: real Codex binaries, the Gatekeeper dialog itself, reasoning-effort metadata refresh in the UI after activation, and startup-latency measurement.

## Real e2e testing

I am using plannotator on a daily basis. I use a custom local-only integration branch including this branch/PR and also https://github.com/backnotprop/plannotator/pull/1143, to fully review/check/test it in real scenarios. So far so good. Please, feel free to ask me about further testing, tweaks or refactors (or modify directly the branch yourself, it's good!). Happy to help, as you know me!

### DIFF
--- a/apps/pi-extension/server/ai-runtime.test.ts
+++ b/apps/pi-extension/server/ai-runtime.test.ts
@@ -78,4 +78,75 @@ describe("createPiAIRuntime Codex discovery", () => {
 			afterSession: true,
 		});
 	}, 15_000);
+
+	test("capabilities?activate= runs discovery once and shares it with the session path", async () => {
+		if (process.platform === "win32") return;
+
+		const dir = mkdtempSync(join(tmpdir(), "plannotator-pi-activate-codex-"));
+		tempDirs.push(dir);
+		const marker = join(dir, "codex-ran");
+		const codex = join(dir, "codex");
+		writeFileSync(codex, `#!/bin/sh\necho ran >> '${marker}'\nexit 1\n`);
+		chmodSync(codex, 0o755);
+
+		const runner = join(dir, "runner.ts");
+		const runtimeUrl = pathToFileURL(join(import.meta.dir, "ai-runtime.ts")).href;
+		writeFileSync(runner, `
+			import { existsSync, readFileSync } from "node:fs";
+			import { createPiAIRuntime } from ${JSON.stringify(runtimeUrl)};
+			const runs = () => existsSync(${JSON.stringify(marker)})
+				? readFileSync(${JSON.stringify(marker)}, "utf8").trim().split("\\n").length
+				: 0;
+			const runtime = await createPiAIRuntime({ cwd: ${JSON.stringify(dir)} });
+			if (!runtime) throw new Error("Pi AI runtime unavailable");
+			const probe = await runtime.endpoints["/api/ai/capabilities"](
+				new Request("http://localhost/api/ai/capabilities"),
+			);
+			const runsAfterProbe = runs();
+			const activate = await runtime.endpoints["/api/ai/capabilities"](
+				new Request("http://localhost/api/ai/capabilities?activate=codex-sdk"),
+			);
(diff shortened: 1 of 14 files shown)


backnotprop:
Reviewed this in depth. The deferral itself is exactly right and we verified the no-spawn boundary holds. It did leave the client stuck on the static fallback model list though: the picker froze, the reasoning effort control disappeared, and the fallback id could overwrite a saved model preference.

We pushed a fix-up commit to your branch (d0058f97) rather than bounce it back: an explicit activate call on the capabilities endpoint (runs your same createBestEffortOnce initializer, only on user gesture), no-clobber semantics for saved preferences, and a membership check for the session model. Your commits are untouched and your tests all still pass. Take a look, and if it reads right to you this is ready to merge.

rNoz:
Thanks for the fix-up addition. Totally cool to merge, go for it.

PR #1155, opened 2026-07-30
feat(review): add Cmd/Ctrl+Shift+Y shortcut to copy feedback

rian-dolphin:
This change adds a keyboard shortcut to the code review UI. The shortcut is `Cmd/Ctrl+Shift+Y`. The shortcut copies the review feedback to the clipboard.

Before this change, the user must click two controls. First, the user clicks the comment icon in the header. Then the user clicks the **Copy Feedback** button.

## Method

The `handleCopyFeedback` callback is available already in `App.tsx`. A new effect connects the keys to that callback. The effect uses the same guards as the `Cmd/Ctrl+Enter` effect:

- The effect does not operate when the focus is in an input field or a text area.
- The effect does not operate when a dialog is open.

A new entry in `packages/review-editor/shortcuts.ts` declares the shortcut. Thus the Settings list shows the shortcut automatically.

## Key selection

`Mod+Shift+Y` is free in the code review registry. `Mod+Shift+C` is not free, because the browser uses those keys for the developer tools. Tell me if you prefer different keys or push a commit to update etc.

## Tests

`bun test` gives 2440 pass and 22 fail. The same 22 tests also fail on `main`. Thus these failures are not related to this change.

### DIFF
--- a/packages/review-editor/App.tsx
+++ b/packages/review-editor/App.tsx
@@ -2842,6 +2842,27 @@ const ReviewApp: React.FC = () => {
     handleApprove, handleSendFeedback, handlePlatformAction
   ]);
 
+  // Cmd/Ctrl+Shift+Y keyboard shortcut to copy feedback, mirroring the
+  // Copy Feedback button in the header and the review sidebar.
+  useEffect(() => {
+    const handleKeyDown = (e: KeyboardEvent) => {
+      if (e.key.toLowerCase() !== 'y' || !(e.metaKey || e.ctrlKey) || !e.shiftKey) return;
+
+      const tag = (e.target as HTMLElement)?.tagName;
+      if (tag === 'INPUT' || tag === 'TEXTAREA') return;
+      if (platformCommentDialog || showExportModal || showNoAnnotationsDialog || showApproveWarning || showExitWarning) return;
+
+      e.preventDefault();
+      handleCopyFeedback();
+    };
+
+    window.addEventListener('keydown', handleKeyDown);
+    return () => window.removeEventListener('keydown', handleKeyDown);
+  }, [
+    platformCommentDialog, showExportModal, showNoAnnotationsDialog, showApproveWarning, showExitWarning,
+    handleCopyFeedback
+  ]);
+
   if (isLoading) {
     return (
       <ThemeProvider defaultTheme="dark">
--- a/packages/review-editor/App.tsx
+++ b/packages/review-editor/App.tsx
@@ -2525,10 +2525,12 @@ const ReviewApp: React.FC = () => {
     if (await copyTextToClipboard(feedbackMarkdown)) {
       setCopyFeedback('Feedback copied!');
       setTimeout(() => setCopyFeedback(null), 2000);
+      toast.success('Feedback copied');
     } else {
       console.error('Failed to copy');
       setCopyFeedback('Failed to copy');
       setTimeout(() => setCopyFeedback(null), 2000);
+      toast.error('Failed to copy');
     }
   }, [totalAnnotationCount, feedbackMarkdown]);
 
@@ -2843,13 +2845,11 @@ const ReviewApp: React.FC = () => {
   ]);
 
   // Cmd/Ctrl+Shift+Y keyboard shortcut to copy feedback, mirroring the
-  // Copy Feedback button in the header and the review sidebar.
+  // Copy Feedback button in the header.
   useEffect(() => {
     const handleKeyDown = (e: KeyboardEvent) => {
-      if (e.key.toLowerCase() !== 'y' || !(e.metaKey || e.ctrlKey) || !e.shiftKey) return;
+      if (!(e.metaKey || e.ctrlKey) || !e.shiftKey || e.altKey || e.key.toLowerCase() !== 'y' || isTypingTarget(e.target)) return;
 
-      const tag = (e.target as HTMLElement)?.tagName;
-      if (tag === 'INPUT' || tag === 'TEXTAREA') return;
       if (platformCommentDialog || showExportModal || showNoAnnotationsDialog || showApproveWarning || showExitWarning) return;
 
       e.preventDefault();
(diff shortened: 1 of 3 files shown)


backnotprop:
## Review (at `36037e89`)

Thanks for this, and welcome. The shortcut choice itself is verified sound: all 32 bindings across the review editor and the seven code-review scopes were enumerated programmatically and `Y` appears in exactly one binding, yours. Vim yank is dead in the review editor, double-tap bindings do not clash, the token normalizes correctly in the registry tests, and the auto-generated docs page renders `Cmd/Ctrl+Shift+Y` in the right slot with your displayOrder choice fitting cleanly. The PR also merges into current main with zero conflicts and correctly reaches the hardened clipboard helper from #1174 because it calls `handleCopyFeedback` rather than duplicating the copy. CI green, typecheck and full suite pass on the merged tree.

Verdict: **needs changes**, all small and inside your 27 lines.

**1. The shortcut is invisible in the app.** The PR body says the Settings list shows it automatically, but the in-app Settings keyboard panel renders a hard-coded array (`packages/ui/components/KeyboardShortcuts.tsx:173-213`); the registry entry you added feeds only the tests and the marketing docs page. Add a row to the `reviewShortcuts` Actions section in that file so users can discover it.

**2. No confirmation in real sessions.** The only visual "copied" affordance (`copyFeedback` at `App.tsx:3196`) lives inside the demo/standalone branch of the header. In agent-backed sessions the shortcut copies silently. The file already has the convention to follow: `toast.success('Agent instructions copied')` at `App.tsx:602`. A `toast.success('Feedback copied')` in the handler covers every mode.

**3. The handler is looser than the binding it declares.** It omits an `altKey` check, so `Mod+Alt+Shift+Y` also fires, and AltGr layouts (Ctrl+Alt) can trigger it while typing. Sibling handlers exclude alt (`App.tsx:1286`, `:1293`); add `!e.altKey`.

**4. Use the shared typing guard.** The handler checks `INPUT`/`TEXTAREA` tags directly; the codebase uses `isTypingTarget` (`packages/review-editor/hooks/useReviewSearch.ts:16`), which also covers contentEditable and is what the neighboring handlers use.

**One decision for the maintainer, not you:** `Ctrl+Shift+Y` opens the Downloads Library in Firefox on Linux, and `preventDefault` runs after the guards, so with focus in an input the browser action still wins. Since remote/Linux is a supported path, we may keep the binding and accept it, or pick another letter. Leaving the binding as is unless told otherwise is fine.

Also noted for our own backlog, not yours: the app-level review shortcuts are all raw window listeners rather than the `useShortcutScope` hook (the exported surface hooks are consumed nowhere), so your PR faithfully matches the file's existing pattern; that migration is pre-existing debt.

Happy to re-review as soon as the four items are pushed. They are maybe fifteen lines total.

rian-dolphin:
Thank you for the review. I made the four changes. The changes are on the branch now.

**1. Discoverability.** I added the row to the `reviewShortcuts` Actions section in `KeyboardShortcuts.tsx`. The row comes after the `Mod+Enter` row. This position agrees with the `displayOrder` value in the registry.

**2. Confirmation.** I added `toast.success('Feedback copied')` to the handler. I also added `toast.error('Failed to copy')` to the failure branch. The handler sets a `'Failed to copy'` state, but no component shows that state. The button tests only for `'Feedback copied!'`. Therefore a failed copy is silent in all modes. PR #1174 made the clipboard helper safe for this condition. Tell me if you want only the four items in this PR. Then I will remove the error toast.

**3. Alt key.** I added `!e.altKey` to the guard.

**4. Typing guard.** I replaced the tag test with `isTypingTarget`. The guard now has the same structure as the `Mod+Shift+G` guard.

I also corrected the comment above the handler. The comment said that the shortcut mirrors the Copy Feedback button in the review sidebar. This is not correct. The sidebar has an independent `handleQuickCopy` function. The sidebar does not call `handleCopyFeedback`.

I put the branch on the current main branch with a rebase. I did not merge main into the branch. PR #1174 changed `handleCopyFeedback`. Therefore the toast change must go on the new `copyTextToClipboard` form. This method prevents a conflict in that part of the file.

I did not change the binding for the Firefox condition on Linux. Tell me if you want a different letter.

One more note: in standalone mode the app now shows the check mark and the toast for the same operation. A test of `!origin` can prevent the toast. But you asked for a confirmation in all modes. Therefore I kept the toast.

backnotprop:
## Delta re-review (at `b02a62d3`)

Verdict: **merge as-is**. All four findings fixed correctly, verified independently, and the rebase onto current main is clean (0 behind, exactly two commits, mergeable) with the toast landing directly on the post-#1174 boolean-returning clipboard helper, gated on its result rather than fired unconditionally.

Per-item: the in-app Settings row sits exactly where the registry displayOrder puts it; the alt-key guard now matches the declared binding (AltGr resolved); the shared typing guard covers contentEditable and mirrors the sibling handler; the comment correction is accurate (the sidebar's copy is genuinely an independent path).

Beyond the asks, three things done better than asked: your observation that the `Failed to copy` state renders nowhere is independently confirmed (its only read site tests the success string, so failures were silent in every mode including the button, a pre-existing defect outside the review's findings) and the error toast now covers it universally; the rebase-over-merge choice was made for the precise right reason; and the unprompted hint text on the Settings row ("Copies the same feedback that gets submitted") resolves a real ambiguity. The double-confirmation in standalone mode is judged acceptable as-is: the redundancy is confined to the demo surface, and gating the toast would weaken the new invariant that every copy attempt reports its outcome. If dedupe is ever wanted, the cleaner path is retiring the `copyFeedback` state entirely, as its failure write is dead and its one live read is the demo checkmark; separate cleanup, not this PR.

Full suite 2793 pass / 0 fail, shortcuts tests green, review build and marketing build green (the docs page emits the entry), no em dashes. One consistency wrinkle noted for a future pass, not a regression: the sidebar's own copy button still confirms via inline checkmark with no toast, so the two copy paths now differ in confirmation style.

The Firefox-on-Linux binding overlap remains the maintainer's call and is unchanged. CI on the new head needs the maintainer workflow approval (first-contribution gate re-armed by the force-push); approving now.

PR #1164, opened 2026-07-31
fix(review): surface partial GitLab comment submissions

backnotprop:
## Summary

- Make GitLab multi-request review submissions distinguish complete, failed-before-mutation, and partial outcomes.
- Keep the submission dialog open for partial results and show the exact failed comments, errors, recovery-file path, and copy action.
- Retry only the server-authorized unposted subset or failed approval so already-posted notes and discussions are not duplicated.
- Preserve the same API behavior in the Bun and Pi review servers.

## Before and after

Before, GitLab inline discussions were posted independently. Mixed success was logged and failed comments were persisted, but the server returned ordinary success and the UI could report the review as complete.

After, successful requests return submission status complete. If GitLab accepts any mutation but rejects another, the server returns an actionable partial result containing the exact failures and safe retry payload. A total failure before any mutation remains an error, so replaying the original request is safe.

## Retry and idempotency semantics

- A partial retry omits the general review body.
- It contains exactly the failed inline comments and uses comment or approve according to the unfinished operation.
- Client response validation rejects malformed retry payloads that include already-posted comments.
- Failed comments continue to be persisted under the Plannotator data directory when possible, and the UI exposes the recovery path plus a copy action.
- If a retry attempt fails before making another mutation, the previous narrowed retry metadata remains available.

A process or network loss after GitLab accepts a mutation but before Plannotator receives the response is still ambiguous. This change does not add a durable cross-process idempotency ledger for that case.

## Runtime parity

The shared provider contract is returned unchanged through both the Bun and Pi server implementations. Focused route tests cover the Bun complete, failed, and mixed responses and the Pi mixed-response contract.

## Validation

- Focused provider, Bun/Pi route, response-parser, and retry-construction tests: 37 passed
- Opt-in rendered submission-dialog tests: 3 passed
- Root TypeScript typecheck: passed
- Review production build: passed
- Hook production build: passed
- Pi extension production build and vendoring: passed
- Branch diff check against current origin/main: passed

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,6 +53,7 @@ jobs:
           packages/ui/components/MarkdownEditor.extensions.test.tsx
           packages/ui/components/sidebar/FileBrowser.test.ts
           packages/editor/editableDocumentsHook.test.tsx
+          packages/review-editor/components/ReviewSubmissionDialog.ui.test.tsx
           packages/review-editor/hooks/useReviewSearch.test.tsx
           packages/ui/components/AnnotationPanel.props.test.tsx
           packages/ui/components/Viewer.consumer.test.tsx
--- a/apps/pi-extension/server/pr.ts
+++ b/apps/pi-extension/server/pr.ts
@@ -9,6 +9,7 @@ import {
 	type PRMetadata,
 	type PRRef,
 	type PRReviewFileComment,
+	type PRReviewSubmissionResult,
 	type PRRuntime,
 	type PRStackTree,
 	type PRListItem,
@@ -85,13 +86,14 @@ export function fetchPRContext(ref: PRRef) {
 export function fetchPRFileContent(ref: PRRef, sha: string, filePath: string) {
 	return fetchPRFileContentCore(prRuntime, ref, sha, filePath);
 }
+/** Submit a review through the Pi Node.js command runtime. */
 export function submitPRReview(
 	ref: PRRef,
 	headSha: string,
 	action: "approve" | "comment",
 	body: string,
 	fileComments: PRReviewFileComment[],
-) {
+): Promise<PRReviewSubmissionResult> {
 	return submitPRReviewCore(
 		prRuntime,
 		ref,
--- a/apps/pi-extension/server/serverReview.ts
+++ b/apps/pi-extension/server/serverReview.ts
@@ -265,6 +265,8 @@ export async function startReviewServer(options: {
 	shareBaseUrl?: string;
 	pasteApiUrl?: string;
 	prMetadata?: PRMetadata;
+	/** Platform review writer override used by isolated runtime tests. */
+	prReviewSubmitter?: typeof submitPRReview;
 	/**
 	 * The initial layer patch is missing per-file content (platform APIs
 	 * withhold patches on very large PRs). Enables the local recompute upgrade
@@ -284,6 +286,7 @@ export async function startReviewServer(options: {
 }): Promise<ReviewServerResult> {
 	const gitUser = detectGitUser();
 	const aiEnabled = resolveAIEnabled();
+	const submitPlatformReview = options.prReviewSubmitter ?? submitPRReview;
 	let draftKey = contentHash(options.rawPatch);
 	let prMeta = options.prMetadata;
 	const isPRMode = !!prMeta;
@@ -2229,16 +2232,16 @@ export async function startReviewServer(options: {
 				}
 
 				console.error(`[pr-action] ${body.action} with ${fileComments.length} file comment(s), target=${targetUrl}, headSha=${targetHeadSha}`);
-				await submitPRReview(
+				const submission = await submitPlatformReview(
 					targetRef,
 					targetHeadSha,
 					body.action as "approve" | "comment",
 					body.body as string,
 					fileComments,
 				);
-				console.error(`[pr-action] Success`);
+				console.error(`[pr-action] ${submission.status === "complete" ? "Success" : "Partial success"}`);
 				prContextLive.refreshAfterWrite(targetUrl, targetRef);
-				json(res, { ok: true, prUrl: targetUrl });
+				json(res, { ok: true, prUrl: targetUrl, submission });
 			} catch (err) {
 				const message = err instanceof Error ? err.message : "Failed to submit PR review";
 				console.error(`[pr-action] Failed: ${message}`);
--- a/apps/pi-extension/serverReview-pr-action.test.ts
+++ b/apps/pi-extension/serverReview-pr-action.test.ts
@@ -1,0 +1,109 @@
+import { afterEach, describe, expect, test } from 'bun:test';
+import { mkdtempSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import type {
+  PRMetadata,
+  PRReviewFileComment,
+  PRReviewSubmissionResult,
+} from './generated/pr-types.ts';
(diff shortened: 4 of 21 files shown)


backnotprop:
Independent review complete. Core design is sound and it fixes a real pre-existing bug worth calling out: the old code threw when all file comments failed even though the review body had already posted, so a retry duplicated the general note; that is now guarded. Verified clean: two-runtime parity is semantically exact between packages/server/review.ts and the pi mirror (same seam, same response shape, same status codes and log strings); GitHub path unchanged (never returns partial); packaging is a non-issue (pr-types.ts already vendored and allow-listed, byte-diffed generated copies match); a partial can never read as success (openUrls, setSubmitted, dialog close, and the agent feedback message all gate on allOk, and the annotation draft survives); retry dedupe holds within a dialog session including the anti-broad-retry rejection; no credential exposure in glab argv or rendered errors. Full suite 2607 pass, typecheck clean, pack --dry-run correct.

Three should-fixes before merge:

1. **The PR's own three UI tests never run in CI.** ReviewSubmissionDialog.ui.test.tsx is skipIf(!hasDom), and the DOM_TESTS step in .github/workflows/test.yml enumerates files explicitly without it. Confirmed: plain bun test = 3 skip; DOM_TESTS=1 = 3 pass. Add the file to the scoped DOM list.

2. **Cancel-after-partial invites a full double-post.** onCancel (App.tsx:3845) discards the partial state, which lives only in the dialog plan. Land 5 of 7, click Cancel, click Post Comments again: the rebuilt plan re-posts all 7, creating 5 duplicate GitLab discussions. Same on page refresh. Keep partial keyed by prUrl in App state, or at minimum relabel Cancel with a warning and guard reopen.

3. **General-comment edits are silently dropped on retry.** buildPRActionRequest forces body:'' on retry while the textarea stays editable, so a note typed after "partially posted" vanishes on Retry Unposted with no feedback. Disable the textarea or annotate it when a partial is active.

Nits, take or leave: spawn-level rejections in the diff_refs and approve calls escape the partial contract (unspawnable glab after the note posted → 500 → full replay; wrap in try/catch and route to partial); an approval-only retry that fails again throws instead of returning partial, so the same failure reports two statuses across attempts; postedFileCommentCount and reviewBodyPosted are parsed but never rendered (and the count is retry-relative, a trap for naive display); a strict-parse failure on ok:true yields failed-without-partial and a full replay under version skew; reviewBodyPosted:false conflates "no body requested" with "body not posted"; pi has 1 pr-action test vs the Bun server's 3.

Test quality note: pr-gitlab.test.ts genuinely pins the loop semantics including the recovery file contents and the throw-vs-partial split. Missing: a retry-of-a-retry staying narrowed, and an end-to-end retry-dedupe through handlePlatformAction.

backnotprop:
Applied the full owner review in commits ba9e05df and 3fa69d33.

## Required findings

1. CI DOM coverage
   - Added packages/review-editor/components/ReviewSubmissionDialog.ui.test.tsx to the exact scoped DOM_TESTS list in .github/workflows/test.yml.
   - Ran that exact CI command: 145 passed across 25 files, including all three ReviewSubmissionDialog UI tests.

2. Cancel, reopen, refresh, and retry dedupe
   - Added parsed tab-scoped recovery keyed by the root PR URL. It retains partial retry contracts, already-complete sibling targets, the original action, and the original general comment.
   - Closing and reopening the dialog or refreshing the same tab restores the exact server-authorized subset. Corrupt or version-skewed recovery fails closed as Retry blocked rather than broadening.
   - Network loss and unrecognized successful responses now also become a persisted blocked state because the remote mutation result is ambiguous.
   - The production multi-target orchestration used by handlePlatformAction is now a testable service. Its regression test covers partial, close/reopen, skipping an already-complete sibling target, retrying only two failed comments, receiving a second partial, and retrying only the final failed comment.

3. General-comment edits on retry
   - The general-comment textarea is disabled whenever a partial or ambiguous result exists.
   - The dialog explains that the body may already be posted and is excluded from safe retry. UI coverage asserts the disabled editor and guidance.

## Correctness nits

- Diff-ref spawn rejection after a posted note now returns partial with every inline comment preserved for retry; the same rejection before any mutation remains an ordinary safe failure.
- Approval spawn and nonzero-exit failures always return partial, including repeated approval-only retries, so status no longer changes between attempts.
- A server-confirmed retry failure keeps status partial and retains the prior narrowed plan.
- A malformed partial target cannot build a broad request.
- postedFileCommentCount is now rendered explicitly as This attempt, avoiding cumulative-count ambiguity. reviewBodyPosted is rendered only when true.
- Pi route coverage now matches Bun for complete, failed, and mixed outcomes.
- The client parser still requires retry comments to exactly equal the failed subset.

## Validation

- Full suite: 2,617 passed, 198 expected DOM skips, 0 failed
- Focused final suite: 47 passed, 0 failed
- Exact CI DOM command: 145 passed, 0 failed
- Root typecheck: passed
- Review, hook, and Pi production builds: passed
- Complete PR and amendment diff checks: passed

## Intentionally deferred

Durable recovery after closing the browser tab or losing the Plannotator process remains outside this PR. The tab record survives refresh, while confirmed failed comments still have the existing local JSON recovery file. An ambiguous network/process loss now blocks automatic replay and requires platform inspection, but there is still no durable cross-process idempotency ledger. reviewBodyPosted remains attempt-scoped telemetry rather than a three-state field; no retry decision relies on it because the exact retry object is the sole authority.

backnotprop:
Re-verified 3fa69d33 finding by finding. All three should-fixes and all four nits are genuinely fixed, and every claim in the point-by-point response is backed by the diff, including the self-correcting copy fix about refresh-vs-tab-close persistence. Highlights of what checked out:

- The CI DOM list now includes the UI test file: 19/19 pass under DOM_TESTS=1, zero skips.
- The recovery design is solid: sessionStorage keyed by root PR URL, restore merges the retained partial onto a fresh target so the narrowed retry survives close/reopen and refresh, cleared only on full success. I tried to construct a surviving double-post and could not: the ambiguous outcomes (network rejection, unparseable 2xx, unrecognized payload) all land in the new blocked state which is enforced at all four submission entry points, and corrupt storage fails closed. Bonus: it also fixes a pre-existing bug where a successful sibling target was re-posted after close/reopen.
- The retry-relative count trap was handled exactly right ("This attempt posted N...").
- Two-runtime law: no new server logic in either runtime; the only shared change is pr-gitlab.ts, vendored byte-identically. Full suite 2617 pass, typecheck clean, pack --dry-run correct.
- The new platformReviewSubmission test drives partial -> close -> rebuild -> restore -> retry -> second partial -> retry and pins the exact request sequence, including that the edited body is never resent.

One should-fix worth a fast-follow issue rather than blocking: **a blocked target has no UI escape**. Nothing clears the recovery outside a fully successful attempt, so one transient fetch rejection permanently disables platform submission for the tab (Confirm disabled, reopen reloads the block from sessionStorage), and the only escape (closing the tab) is exactly the path that discards the narrowing. Fail-closed is right, but it needs an explicit "I checked the MR, start over" affordance; same applies to the corrupt-recovery block and any future storage version bump.

Small residuals, none blocking: the step-1 note POST is the one write still outside the spawn-rejection guard (500 there implies full replay, low risk as the first mutation); recovery is origin-scoped so a server restart (new port) loses it while the draft survives (worth a line in the retry copy); openPlatformDialog overrides a requested Approve with the recovered comment action without explanation; a recovery with only success targets shows the stale general comment in an editable box; the Dialog:154 invariant throw gets miscategorized as ambiguous; and reviewSubmissionRecovery.ts deserves its own direct test file.

Approving.

PR #1168, opened 2026-07-31
feat(ui): render markdown reference links

### LINKED ISSUE #923: Markdown Reference Links should be supported
My markdown reference link is rendering as plaintext instead of generating a link.

Reference links are formatted close to normal markdown links, but slightly different. Instead of a `[link](https://....)` a '[reference link][id of link]` uses `[ ]` for the link portion instead of `( )`.

Reference links are defined in a file through the following syntax:

```
[id of link]: https://github.com
```

rNoz:
## Summary

Closes #923.

The simplified markdown parser (`parseMarkdownToBlocks`) only understood inline links `[text](url)`. CommonMark reference links rendered as raw text: both the reference `[text][id]` and its `[id]: url` definition showed literally in the document. This adds reference-link support by resolving references into inline links before the block split, so the existing inline renderer draws them.

## Approach

A single pure function, `resolveReferenceLinks(markdown)`, runs at the top of `parseMarkdownToBlocks`, right after frontmatter extraction. It does two passes:

1. Collect link reference definitions (`[label]: destination "optional title"`, up to three leading spaces, `<url>` or bare destination, first definition wins, labels normalized case-insensitively with collapsed internal whitespace, `^`-prefixed labels rejected so GFM footnotes are left alone). A line only counts as a definition when it can start a block (the previous line is blank, a code fence, another definition, or the document start) and its trailing content is empty or a valid quoted or parenthesized title, so ordinary prose is never mistaken for a definition and a definition can never interrupt a paragraph.
2. Rewrite references into inline links: full `[text][id]`, collapsed `[text][]`, and shortcut `[text]`, plus the image forms `![alt][id]`. Only definitions that are actually referenced somewhere are blanked; an unreferenced definition (or one whose only reference sits inside a protected region) is left as-is. Blanked lines are replaced with an empty line, not removed, so line count, and therefore every block's `startLine`, stays accurate.

Fenced code blocks and HTML blocks share the exact block-boundary detection used by the rest of the parser, so a reference-looking pattern inside either is never rewritten and a `[id]: url` inside either is never collected as a definition. Because the output is ordinary `[text](url)` markdown, no renderer change is needed.

## Correctness and safety

- Code-aware and HTML-aware: content inside fenced code blocks, inline code spans, and raw HTML blocks (`<pre>`, `<details>`, and similar) is never rewritten, and a reference-shaped line inside any of them is never collected as a definition.
- GFM footnote definitions (`[^1]: ...`) are left untouched, not consumed as reference definitions.
- A shortcut `[text]` is left literal when an inline `(...)` destination follows it and when it does not name a definition, so bracketed prose like `[TODO]` or an array index `[0]` never becomes a false link. A checkbox marker (`- [x]`) at a list-item start is never treated as a shortcut reference.
- An unknown full or collapsed reference stays literal, matching CommonMark.
- The document's line endings (LF or CRLF) do not affect resolution.
- Scanning is linear in document length, so an adversarial or very large document (for example a converted web page) does not cause a slowdown, and a shared linear close index between the reference resolver and the pre-existing HTML-block parser means a long but well-formed HTML block (for example a large `<details>` section) still renders as a single block rather than being truncated for performance.
- A document that defines no references, or none that are actually referenced, is returned unchanged (fast path), so the common case is untouched.

## Validation

226 targeted parser tests pass, including full/collapsed/shortcut/image resolution, case- and whitespace-insensitive labels, the angle-bracket destination, first-definition-wins, unknown references left literal, protected fenced code, inline code, HTML blocks and footnotes, CRLF handling, consumed-only definition blanking, and linear-time scanning on large documents including long well-formed HTML blocks. A full UI and typecheck pass (including the strict consumer) is also clean. No `node:` imports were added, so the browser-safe boundary holds.

## Known limitations

- An unbalanced `)` inside a destination can still truncate the resulting href; a nested reference like `[see [1]][id]` can still produce an extra link (its href is still sanitized); and a definition placed directly under a heading with no blank line in between is not recognized, matching the parser's existing prose-continuation handling.
- Migration note: a draft annotation or share link created before this change, whose highlighted text spans raw `[text][id]` reference markup, will not auto-restore its position after upgrading; the annotation is preserved but appears unanchored, with a console warning. This is a one-time transition case for pre-existing data, not an ongoing limitation.

## Compatibility

Additive. Documents without reference definitions are unchanged (identical block output). No public API, prop, or payload changes. The function is exported so it can be unit-tested and reused.

## Related work

None. This is a self-contained parser feature.

### DIFF
--- a/packages/ui/HANDOFF.md
+++ b/packages/ui/HANDOFF.md
@@ -267,6 +267,8 @@ interface Annotation {
 
 **Honesty note:** the failure path (step 4) is exercised in real use but is **not covered by automated tests** — nothing in the suite asserts the stale-anchor behavior. Treat the described degradation as accurate-but-unverified-by-CI, and test it in your integration if you depend on it.
 
+**Migration caveat — reference-style link resolution (#923):** `parseMarkdownToBlocks` now rewrites CommonMark reference links (`[text][id]`) and blanks their `[id]: url` definitions before splitting into blocks, so documents containing that syntax render differently than they did before this pass existed — a `[text][id]` pair that used to render as literal bracket text now renders as a link, and the definition line disappears from the rendered DOM entirely. That changes both the text and the per-tag DOM index at the affected positions. Any annotation whose `startMeta`/`endMeta` was captured against the *old* (pre-resolution) render of such a document — i.e. persisted before a host upgrades past this change — can restore onto the wrong text after upgrading, same as any other DOM-structure change described above; the text-search fallback (step 3) is the recovery path, and `originalText` is what to fall back to if you need to re-anchor server-side.
+
 ---
 
 ## Known rough edges (and why they're fine for now)
--- a/packages/ui/utils/parser.test.ts
+++ b/packages/ui/utils/parser.test.ts
@@ -1220,6 +1220,76 @@ describe("parseMarkdownToBlocks — raw HTML blocks", () => {
   });
 });
 
+describe("parseMarkdownToBlocks / resolveReferenceLinks — unclosed-HTML-opener perf (PR #1168 follow-up)", () => {
+  // Root cause: the balanced open/close depth scan for a multi-line HTML
+  // block does not advance the outer index when it fails to find a closing
+  // tag — so a document with many consecutive unclosed openers (e.g.
+  // thousands of bare `<div>` lines with no `</div>` anywhere) makes EVERY
+  // one of them independently re-scan all the way to end-of-document. That
+  // is O(N^2) work for N such lines, a real DoS-shaped hazard well within
+  // the 2MB annotate cap. Both markProtectedLines (used by
+  // resolveReferenceLinks) and parseMarkdownToBlocks's own HTML-block
+  // section duplicate this exact scan, so both must be fixed.
+  const N = 8000;
+  // Generous bound: a linear/bounded fix finishes in well under 100ms for
+  // this input; the pre-fix O(N^2) scan takes multiple seconds (measured
+  // ~2.2s for N=8000 during triage). 800ms leaves large machine-variance
+  // headroom while still failing clearly against the quadratic behavior.
+  const TIME_BOUND_MS = 800;
+
+  test("parseMarkdownToBlocks stays fast with many consecutive unclosed <div> openers", () => {
+    const md = Array.from({ length: N }, () => "<div>").join("\n");
+    const start = performance.now();
+    const blocks = parseMarkdownToBlocks(md);
+    const elapsed = performance.now() - start;
+    expect(blocks).toHaveLength(N);
+    expect(blocks.every((b) => b.type === "html")).toBe(true);
+    expect(elapsed).toBeLessThan(TIME_BOUND_MS);
+  });
+
+  test("resolveReferenceLinks stays fast with many consecutive unclosed <div> openers", () => {
+    const md =
+      Array.from({ length: N }, () => "<div>").join("\n") +
+      "\n\n[x][id]\n\n[id]: https://e.com";
+    const start = performance.now();
+    const result = resolveReferenceLinks(md);
+    const elapsed = performance.now() - start;
+    expect(result).toContain("[x](https://e.com)");
+    expect(elapsed).toBeLessThan(TIME_BOUND_MS);
+  });
+
+  test("parity: a real <details>...</details> block stays intact and identically protected/parsed among thousands of unclosed <div> decoys", () => {
+    const decoyCount = 5000;
+    const decoys = Array.from({ length: decoyCount }, () => "<div>").join("\n");
+    const md =
+      `${decoys}\n\n` +
+      "<details>\n<summary>Notes</summary>\n\n[id]: https://from-details.com\n\n</details>" +
+      "\n\nAfter.";
+
+    const start = performance.now();
+    const blocks = parseMarkdownToBlocks(md);
+    const resolved = resolveReferenceLinks(md);
+    const elapsed = performance.now() - start;
+    expect(elapsed).toBeLessThan(TIME_BOUND_MS);
+
+    // parseMarkdownToBlocks: the real <details> block is captured whole,
+    // undisturbed by the decoys before it, followed by its own paragraph.
+    const detailsBlock = blocks.find((b) => b.type === "html" && b.content.startsWith("<details>"));
+    expect(detailsBlock).toBeDefined();
+    expect(detailsBlock!.content).toBe(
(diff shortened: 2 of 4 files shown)


backnotprop:
In-depth review done. First: the core design call is exactly right. Blanking definition lines in place instead of deleting them preserves startLine, which the checkbox reporter and section grouping depend on, and that was verified downstream. Security also checks out end to end: resolved URLs go through sanitizeLinkUrl (javascript:/data:/vbscript:/file: blocked, verified), hrefs travel as React props so attribute smuggling is impossible, and a reference definition carries exactly the trust of an inline link. Idempotence confirmed. Full suite and typecheck clean, 142 new tests pass.

That said, several content-corruption paths are reachable with ordinary plan markdown, so this needs a round of fixes:

1. **Blocking: indented fences are rewritten.** markFencedLines strips only up to 3 leading spaces before testing for a fence, but the block loop uses line.trim(), so every fence nested under a list item (4+ spaces, extremely common in plans) is code to the renderer but unprotected in the resolver. Verified: `const v = ref[a][b];` inside a fence under a numbered step gets a URL substituted into the highlighted code block. Fix: share the block loop's trim-based fence detection.

2. **HTML blocks are unprotected.** Content inside <pre> gets rewritten, and a definition line inside <details> gets blanked, deleting a line out of the HTML block body.

3. **GFM footnotes are eaten.** The definition regex accepts ^ in labels, so [^1]: https://source turns Claim[^1] into a link labelled ^1 and deletes the footnote. One-line fix: reject labels starting with ^.

4. **Definitions are blanked unconditionally, even when nothing consumes them.** "[owner]: alice" under an Owners heading vanishes from the rendered plan; a definition whose only reference sits inside a fence is deleted while the reference stays literal. Suggested fix that also softens 2, 3, and 5: track which labels resolveRefsInLine actually consumed and blank only those definition lines.

5. **Plan diff (default clean mode) can show an empty modified block.** MarkdownChunk parses each diff chunk independently while computePlanDiff diffs raw markdown, so a version change that only edits a definition URL renders as a +2/-2 badge with an empty block; the user sees nothing changed. The consumed-labels fix above makes this much rarer; acceptable as a documented follow-up if you prefer.

6. **CRLF documents silently get nothing.** The definition regex ends [ \t]*$ so a trailing \r kills the match; the rest of the parser is CRLF-tolerant. Windows-authored annotate files never see the feature.

7. **Quadratic backtracking guard needed.** A long unterminated bracket run is O(n squared): measured 6.3s at 128k chars, extrapolating to tens of minutes at the 2MB annotate cap, reachable via annotate on a hostile or Jina-converted page. Cheap guard: skip lines with no ], or bound the label length in the regex. Realistic docs are fine (+12ms on a 1.6MB doc with refs).

Nits, take or leave: the resolver knows ~~~ fences but the block loop does not, so a stray ~~~~ silently disables the feature for the rest of the document; destinations are not escaped when re-inlined so a ) in a URL truncates the href and leaks stray text; nested brackets ([see [1]][id]) produce a spurious shortcut link; backslash-escaped brackets still resolve; definitions directly under a heading with no blank line are not recognized (degrades safely, but a ## References section written tightly will not work).

One release-notes item rather than a code change: drafts and share URLs created before this PR whose anchor text spans raw [text][id] markup will fail text-search restore after upgrade (annotation survives unanchored with a console warning). Narrow and one-time.

The new tests are good on everything the implementation considered; the fix round should add cases for indented fences, HTML blocks, footnotes, unreferenced definitions, CRLF, and an explicit idempotence assertion. Happy to re-review quickly on push.

rNoz:
Pushed the review fixes. Protected regions, footnotes, unused definitions, CRLF, bounded scanning, and clean diff visibility are covered. Checks are passing.

backnotprop:
Re-verified 9440be06 by re-running every original reproduction plus a 20,000-document differential fuzz (zero protected-region leaks, zero line-count changes, zero idempotence failures). All seven findings are genuinely fixed, and two of your choices are better than what the review suggested: sharing the block parser's exact conditions (single findHtmlBlockEnd used by both passes, byte-identical void-tag branches) and blanking only consumed labels, which also fixed the clean-diff invisibility with no diff-engine change needed. The quadratic scan is now clean linear: 128k chars went from 6.3s to 95ms, and the 2MB worst case from an extrapolated ~27 minutes to 1.4s. CRLF, footnotes, indented fences, HTML blocks, unused definitions: all confirmed against the original repro inputs. 25 new tests, full suite 2631 pass, typecheck clean. Removing ~~~ protection to agree with the block parser was also the right call.

One new regression from the perf fix, and it is the only thing to hold for: **MAX_HTML_BLOCK_SCAN_LINES = 2000 now applies to the pre-existing parseMarkdownToBlocks HTML scan.** A <details> (or raw <table>) whose closing tag sits more than ~2000 lines below the opener no longer renders as one block: measured, a 1999-line body splits into opener-alone, body-as-paragraphs, close-alone, so a plan wrapping a long log in <details> loses its collapsible and leaks raw HTML as text. Mitigating facts, both measured: the same commit fixes a pre-existing O(N squared) hang (40k unclosed divs: 40.5s on main, 26ms here), and the closeExistsFromLine suffix array already handles the no-close-anywhere pathological shape without any cap. The cap only guards the rare close-exists-far-away case, which is still slow at 2000 anyway. Suggested: raise it substantially or scale with document length, e.g. Math.max(20000, lines.length).

Residual nits, all acceptable as-is: unbalanced ) in a destination still truncates the href (your new test covers only the balanced case); mixed backslash-escape [not a ref\][id] still resolves; [see [1]][id] still yields a spurious link (explicitly documented and sanitization-guard-tested, so the invariant that matters holds); tight-heading definitions still degrade to old behavior; the clean-diff view is now consistently source-level, which disagrees with the rendered view but defensibly so. The HANDOFF note on pre-upgrade draft re-anchoring covers the migration caveat.

Bump the cap and this is ready.

rNoz:
Pushed the final HTML-block fix. Long valid blocks are preserved, while both unclosed-opener adversarial shapes now use a shared linear close index. Checks are passing. If you want a tiny palate cleanser after this parser one, #1154 is a small Vim HUD scroll fix.

backnotprop:
Verified fe017f1c empirically. The prefix-sum close index is a better answer than the raised cap the review suggested: long valid details blocks at 1999, 5000, and 30000 lines all render as a single HTML block (regression gone), 40k unclosed divs parse in 19ms, and the residual adversarial shape that previously cost 8.7s even under the cap (100k openers plus one trailing close) now parses in 51ms. Resolver idempotence holds on protected content, and all 176 parser tests pass. Nothing left from my side: merge-ready.

PR #1171, opened 2026-07-31
fix(annotate): enforce archive read-only surfaces

backnotprop:
## Summary
- make standalone archive document surfaces read-only across Markdown, raw HTML, annotation panels, paste uploads, drafts, code annotations, and submit shortcuts
- reject archive document mutations in both Bun and Pi plan servers
- use the plan area border-box metric for both initial and ResizeObserver action-label bucketing
- cover archive, writable annotate, raw-HTML selection, server parity, and responsive thresholds with regressions

## Verification
- bun test (2631 passed, 203 skipped, 0 failed)
- DOM_TESTS=1 bun test packages/editor/App.archiveReadOnly.test.tsx packages/editor/actionsLabelMode.test.ts packages/ui/components/Viewer.consumer.test.tsx packages/ui/components/AnnotationPanel.props.test.tsx (31 passed)
- bun test packages/server/api-404-guard.test.ts (9 passed)
- bun run typecheck
- bun run --cwd apps/review build
- bun run build:hook

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,6 +69,8 @@ jobs:
           packages/ui/hooks/useLinkedDoc.test.tsx
           packages/ui/components/DocBadges.test.tsx
           packages/editor/planDiffAutoExit.test.tsx
+          packages/editor/App.archiveReadOnly.test.tsx
+          packages/editor/actionsLabelMode.test.ts
 
   pi-extension-ai-runtime-windows:
     # Exercises the Pi extension's Node/jiti server mirror on Windows with an
--- a/apps/pi-extension/server/serverPlan.ts
+++ b/apps/pi-extension/server/serverPlan.ts
@@ -53,6 +53,7 @@ import {
 } from "./reference.ts";
 import { handleFileBrowserStreamRequest } from "./file-browser-watch.ts";
 import { warmFileListCache } from "../generated/resolve-file.ts";
+import { isArchiveDocumentMutation } from "../generated/archive-mode.ts";
 
 export interface PlanReviewDecision {
 	approved: boolean;
@@ -169,6 +170,11 @@ export async function startPlanReviewServer(options: {
 		if (url.pathname === "/api/done" && req.method === "POST") {
 			resolveDone?.();
 			json(res, { ok: true });
+		} else if (
+			options.mode === "archive" &&
+			isArchiveDocumentMutation(req.method ?? "GET", url.pathname)
+		) {
+			json(res, { error: "Archive is read-only" }, 403);
 		} else if (url.pathname === "/api/archive/plans" && req.method === "GET") {
 			const customPath = url.searchParams.get("customPath") || undefined;
 			if (!cachedArchivePlans)
--- a/apps/pi-extension/vendor.sh
+++ b/apps/pi-extension/vendor.sh
@@ -29,7 +29,7 @@ for f in config-types storage-types workspace-status-types; do
 done
 
 # Everything else in the original flat list stays sourced from packages/shared.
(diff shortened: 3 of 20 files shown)


backnotprop:
## Final review (adversarial, at `bb5d0e08`)

Typecheck clean; full suite 2631 pass / 0 fail; targeted DOM run of the four touched/new test files 32 pass. Verdict: **merge as-is**. No blocking or should-fix findings. Everything attacked held up.

### Server enforcement: complete and drift-free
Ran both runtimes in archive mode and probed every mutating route directly. Byte-identical rejection surface across Bun and Pi: approve / deny / save-notes / upload / draft-delete all 403 `{"error":"Archive is read-only"}`, external-annotations and editor-annotation 404, ai/session 503. A 3 MB unconsumed-body POST to Pi's `/api/upload` still returns a clean 403 with no hang. The check is server-authoritative (`mode === "archive"` from server options), never a client flag. Coverage is exhaustive: the plan servers have no `/api/feedback`, `/api/exit`, or `/api/source/save` at all, and the annotation/AI runtimes are already nulled in archive in both runtimes.

### Overlay leak, both directions: clean
`documentReadOnly` derives only from `useArchive.init()`, which runs only when `/api/plan` returns `mode:"archive"`. In-session archive browsing takes the linked-doc branch with `archiveMode` false, so a live session cannot be bricked; there is no toggle that can stick. Conversely, in standalone archive a linked doc opened from an archived plan inherits `readOnly` on both Viewer and HtmlViewer. All input paths verified covered: selection highlighter, vim mode, pinpoint, code-block handlers, image paste, code-file popout, notes menu, submit shortcuts.

### Drafts: clean
Instrumented the new test's fetch recorder: a standalone archive mount issues only `GET /api/plan` and `GET /api/archive/plans`, zero `/api/draft` traffic. No draft creation, no tombstone DELETE. Keying cannot collide with a live review either.

### actionsLabelMode: clean
The bug was real and large: the plan area carries `px-2 md:px-10 xl:px-16` padding, so the old `contentRect.width` was 16 to 128 px narrower than the initial `getBoundingClientRect()`. Both paths now read the border box with a rect fallback, entries are target-filtered, and the setState is a no-op on equal values. The test is not vacuous: it feeds divergent contentRect vs borderBoxSize and asserts the border-box value wins, plus the fallback path and the `{box:"border-box"}` observe option.

### Shared module / vendoring / CI: clean
`archive-mode.ts` is zero-dep, correctly placed in `shared` (server-only, not `core`); the `./archive-mode` export entry is present; `vendor.sh` covers it and `npm pack --dry-run` includes `generated/archive-mode.ts` in the Pi tarball with the exact-specifier import typechecking. Both new test files are registered in the `DOM_TESTS=1` CI step, which the DOM-gated App test genuinely needs.

### Nits (fine as follow-ups, none merge-relevant)
- During a live review, an archived plan opened via the Archive sidebar overlay remains annotatable. Pre-existing and unchanged; annotations attach to the linked doc, never the archive file. Worth one explicit line in CLAUDE.md that the overlay is intentionally out of scope.
- `AnnotationPanel`'s `readOnly` now also suppresses `renderCardFooter` and direct-edit discard. No in-repo consumer combines them, but it is a published-seam semantic change for Workspaces hosts; the JSDoc documents it, so deliberate.
- One-liner test gap: `requestedRoutes` is already captured in the archive App test, so asserting no `POST /api/draft` is cheap (verified to hold).

### Pre-existing, out of scope
- `archive.fetchPlans()` runs in the same tick as `init()`, so the auto-select-first-plan branch for `--path` never fires; harmless today because `/api/plan` seeds the plan.
- CLAUDE.md says archive mode has "sharing disabled", but the hook server passes the global `sharingEnabled` and the share path is live in the archive browser. Doc/code mismatch worth a follow-up issue.

PR #1174, opened 2026-08-01
fix(ui): fall back to legacy copy in insecure browser contexts

### LINKED ISSUE #1173: Clipboard copy actions fail when Plannotator runs in insecure browser contexts
## Problem

Several Plannotator UI copy actions call `navigator.clipboard.writeText(...)` directly. This breaks when the app is opened in an insecure browser context, such as a non-localhost HTTP URL, remote/devcontainer session, or an environment where the Clipboard API is unavailable or rejects writes.

In those cases, copy buttons either silently fail or show a generic failure even though the browser could still support the legacy `document.execCommand('copy')` path.

The error message in brower console:
```
(index):4521 Failed to copy: TypeError: Cannot read properties of undefined (reading 'writeText')
    at Ce ((index):4521:1930)
    at $T ((index):61:128930)
    at (index):61:134261
    at On ((index):61:15163)
    at Xk ((index):61:130183)
    at u0 ((index):62:28961)
    at pee ((index):62:28782)
```

## Expected behavior

Copy buttons should work consistently in supported browsers, including when `navigator.clipboard.writeText` is unavailable or rejects due to browser security restrictions.

If all copy strategies fail, the UI should report the failure without crashing.

### Plannotator version

plannotator 0.25.1

### OS

Linux

### Agent

Other

### Where did it happen?

Annotate

backnotprop:
Closes #1173

**Bug:** navigator.clipboard only exists in secure browser contexts. Remote-mode Plannotator serves plain HTTP on a non-localhost host, so navigator.clipboard is undefined there and every bare navigator.clipboard.writeText(...) call threw TypeError: Cannot read properties of undefined (reading 'writeText'). Reported from annotate mode on 0.25.1; it affected every copy button. Remote-mode HTTP is the main affected environment.

**Fix:** New copyTextToClipboard(text): Promise<boolean> in packages/ui/utils/clipboard.ts. It tries the async Clipboard API (guarded against synchronous throws), falls back to the existing copy-event / execCommand path, returns whether the copy actually happened, and never throws. All 27 bare call sites across packages/ui, packages/editor, and packages/review-editor now go through it, preserving each site's UX (Copied states only on success, existing error toasts and console errors on failure, fire-and-forget sites stay fire-and-forget). GoalSetupSurface gains the fallback and keeps its error message for the all-strategies-failed case. copyTextPreservingFocus keeps its exported signature and behavior unchanged.

**Tests:** DOM-gated unit tests (packages/ui/utils/clipboard.test.ts, registered in .github/workflows/test.yml) cover: Clipboard API absent (falls back, returns the execCommand result), writeText rejecting, writeText throwing synchronously, writeText resolving (no fallback invoked), everything failing (resolves false, never throws), and the fallback copy event carrying the text payload. Typecheck, full bun test, and the CI DOM test set all pass.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,6 +48,7 @@ jobs:
           packages/ui/annotationDraftPersistence.test.tsx
           packages/ui/codeAnnotationDraftPersistence.test.tsx
           packages/ui/components/html-viewer/srcdoc.test.ts
+          packages/ui/utils/clipboard.test.ts
           packages/ui/components/InlineMarkdown.resolveLinkedDoc.test.tsx
           packages/ui/components/MarkdownDiff.frozen.test.tsx
           packages/ui/components/MarkdownEditor.extensions.test.tsx
--- a/packages/editor/App.tsx
+++ b/packages/editor/App.tsx
@@ -4725,7 +4725,7 @@ const App: React.FC = () => {
             onClose={() => setIsPanelOpen(false)}
             onQuickCopy={async () => {
               const output = getCurrentFeedbackPayload();
-              await copyTextToClipboard(wrapCopiedFeedback(output));
+              return copyTextToClipboard(wrapCopiedFeedback(output));
             }}
             onShare={canShareCurrentSession ? () => { setIsPanelOpen(false); setInitialExportTab('share'); setShowExport(true); } : undefined}
             otherFileAnnotations={otherFileAnnotations}
--- a/packages/editor/App.tsx
+++ b/packages/editor/App.tsx
@@ -25,6 +25,7 @@ import { getCallbackConfig, CallbackAction, executeCallback } from '@plannotator
 import { useAgents } from '@plannotator/ui/hooks/useAgents';
 import { useActiveSection } from '@plannotator/ui/hooks/useActiveSection';
 import { storage } from '@plannotator/ui/utils/storage';
+import { copyTextToClipboard } from '@plannotator/ui/utils/clipboard';
 import { configStore, useConfigValue } from '@plannotator/ui/config';
 import { CompletionOverlay } from '@plannotator/ui/components/CompletionOverlay';
 import { useUpdateCheck } from '@plannotator/ui/hooks/useUpdateCheck';
@@ -3829,10 +3830,9 @@ const App: React.FC = () => {
   // (utils/agentInstructions.ts) so it's easy to edit independently of UI code.
   const handleCopyAgentInstructions = async () => {
     const payload = buildPlanAgentInstructions(window.location.origin);
-    try {
-      await navigator.clipboard.writeText(payload);
+    if (await copyTextToClipboard(payload)) {
       toast.success('Agent instructions copied');
-    } catch {
+    } else {
       toast.error('Failed to copy');
     }
   };
@@ -3845,10 +3845,9 @@ const App: React.FC = () => {
       toast.error('Failed to create share link');
       return;
     }
-    try {
-      await navigator.clipboard.writeText(url);
+    if (await copyTextToClipboard(url)) {
       toast.success('Share link copied');
-    } catch {
+    } else {
       toast.error('Failed to copy');
     }
   };
@@ -4726,7 +4725,7 @@ const App: React.FC = () => {
             onClose={() => setIsPanelOpen(false)}
             onQuickCopy={async () => {
               const output = getCurrentFeedbackPayload();
-              await navigator.clipboard.writeText(wrapCopiedFeedback(output));
+              await copyTextToClipboard(wrapCopiedFeedback(output));
             }}
             onShare={canShareCurrentSession ? () => { setIsPanelOpen(false); setInitialExportTab('share'); setShowExport(true); } : undefined}
             otherFileAnnotations={otherFileAnnotations}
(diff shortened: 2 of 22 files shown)


backnotprop:
## Adversarial review (at `45a60cba`)

Verdict: core fix is correct and the load-bearing property holds, verified empirically: with `navigator.clipboard` undefined, the fallback runs in the same synchronous turn as the click handler, so the user-gesture window is preserved on the exact path issue #1173 hits. `copyTextPreservingFocus` is byte-identical to main (diffed), so the published `@plannotator/ui` surface is untouched. All 27 conversions were diffed against main; toasts, timeouts, and state ordering match. Typecheck clean; new tests skip cleanly without DOM and pass with it; full DOM CI set 160 pass / 0 fail.

Two findings should be fixed before merge; both are small.

**1. The fallback textarea closes popout dialogs mid-copy** (`TablePopout.tsx:103,112`, `CodeFilePopout.tsx:461`). The fallback appends a textarea to `document.body` and focuses it. `PopoutDialog` closes on focus-out unless the focus target matches `ANNOTATION_SELECTORS`, and a bare body-level textarea matches none of them. In remote mode over HTTP, clicking "Copy as markdown" in a table popout copies the text and then closes the popout underneath the click, and the "Copied!" state lands on a closed dialog. PR-introduced: these sites had no textarea fallback on main. Cleanest fix: tag the fallback textarea with a marker attribute and add it to `ANNOTATION_SELECTORS`. `AnnotationToolbar` is not affected (its button sits inside `.annotation-toolbar`).

**2. Quick copy now shows "Copied" on failure** (`packages/editor/App.tsx:4728` via `AnnotationPanel.tsx:249`). `onQuickCopy` is typed `Promise<void>`, so the boolean is discarded and the panel unconditionally sets the copied state. On main a failure rejected and the checkmark never showed. Fix: widen the prop to `Promise<boolean>` (additive on the published seam) and gate the state.

Worth taking in the same pass:

**3. The fallback can report success when nothing was copied** (`clipboard.ts:12-24`, confirmed with a probe): the copy handler sets `copied = true` even when `execCommand` returns false, and when `clipboardData` is null the handler still flags success after `preventDefault()` suppressed the native copy. Inherited from main, but the boolean now drives every "Copied!" badge. One-line tighten: only set the flag when `setData` actually ran.

**4. Pin the synchronous-fallback property in a test.** None of the 7 tests fails if someone inserts an `await` before the fallback, which would silently kill the fix in real browsers while staying green. A 6-line test (call without awaiting, assert `execCommand` fired before return) covers it.

**5. Nit:** `GoalSetupSurface` now reports every failure as "Clipboard is unavailable in this browser", including permission-denied with the API present. Preserve the real error when one exists.

Noted, no action needed: share-link copy awaits a network round-trip before copying, so Safari's strict gesture window may still expire; not a regression (main threw a TypeError there) but #1173 is not fully closed for that one path. `apps/skills/extra/plannotator-compound/assets/report-template.html` has the same bug class in a generated artifact; follow-up material.

Also verified clean: no cross-listener hijack with Viewer's document copy listener (the fallback textarea is excluded by Viewer's own tag guard), zero bare `navigator.clipboard` left in packages/, clipboard reads unaffected (paste uses the native event and works in insecure contexts), no lockfile or artifact churn, tests excluded from the published package.

backnotprop:
## Delta re-review (at `9c54237a`)

Verdict: **merge as-is**. All five fixes verified mechanically, not by reading.

- **Popout focus-out:** the marker attribute is matched by both guard branches of `PopoutDialog` (target and relatedTarget run the same selector check), including the second focusout fired when the textarea is removed while focused. No legitimate close is suppressed: the only producer of the attribute appends and removes it within one synchronous function body, and a test asserts no marker element survives the copy. The `ANNOTATION_SELECTORS` export is purely additive.
- **onQuickCopy widening:** proved additive with an isolated strict tsc check in both directions (host `Promise<void>` and `Promise<boolean>` both assign; narrowing is rejected, pinned with `@ts-expect-error`). `PanelProps` is not exported, so the breaking direction is unreachable for hosts. Void resolution keeps today's flash.
- **False-success paths:** both original probes now resolve `false`, and the null-`clipboardData` case no longer suppresses the native copy. The `&&` tighten cannot discard a real success: in the odd browser where the handler ran but `execCommand` reported false, the textarea retry performs its own copy, so the worst case is a false negative, strictly the safer direction.
- **Synchronicity test:** mutation-tested. Inserting `await Promise.resolve()` before the fallback fails exactly that test and nothing else, so the property the fix rests on now has a real guard.
- **GoalSetupSurface:** all four error paths traced; real rejection errors propagate again, generic message reserved for API-absent/total failure, gesture window preserved.

Typecheck (incl. strict-consumer) pass; clipboard tests 10/10 with DOM, 10 clean skips without; full CI DOM set 163/0.

One behavior delta to record: `copyTextPreservingFocus` is no longer byte-identical to main in one narrow branch. When a copy handler ran but `execCommand` reported false, main declared success and skipped the retry; it now takes the textarea retry (a redundant second copy of the same text, focus restoration verified intact). Benign and arguably a fix, noted here since the first commit advertised unchanged behavior.

Non-blocking notes for follow-up: the marker attribute literal lives in `clipboard.ts` and `PopoutDialog.tsx` independently (tied together only by the test, which is the right layering); `GoalSetupSurface` double-invokes `writeText` on the reject path and partially duplicates the helper's strategy ladder; the AnnotationPanel seam-contract test has no `onQuickCopy` coverage (pre-existing gap).

PR #1177, opened 2026-08-01
fix(uninstall): require host cleanup before binary removal

### LINKED ISSUE #1175: Keep uninstall atomic when host integration cleanup fails
## Problem

The uninstaller should have one simple contract: remove Plannotator itself and its installed host integrations, while letting the user choose whether to retain or purge local Plannotator data.

The current `--skip-hosts` recovery path violates that contract. It can remove the binary while leaving hooks or plugins that still invoke it, creating a broken host configuration.

## Required behavior

- Normal uninstall removes the binary, installer-owned assets, and detected Plannotator host integrations while retaining local data.
- Full purge performs the same removal and additionally deletes local Plannotator data after confirmation.
- If a host integration cannot be removed safely, stop before deleting the binary.
- Identify the blocking configuration and provide exact manual cleanup instructions, then let the user rerun uninstall.
- Do not report a successful uninstall that leaves active hooks pointing to a missing binary.
- Remove or retire `--skip-hosts` as a binary-removal bypass.

## Tests

- A malformed or uneditable host configuration leaves the binary and integration intact and exits with actionable instructions.
- After manual cleanup, rerunning uninstall succeeds.
- Normal uninstall retains local data.
- Purge deletes local data.

## Context

This supersedes the earlier proposal to preserve and merely report skipped host integrations. The product decision is that host integrations are part of the installation and must be removed with it.

backnotprop:
## Summary

- retire the host-cleanup bypass from CLI parsing, types, runtime branches, help, docs, diagnostics, and tests
- always run recognized host plugin/config cleanup before binary removal
- give host-specific manual cleanup commands or config-field guidance, followed by the correct uninstall rerun command
- preserve the existing fail-safe gate that keeps the CLI and Windows PATH entry available after cleanup errors
- add regressions for malformed and uneditable host config, unavailable and failing plugin managers, manual-cleanup retry, data retention, purge, and platform safeguards

## Validation

- focused uninstall/CLI suite: 56 passed, 0 failed
- full suite: 2,670 passed, 204 skipped, 0 failed
- bun run typecheck
- bun run build:marketing
- git diff --check
- exhaustive repository and read-only Mintlify docs searches found no remaining retired-flag references

## Docs

The separate Mintlify docs checkout contains no references requiring a companion change.

Closes #1175

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,10 +236,10 @@ Purge requires typing `purge` at the prompt and explains that the data is
 local-only: it is not stored on a Plannotator server and cannot be recovered.
 For automation, pass `--yes` (or `-y`); non-interactive removal refuses to run
 without it. Use `--dry-run` to preview recognized work without making changes.
-If a broken or unavailable host blocks cleanup, `--skip-hosts` leaves host
-plugin managers and shared host configuration untouched so the remaining
-installer-owned components and binary can still be removed; clean up those
-host integrations manually afterward.
+Host integrations are always part of uninstall. If a broken or unavailable
+host prevents safe cleanup, the command names the blocking plugin manager or
+configuration, gives exact manual cleanup instructions, and stops before
+deleting the binary. Complete that cleanup and rerun uninstall.
 These mechanics keep the ordinary confirmation default-negative, make the
 irreversible outcome require a stronger explicit word, and still give package
 managers and scripts a conventional non-interactive flag.
@@ -253,8 +253,7 @@ updates preserve the file's indentation, line endings, and trailing-newline
 style. Custom
 or unrecognized files, separately installed optional skills, project-local
 integrations, external plan-save locations, and invalid configs are preserved
-(malformed host config is a fail-safe error unless `--skip-hosts` is explicit).
-If cleanup reports an error,
+(malformed host config is a fail-safe error). If cleanup reports an error,
 the CLI remains available for a safe retry, and its Windows PATH entry is
 retained or restored when possible. If PATH restoration itself fails, the
 output gives the full CLI path for retry and asks for manual PATH repair.
--- a/apps/hook/server/cli.test.ts
+++ b/apps/hook/server/cli.test.ts
@@ -35,7 +35,6 @@ describe("CLI top-level help", () => {
     expect(output).toContain("plannotator copilot-last [--gate] [--json] [--hook]");
     expect(output).toContain("plannotator setup-goal <interview|facts>");
     expect(output).toContain("plannotator uninstall [--purge] [--yes]");
-    expect(output).toContain("[--skip-hosts]");
     expect(output).toContain("Run 'plannotator <command> --help' for command-specific usage.");
     expect(output).toContain("running 'plannotator' without arguments is for hook integration");
   });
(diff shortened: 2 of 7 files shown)


backnotprop:
## Review (at `0b5ae578`)

Verdict: **merge as-is**. Full suite 2670 pass / 0 fail; focused uninstall/cli/install tests 148 pass. Typecheck clean. `scripts/` and `.github/` untouched, so the install-script invariants and CI gates are unaffected by construction.

**Branch topology:** not stacked and not divergent. #1170 was squash-merged to main, and `git diff` between the old branch head and main over every uninstall file is empty, so this is a single clean commit on current content. All three of #1170's late hardening mechanisms were re-verified empirically on this PR's compiled binary: case-variant purge refusal, the TOCTOU boundary against a hostile mid-run data-dir swap, and symlink/hardlink-safe unlinking.

**The design reversal holds.** Retiring `--skip-hosts` was the risk here, since the flag existed to escape dead-ends. All three original dead-end repros were run end to end on a compiled binary, and each is now a real recovery loop: block with guidance naming the exact file and field, follow the guidance, rerun completes with the binary removed and unrelated content intact. That includes the chmod-400 host config containing a managed hook (the user's own hook in the same file survived) and the malformed-but-unrelated Gemini settings. No unrecoverable state was found, and `grep` over the whole repo returns zero residual mentions of the retired flag. The compiled binary rejects `--skip-hosts` with exit 1. The retry command correctly propagates `--purge`.

**The new tests are genuine.** All six fail when pointed at pre-PR code, and every item in the PR body's regression list has a real assertion behind it, including the full block, repair, rerun-completes loop for both the malformed and uneditable cases, and the escaped-`"plannotator"`-spelling fail-safe.

**Fail-safe gate verified:** after a host error the CLI stays on PATH and runnable, data is preserved, and a post-repair rerun completes fully. Windows PATH retention tests all pass.

Two should-fixes for a follow-up, neither blocking:

**1. The missing-host-CLI guidance omits the recovery that works when the host is gone for good.** The Droid message says only "restore droid on PATH and run the plugin uninstall", but a permanently uninstalled Droid is the usual reason it is off PATH, and that was exactly the case the retired flag used to cover. The working alternative was verified: removing the `plannotator@plannotator` entry from `enabledPlugins` in `~/.factory/settings.json` makes detection go false and the next run completes. The message should name it (`uninstall.ts:516`).

**2. The headline invariant is exact-match only (pre-existing).** `isManagedHook` (`uninstall.ts:1540`) requires exact equality with the managed command strings, so a user who customized their hook with a documented flag, for example `plannotator --browser Safari`, is unmatched: reproduced with the binary deleted and the hook left dangling with no error. Not a regression, but it is now the claim this PR makes, so it deserves closing: before binary removal, re-scan host configs for any residual command containing a managed binary path and error (or at least warn with the filename) instead of proceeding.

Two nits: the malformed-config guidance tells users to remove Plannotator hooks from files that contain none (the actionable strict-JSON clause comes first, so harmless), and custom Kiro/Amp files at managed paths land in `preserved` while still referencing Plannotator, which is deliberate but technically a dangling reference under the strict reading.

PR #1191, opened 2026-08-04
chore(deps): bump @pierre/diffs to 1.3.2

backnotprop:
Pure pin bump of `@pierre/diffs` from exact `1.3.1` to exact `1.3.2` in the six package.json files (root, `packages/ui`, `packages/server`, `packages/review-editor`, `apps/review`, `apps/pi-extension`) plus `bun.lock`. No Plannotator code changes.

## Upstream delta (diffs-v1.3.1 to diffs-v1.3.2)

The Pierre repo history is linear between the two tags. The shipped `packages/diffs` source delta is four commits plus the version bump, two more than initially briefed:

- `94b65d72` fix(diffs): keep scroll anchor stable when CodeView removes items (#1047). This is the commit we want: it protects scroll position when virtualized CodeView items are removed, relevant to our virtualized all-files review view.
- `218def7f` diffs: add CodeView.removeItem API (#1050). New API only. **Adoption is deliberately deferred**; this PR changes no Plannotator code.
- `76207447` [diffs/edit] hide selection action when selection is offscreen (#1039). Editor surface polish; Plannotator does not use the diffs edit surface.
- `e72c4282` [diffs/edit] add `keymap` editor option (#1040). Same, editor surface only.
- `ecdfad73` updates Shiki to 4.4.1 in the Pierre repo's dev catalog only; the published `shiki` dependency range is unchanged (`^3.0.0 || ^4.0.0`) and our lockfile keeps shiki `3.23.0`.

## Young release disclosure and provenance verification

`1.3.2` was published `2026-08-04T02:19:04Z`, under one day before this bump, well inside our 7 day `minimumReleaseAge` posture (it installs via the `bunfig.toml` excludes for `@pierre/*`). The maintainer approved proceeding, compensated by the same deep provenance check used for the 1.3.1 bump:

- Tarball `diffs-1.3.2.tgz` downloaded from the registry: shasum `1dbe298a4f0b6f2935937a4c900a8df8f46eb0bf` and sha512 both match the registry `dist` metadata and the new `bun.lock` integrity line exactly.
- Registry signatures present (keyid `SHA256:DhQ8wR5APBvFHLF/+Tc+AYvPOdTpcIDqOhxsBHRwC7U`, the npm registry key).
- Zero lifecycle scripts in the shipped package (`scripts: {}`, no install/postinstall hooks). The repo-side `prepublishOnly` is publish-machine only and is not shipped.
- Maintainer set unchanged from 1.3.1: the same six accounts (fat, imownbey, mdo, nicolas.pierreco, slexaxton, amadeusdemarzi).
- Source byte comparison: every `sourcesContent` entry shipped in the tarball's 394 sourcemaps was compared against the `diffs-v1.3.2` tag (`39ef68a7`) of the Pierre repo. 191 unique repo files are byte identical, zero mismatches, zero files missing from the tag. Entries without content are 199 `.d.ts.map` declaration maps (never carry sources) and 2 generated asset shims with empty source lists.
- Shipped `package.json` matches the tag modulo the expected publish substitutions: `workspace:*` resolved to `@pierre/theme 2.0.0` and `@pierre/theming 1.0.0`, `catalog:` entries resolved to `diff 9.0.0`, `hast-util-to-html 9.0.5`, `lru_map 0.4.1`, all matching the tag's `pnpm-workspace.yaml` catalog. The 1.3.2 dependency set is byte identical to 1.3.1's.
- `@pierre/theme` resolution stays `2.0.0` and `@pierre/theming` stays `1.0.0` (unchanged, as required).

## Lockfile accounting

The `bun.lock` diff is exactly 7 changed lines, every one explained:

- 6 workspace dependency declarations: `"@pierre/diffs": "1.3.1"` to `"1.3.2"` (root, pi-extension, apps/review, review-editor, server, ui).
- 1 resolution entry: `@pierre/diffs@1.3.1` to `@1.3.2` with the new sha512 integrity (matches the tarball digest computed locally). The recorded dependency object is unchanged.

No other resolution moved: root `diff` stays `8.0.4`, `shiki` stays `3.23.0`, `@pierre/theme` stays `2.0.0`, `@pierre/theming` stays `1.0.0`.

## Verification

| Check | Result |
|---|---|
| Typecheck (`bun run typecheck`, all 7 tsconfigs) | pass |
| Full `bun test` | 2793 pass, 0 fail, 214 skip (no flake isolation needed) |
| `DOM_TESTS=1` isolated file-browser test (per test.yml) | 6 pass, 0 fail |
| `DOM_TESTS=1` seam-contract + DOM suite (per test.yml) | 155 pass, 0 fail |
| Build: `apps/review` | pass, 18,132.87 kB (gzip 5,418.50 kB) |
| Build: `build:hook` | pass, plan editor 23,026.88 kB (byte identical to 1.3.1) |
| Build: `build:opencode` | pass, copies verified |
| Build: `apps/pi-extension` | pass (vendor + HTML copies) |
| Bundle size delta vs 1.3.1 | review editor +786 bytes raw, +0.25 kB gzip; plan editor unchanged (does not bundle @pierre/diffs) |
| #880 fresh-install check | `bun pm pack apps/pi-extension`, clean-cache npm install of the tarball, `preloadFile` from `@pierre/diffs/ssr` renders 50 kB of prerendered HTML under plain Node v24.15.0 |
| CDP runtime, standalone demo HTML | 1.3.1 and 1.3.2 identical DOM counts (330 elements) and identical console output; 0 page exceptions |
| CDP runtime, live review server | initial load: 0 app console errors (one benign `/api/draft` 404 draft-restore probe), 0 exceptions; opening a changed file renders the pierre CodeView with identical DOM counts on 1.3.1 and 1.3.2 and 0 errors |

One pre-existing note surfaced during CDP testing: opening the `bun.lock` diff logs a `DiffHunksRenderer.processDiffResult: deletionLine and additionLine are null` console error. It reproduces identically on 1.3.1 with the same hunk, so it is not a 1.3.2 regression; the view still renders.

## Deferred

`CodeView.removeItem` (#1050) is not adopted here on purpose. If we want it for the all-files view later, that is its own change with its own review.

### DIFF
--- a/apps/pi-extension/package.json
+++ b/apps/pi-extension/package.json
@@ -52,7 +52,7 @@
   },
   "dependencies": {
     "@joplin/turndown-plugin-gfm": "^1.0.64",
-    "@pierre/diffs": "1.3.1",
+    "@pierre/diffs": "1.3.2",
     "@plannotator/webtui": "0.1.0",
     "chokidar": "^5.0.0",
     "diff": "^8.0.4",
--- a/apps/review/package.json
+++ b/apps/review/package.json
@@ -12,7 +12,7 @@
     "@plannotator/review-editor": "workspace:*",
     "@plannotator/server": "workspace:*",
     "@plannotator/ui": "workspace:*",
-    "@pierre/diffs": "1.3.1",
+    "@pierre/diffs": "1.3.2",
     "dompurify": "^3.3.3",
     "react": "^19.2.3",
     "react-dom": "^19.2.3",
--- a/bun.lock
+++ b/bun.lock
@@ -7,7 +7,7 @@
       "dependencies": {
         "@anthropic-ai/claude-agent-sdk": "^0.2.92",
         "@opencode-ai/sdk": "^1.3.0",
-        "@pierre/diffs": "1.3.1",
+        "@pierre/diffs": "1.3.2",
         "diff": "^8.0.4",
         "dockview-react": "^5.2.0",
         "dompurify": "^3.3.3",
@@ -88,7 +88,7 @@
       "version": "0.25.1",
       "dependencies": {
         "@joplin/turndown-plugin-gfm": "^1.0.64",
-        "@pierre/diffs": "1.3.1",
+        "@pierre/diffs": "1.3.2",
         "@plannotator/webtui": "0.1.0",
         "chokidar": "^5.0.0",
         "diff": "^8.0.4",
@@ -128,7 +128,7 @@
       "name": "@plannotator/review",
       "version": "0.0.1",
       "dependencies": {
-        "@pierre/diffs": "1.3.1",
+        "@pierre/diffs": "1.3.2",
         "@plannotator/review-editor": "workspace:*",
         "@plannotator/server": "workspace:*",
         "@plannotator/ui": "workspace:*",
@@ -205,7 +205,7 @@
         "@base-ui/react": "^1.6.0",
         "@fontsource-variable/geist-mono": "5.2.7",
         "@fontsource-variable/inter": "^5.2.8",
-        "@pierre/diffs": "1.3.1",
+        "@pierre/diffs": "1.3.2",
         "@plannotator/shared": "workspace:*",
         "@plannotator/ui": "workspace:*",
         "highlight.js": "^11.11.1",
@@ -226,7 +226,7 @@
       "name": "@plannotator/server",
       "version": "0.25.1",
       "dependencies": {
-        "@pierre/diffs": "1.3.1",
+        "@pierre/diffs": "1.3.2",
         "@plannotator/ai": "workspace:*",
         "@plannotator/shared": "workspace:*",
         "@plannotator/webtui": "0.1.0",
@@ -272,7 +272,7 @@
         "@fontsource-variable/inter": "^5.2.8",
         "@lezer/common": "^1.5.2",
         "@lezer/highlight": "^1.2.3",
-        "@pierre/diffs": "1.3.1",
+        "@pierre/diffs": "1.3.2",
         "@plannotator/atomic-editor": "^0.8.0",
         "@plannotator/core": "workspace:*",
         "@plannotator/markdown-editor": "^0.4.0",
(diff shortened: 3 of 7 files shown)


backnotprop:
## Delta review (at `2dd27702`)

Verdict: **merge as-is**. Diff shape confirmed exactly: six one-line pins plus exactly seven lockfile lines with the dependency object byte-identical (only version and integrity moved); theme, theming, root diff, and shiki resolutions all unchanged; zero lockfile drift on reinstall.

The four-commit upstream reading was verified against the tag range and, independently, against the shipped dist: only seven JS files changed and they map one-to-one onto the four commits. The edit-surface commits are confirmed unmounted in our code (no `@pierre/diffs/edit` import anywhere; the keymap/createEditor grep hits are our own unrelated code), and `removeItem` is genuinely not adopted. Stated plainly for the record: the two commits we do inherit both touch `components/CodeView.ts`, which our all-files view mounts, so this is a live code path change, not an inert bump; that is exactly where the scroll-anchor fix we want lives, and tests, builds, and the CDP checks are green over it.

Provenance spot-checks pass: three-way hash agreement between a locally packed tarball, the registry metadata, and the new lockfile line; zero lifecycle scripts; unchanged six-account maintainer set; signatures present. The pre-existing DiffHunksRenderer console complaint was proven pre-existing by a stronger method than runtime reproduction: the emitting file is byte-identical between 1.3.1 and 1.3.2.

Full battery green: 2793 pass / 0 fail, DOM sets 6/6 and 163/163, four builds, bundle delta matching the PR's numbers (+786B review, plan editor byte-identical).

Two low findings, neither blocking: the PR's Shiki sentence is imprecise (the dep range is untouched, but `worker-portable.js`, which we do not import, did move its inlined Shiki 4.2.0 to 4.4.1; the `worker.js` we bundle is byte-identical), and the DOM test count is misreported for the third PR in this train (155 claimed, 163 actual, all passing; the reporting pipeline deserves a fix at the source). Info: the squash-merge rebase hazard from stage 2 does not apply here since this branches directly off main.

PR #1212, opened 2026-08-05
fix(ui): render language-less code blocks as plain text

### LINKED ISSUE #1210: Language-less code blocks get auto-detected syntax highlighting instead of plain text
## Problem

Triple-backtick code blocks with no language tag (a bare fence with nothing after the backticks) are rendered with syntax highlighting. highlight.js auto-detects a language and colors tokens like `in`, `and`, `open`, `is`, `to`, which is wrong for plain-text blocks.

This shows up prominently when annotating documents that use fenced blocks as plain-text quote/preview containers (e.g. tweet drafts), where ordinary English words get colored as if they were code keywords.

Language-less code block auto-highlighted

## Cause

In `packages/ui/components/blocks/CodeBlock.tsx`, `hljs.highlightElement()` is called unconditionally. When `block.language` is `undefined`, highlight.js falls back to auto-detection instead of leaving the content as plain text.

The parser (`packages/ui/utils/parser.ts`) correctly sets `language` to `undefined` for a bare fence, so the component already has the information it needs to skip highlighting.

## Expected

A code block with no language should render as monospaced plain text, with no token coloring.

zeke:
This PR makes bare fenced code blocks (triple backticks with no language tag) render as plain monospaced text instead of getting auto-detected syntax highlighting.

`hljs.highlightElement()` was called unconditionally in three components. When a block has no language, highlight.js falls back to auto-detection and colors ordinary words like `in`, `and`, `open`, `is`, `to`. This is most visible in annotate mode when a document uses fenced blocks as plain-text quote/preview containers.

The parser already sets `language` to `undefined` for a bare fence, so the fix just guards the highlight call in:

- `packages/ui/components/blocks/CodeBlock.tsx`
- `packages/ui/components/plan-diff/PlanCleanDiffView.tsx`
- `packages/ui/components/Viewer.tsx`

Blocks with a language are unchanged.

Fixes https://github.com/backnotprop/plannotator/issues/1210

### DIFF
--- a/packages/ui/components/Viewer.tsx
+++ b/packages/ui/components/Viewer.tsx
@@ -630,7 +630,11 @@ export const Viewer = forwardRef<ViewerHandle, ViewerProps>(({
           const block = blocks.find(b => b.id === codeEl.closest('[data-block-id]')?.getAttribute('data-block-id'));
           codeEl.removeAttribute('data-highlighted');
           codeEl.className = `hljs font-mono${block?.language ? ` language-${block.language}` : ''}`;
-          hljs.highlightElement(codeEl);
+          // Skip highlighting language-less fences so highlight.js doesn't
+          // auto-detect a language and color plain text.
+          if (block?.language) {
+            hljs.highlightElement(codeEl);
+          }
         }
       });
 
--- a/packages/ui/components/blocks/CodeBlock.tsx
+++ b/packages/ui/components/blocks/CodeBlock.tsx
@@ -16,13 +16,17 @@ export const CodeBlock: React.FC<CodeBlockProps> = ({ block, onHover, onLeave })
   const containerRef = useRef<HTMLDivElement>(null);
   const codeRef = useRef<HTMLElement>(null);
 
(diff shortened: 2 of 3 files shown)


zeke:
Heads up @backnotprop! This is not a slop PR. 🤚

This is intended to be a small meaningful change to a real bug. Hope it's helpful.

PS I love Plannotator. So useful. Thanks for creating it.

backnotprop:
Hey, great.

I'll get this in and deploy a patch for you. There was reason to do this early on and specifically use highlight JS. Without auto detection I'll probably still end up removing it, but for better reasons than I first assumed: we already ship a second highlighter (Shiki, via the diff renderer), so hljs is redundant, and hljs hasn't published to npm in a long while even though fixes keep landing in git. Size-wise it's actually not the heavy part of our bundle.

backnotprop:
and thanks

backnotprop:
Merged. Clean diagnosis and a surgical fix; the screenshot made the case instantly. This ships in the next release. Welcome aboard as a contributor!

PR #1228, opened 2026-08-07
fix(pr-artifacts): read GitLab uploads through the token-readable API

yuensunn:
## Problem

GitLab serves `/uploads/<secret>/<file>` from a Rails web route that only honors **session cookies**. The `PRIVATE-TOKEN` header that `pr-artifact-document.ts` already resolves via `glab config get token` is not accepted there, so every GitLab upload attachment referenced by an MR is unreadable.

Worse, the failure is silent. The sign-in hop answers HTTP 200 with an HTML login page, which then gets rendered *as the artifact* in the artifact pane.

Measured against a self-hosted GitLab 18.8.0-ee instance with a valid token (`/api/v4/user` → 200):

```
/uploads/<secret>/pack.md                        → 302 → /users/sign_in
/<group>/<repo>/uploads/<secret>/pack.md         → 302 → /users/sign_in
/-/project/<id>/uploads/<secret>/pack.md         → 404
/api/v4/projects/<id>/uploads/<secret>/pack.md   → 200  ✅
```

## Changes

1. **Route GitLab uploads through the uploads REST API.** `gitlabUploadApiUrl()` rewrites both the bare `/uploads/...` and project-scoped `/<projectPath>/uploads/...` forms to `GET /projects/:id/uploads/:secret/:filename`, wired into the existing `providerContentUrl()` alongside the `/-/blob/` → `/-/raw/` rewrite. The allowlist and credential plumbing are unchanged — they were already correct.

2. **Stop following redirects into a provider sign-in page.** Landing on `/users/sign_in`, `/users/auth/...`, `/login`, `/session`, or `/oauth/authorize` now throws `PRArtifactDocumentError(401)` with the matching `gh auth login` / `glab auth login --hostname <host>` hint. An auth gap becomes legible instead of surfacing as a corrupt artifact. This is worth having independently of change 1.

3. **Refine the opaque content type.** The uploads API answers every file with `application/octet-stream`, and `/api/pr-artifact-content` serves the provider type with `nosniff` — without this, images and video would stop rendering. Falls back to an extension lookup only when the provider type is `application/octet-stream`.

## Verification

- `bun test packages/shared/` — 761 pass, 0 fail (4 new tests)
- `tsc --noEmit -p packages/shared/tsconfig.json` — clean
- Full `bun test` — 2789 pass / 25 fail; unmodified `main` produces the identical 25 failures (Pi-extension and AI-runtime suites, unrelated to this change)
- Live end-to-end fetch of a real MR upload on GitLab 18.8.0-ee: returns the actual file, `contentType: text/markdown`

## Known consideration

`GET /projects/:id/uploads/:secret/:filename` was added in GitLab 17.4. On older self-hosted instances it 404s, which surfaces as a 502 — where today those uploads fail anyway (redirect to sign-in). Happy to add a version gate or a fallback to the web route if you'd prefer explicit handling.

## Note on CI

The `Test` job failure on this branch is `edit-mode Discard repaints the pristine diff (DOM) > pristine content is visible ... promptly after cancelEdit`. It reproduces identically on upstream `main` (run 31138933804) and is unrelated to this change.

### DIFF
--- a/packages/shared/pr-artifact-document.test.ts
+++ b/packages/shared/pr-artifact-document.test.ts
@@ -41,6 +41,20 @@ const github: GithubPRMetadata = {
   url: 'https://github.com/acme/widgets/pull/1',
 };
 
+const gitlabMetadata: GitlabMRMetadata = {
+  platform: 'gitlab',
+  host: 'gitlab.example.com',
+  projectPath: 'acme/widgets',
+  iid: 7,
+  title: 'Artifacts',
+  author: 'reviewer',
+  baseBranch: 'main',
+  headBranch: 'feature',
+  baseSha: 'base',
+  headSha: 'head',
+  url: 'https://gitlab.example.com/acme/widgets/-/merge_requests/7',
+};
+
 describe('isPRArtifactDocumentUrlAllowed', () => {
   test('allows referenced GitHub uploads and raw files from the active repository', () => {
     expect(isPRArtifactDocumentUrlAllowed(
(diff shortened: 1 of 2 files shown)


yuensunn:
### Before / after in the artifact pane

Reproduced against an MR whose description links a `/uploads/<secret>/markdown-sample.md` attachment, on a self-hosted GitLab with SSO. Same MR, same `glab` credentials, only the patch differs.

**Before** — the artifact pane renders the login page's HTML source instead of the file. The pane shows a `<!DOCTYPE html>` document with a hidden auto-submitting auth form, because the sign-in redirect answers HTTP 200 and that body is passed through as the artifact content.

**After** — `markdown-sample.md` renders as markdown: headings, lists, block quotes, code samples, and the `From PR description by @…` footer.

Also confirmed at the API layer on a separate MR — `GET /api/pr-artifact-document?url=<upload>` on the local review server returns `HTTP 200` and the real file content, where the same request on `main` returns the login page body.

Screenshots attached below.
Screenshot 2026-08-07 at 11 01 48 AM
Screenshot 2026-08-07 at 11 05 25 AM

backnotprop:
**TLDR:** merging this. I pushed 5 commits onto your branch (4 fixes + 1 test) covering the pre-17.4 fallback you offered, a tighter upload pattern, direct-401 diagnosis, and dropping the active content types. Nothing in your diagnosis or your seam choice changed. Please review and push back on anything you disagree with.

*Note: this review was AI-assisted (Claude Code), with me driving and checking the conclusions.*

First, the good part. Your diagnosis is correct and unusually well evidenced: the four-row table of what each GitLab route actually answers with a valid token is exactly the artifact that makes this reviewable, and "the sign-in hop returns 200 with an HTML login page, which then renders as the artifact" is the kind of failure that normally ships unnoticed for a year. The rewrite is at precisely the right seam. `providerContentUrl()` is already the single place where a provider URL becomes a fetchable URL, so the allowlist, the credential plumbing, and the redirect guard all keep working unchanged around it, which is why the change reads as small. You were also right that the sign-in guard is worth having independently of the rewrite.

We also ran an adversarial pass specifically on token handling, since the rewrite moves a `PRIVATE-TOKEN` request to a new path: 13 crafted inputs (cross-origin redirect chains, port-shifted hosts, protocol downgrades, project-path confusion, encoded traversal) came back with zero leaks. The credential stays on the provider origin in every case.

What the 5 commits change:

1. `ebfe1a2b` **rewrite only real GitLab upload paths.** `GITLAB_UPLOAD_RE` matched `/uploads/<anything>/<anything>`, and the remainder allowed slashes. Since MR bodies are attacker-authored in the general case, that made the rewrite into a way to append a chosen path to a credentialed `/api/v4/...` GET, with GitLab's router as the only thing standing in the way. Now pinned to the shape GitLab actually mints: a 32 lowercase hex secret and a single filename segment. Anything else is fetched verbatim as before, so nothing legitimate loses the rewrite. Your test fixtures moved to a real 32-hex secret.

2. `8fbd3fe8` **diagnose a direct provider refusal as a missing login.** Your sign-in guard only fires on a redirect, but the `/api/v4` routes refuse a bad or absent token directly with 401/403 JSON, which fell through to `Artifact host returned HTTP 401` at status 502. Same wrong-looking outcome your guard set out to fix, just via a different hop. The messaging is hoisted into one helper and reused for both. GitHub's 403 stays a plain transport status on purpose, because it also covers rate limiting and SSO enforcement, where a `gh auth login` hint would send the reader down the wrong path.

3. `852a5950` **fall back to the upload web route when the API route is absent.** This is the handling you offered in "Known consideration", and yes, it is worth having: on a pre-17.4 self-hosted instance with a public project, the web route reads anonymously today, so the rewrite turns a working upload into a 502. On a 404 from the rewritten URL we retry the original route once. Narrow by construction: only the upload rewrite, only 404, only when the 404 came from the rewritten URL itself, no version probing. It reuses the loop's redirect budget so the request count stays bounded, and it is same-origin so your `shouldSendProviderAuth` logic applies unchanged. A private project on an old instance still gets the actionable 401 through the sign-in guard rather than a bare 404.

4. `950857bd` **drop `html` and `js` from the extension map.** These were the one part of change 3 that had no functional upside. An HTML artifact is read through `/api/pr-artifact-document`, which always answers `text/plain; charset=utf-8`, so nothing ever consulted the map for it. Keeping them just meant the media route could label an upload as active content, with the `Content-Security-Policy: sandbox` header as the only thing preventing that from mattering, forever.

   On `svg` we kept it, and here is the reasoning so you can disagree: `.svg` is not in the review editor's `IMAGE_EXTENSIONS`, so an svg upload arrives either as an authored markdown image, rendered through ``, which is a non-scripting context by spec, or as a resource referenced from an HTML artifact, which renders inside an `<iframe sandbox="">`. Both are served by the media route with `Content-Security-Policy: sandbox` plus `X-Content-Type-Options: nosniff`, so even a direct navigation to the proxy URL lands in a sandboxed opaque-origin document that cannot run script. Removing it would cost real rendering (nosniff means an unrefined octet-stream svg simply does not appear) and would not close any reachable capability. `css` also stays, since `shouldRewriteCss` keys on `text/css` to rewrite provider references.

5. `42d4f079` **pin the token-reach invariant.** Since the rewrite changes where a credentialed request goes, the thing worth guarding is that it did not also change how far the credential can travel. GitLab object storage answers uploads with a 302 to a signed URL on an unrelated host, which is exactly the hop a token must not follow. Verified the test bites by forcing `shouldSendProviderAuth` to return true for gitlab: it fails on the second request's header.

Gates: `bun test packages/shared` 768 pass 0 fail, `tsc --noEmit` clean on `packages/shared`, `packages/server`, and `apps/pi-extension`. `apps/pi-extension/vendor.sh` re-run and the generated Pi copy confirmed byte-identical apart from the expected import-specifier rewrite. Your original commit is preserved as-is at the base of the branch.

Thanks for the careful writeup, and especially for measuring the routes instead of guessing at them. If any of the above reads wrong to you, particularly the `svg` call or the narrowness of the 404 retry, say so and push back.

PR #1232, opened 2026-08-07
feat(annotate): manual refresh rendered HTML from disk

leoreisdias:
## Context

Markdown annotate sessions already reflect agent edits as soon as the underlying file changes, but rendered HTML sessions keep showing the snapshot loaded at startup. That makes the review surface look stale after an agent updates the source file.

This PR adds the first, deliberately manual checkpoint: local HTML annotate sessions get a **Refresh** action beside **Hide tools**. The user stays in control while we validate the refresh and annotation-restoration behavior before considering filesystem-driven automatic refresh.

## What Changed

- Add a Refresh action for local `.html` / `.htm` annotate sessions.
- Fetch the current rendered document through the existing `/api/doc` route and remount the sandboxed HTML viewer, including when the HTML text is unchanged.
- Reapply existing annotations after the iframe reload and report anchors that no longer match, while retaining their comments in the annotations panel.
- Reset stale HTML diff and share state after a successful refresh.
- Ignore refresh responses that belong to an older request or a document the user has already left.
- Invalidate asynchronous share results against the complete sharing context so an older request cannot restore a link that omits newer HTML, annotations, or attachments.
- Keep the refresh lifecycle in `useHtmlRefresh`; `App` only supplies the active path, capability flag, and successful-snapshot transition.

## Contracts Preserved

- Refresh is available only for local rendered HTML in active annotate mode. URL annotations, archive views, Markdown-converted HTML, and non-HTML documents are unchanged.
- HTML source saving remains disabled; this only reloads the rendered source from disk.
- Annotation comments survive even when their original text anchor disappears.
- The existing sandboxed iframe bridge remains the only path used to restore HTML annotations.

## Review Focus

The important areas are:

1. Request and navigation races in `useHtmlRefresh`.
2. Annotation restoration acknowledgements across the iframe bridge.
3. Share-request invalidation when content changes during an asynchronous request.
4. The capability check that keeps Refresh off unsupported or read-only surfaces.

## Verification

- Default test invocation cleanly skips the new DOM-gated tests.
- 14 focused DOM tests pass, covering the action UI, source snapshot handling, annotation restoration, and stale share requests.
- Full TypeScript project-reference matrix passes.
- Review and hook production bundles build successfully.
- Manual browser check against a local rendered HTML file confirms the document mounts and Refresh reloads it with a success notification.


backnotprop:
Pushed three commits to your branch as maintainer edits: a fix for a share-link race the review turned up (the share request context included the resolver's identity, which changes as a side effect of its own success, so the first short link on an HTML session discarded its own result), a regression test for it folded into your contentRevision suite, and a merge of current main. Everything is green: typecheck plus the full HTML-surface DOM suite including the new 0.27.8 tests. The feature itself reviewed very well, nice work on the refresh lifecycle and the restoration reporting. This is now at the front of the queue.

backnotprop (approved):
Two review passes on this branch, with the fixes pushed as maintainer commits so the feature was not held up.

Round 1 found one regression in the share-link invalidation: the request context keyed on the resolver's identity, which changes as a side effect of its own success, so the first short link on an HTML session discarded its own result and left the button loading. Fixed in d59003e7, with a regression test in be48c80e that fails on the unfixed hook.

Round 2 found that /api/share-html answered the root document from the startup snapshot, so a share link created after a Refresh carried the pre-refresh page. Fixed in a2a0e200 on both the Bun server and the Pi mirror, with tests on each side. Also in this round: aria-disabled instead of disabled on the Refresh button so keyboard focus is not dropped (ec4d8a03, with the HtmlSurfaceActions test now asserting state instead of pinning copy), an honest rename of the vimHud test (afff38a9), a comment on the bridge's change-only restore ack (85d226a0), and a fix for the PR's own invalidation tests, which were timing-sensitive and behind the red CI on the previous head (e5c5b830).

Verified independently at e5c5b830: the refresh races (stale response, navigation away, double click, old-iframe messages after remount), capability gating (URL, live app, archive, converted HTML, markdown, and read-only views all excluded), the comment-only clamp and sandboxed-restore-only path untouched, the merge commit a pure merge, typecheck clean, 607/607 in the CI seam-contract and DOM group, 55/55 Bun annotate, 23/23 Pi annotate.

Two pre-existing asymmetries noted for later, not for this PR: the linked-doc branch of /api/share-html is uncapped while /api/doc caps at 2MB, and the deleted-file fallback needs a lexical containment match under symlinked roots.

The refresh lifecycle, remount, and restoration reporting are well built. Approving. Merge once CI finishes on this head.

backnotprop:
Follow-up to the approval above, covering two more maintainer commits and a browser verification run.

We ran the feature end to end in a real browser (Playwright, recorded) against e5c5b830: gating on html vs md vs converted sessions, refresh with three annotation kinds including one whose anchor was destroyed, unchanged-file refresh, missing-file refresh, keyboard activation and double-click, share-html after refresh, draft across reload, and the Esc ladder plus Hide tools before and after a refresh. Everything passed. Two things came out of it and are now fixed on the branch:

- 017bb64e: the singular toast read "1 annotation no longer match the HTML"; it now agrees.
- f7a70429: a browser reload after a Refresh reverted the page to the startup snapshot because /api/plan never re-read disk, while the draft annotations persisted against the replaced page. /api/plan and /api/share-html now share one root read on both runtimes (current bytes under the 2MB cap, startup snapshot only for a missing or oversized file). The version-diff fields are omitted rather than recomputed when the served bytes differ from the snapshot, which is exactly what the in-app Refresh already does on the client, so reload and refresh converge. Non-HTML sessions are untouched. Tests on both servers cover the version chain and the fallback.

Independently re-verified at f7a70429: both runtimes match line for line, every non-root session mode traced to an unchanged payload, the UI handles the omitted fields with no dead badge, typecheck clean, Bun annotate 56/56, Pi annotate 24/24, DOM group green, CI green. Approval stands.

PR #1091, opened 2026-07-20
feat(annotate): add strict atomic result output

rNoz:
> **Important/Disclaimer (same note in all my PRs):** This is not spam (I created 4 PRs after the weekend of hard work + 2 from previous weeks), just my willing to help: I have been stress testing this tool (thank you for Plannotator!), finding some enhancements, proposing a couple of improvements and fixing some issues. I keep all these features/fixes in topic branches (to be able to publish them, now here as PRs), but I have been using all them in a local integration branch that I use to run plannotator with all these contributions (I use it in all projects and CLI harnesses I own). Therefore, apart from the tests per branch and ad hoc manual tests, I have been using it flawlessly. I decided to create the PRs, in case they are useful for you. Feel free to drop, update, extend.... anything you want. I am willing to contribute, so if you prefer that I tackle something or rebase/solve merge conflicts in the future (At 2026-07-20: all is ready!), please, don't hesitate to ping me. I am sharing this amount of work in case it is useful for you all.
> The plannotator "binary" that I use many times per day, it is based on the integration branch that has these PRs (branches cherrypicked) in order: 
> - https://github.com/backnotprop/plannotator/pull/1076, https://github.com/backnotprop/plannotator/pull/1078 (using them for 2 weeks)
> - https://github.com/backnotprop/plannotator/pull/1087, https://github.com/backnotprop/plannotator/pull/1089, https://github.com/backnotprop/plannotator/pull/1091, https://github.com/backnotprop/plannotator/pull/1092 (using them for 3 days; now just rebased)
>
> image

## Summary

Direct `plannotator annotate` gates now support optional fail-closed approval status and atomic structured-result files.

`--require-approval` makes annotated and dismissed decisions exit nonzero after their JSON is published. `--result-file <path>` writes the exact newline-terminated stdout record through a private same-directory temporary file and atomically publishes it without overwriting an existing destination.

Existing plaintext, JSON, and hook-native paths still use the legacy emitter and retain their current output and zero exit status.

Implements https://github.com/backnotprop/plannotator/issues/1090

## Why

Structured stdout identifies each annotate decision, but every legacy decision exits successfully. Automation must parse stdout before it can distinguish approval from requested changes.

Detached callers also need an atomic readiness boundary. Shell redirection can expose an empty or partial file, while a pre-existing or concurrently created destination can be mistaken for the current result. Atomic no-clobber publication makes one complete JSON record the only successful file outcome.

[...]

PR #1237, opened 2026-08-08
fix(annotate): persist submitted feedback before deleting the draft (#678)

### LINKED ISSUE #678: Annotate feedback can be lost when submit auto-closes localhost session after CLI timeout on Windows/Codex

backnotprop:
## TLDR

An annotate submit could silently lose all feedback when the invoking CLI/agent had already timed out: the server settled the decision promise with nobody listening, deleted the draft, and the submitted annotations then existed nowhere. Both runtimes now write a durable record of the submitted feedback to `~/.plannotator/history/{project}/{slug}/submissions/{timestamp}.md` before the draft is deleted. If that write fails, the draft is kept as the recovery copy. With annotate history disabled, nothing new is written and the legacy submit behavior is preserved unchanged.

## The loss window

`plannotator annotate` hands the browser's decision back through a promise the invoking CLI/agent awaits. When the agent-side shell command times out (the Codex-on-Windows repro in #678), the user can keep reviewing in the still-open browser and click Send Feedback. The server then:

1. settles the decision promise (nobody is listening anymore),
2. deletes the annotation draft,
3. responds OK and the page closes.

At that point the feedback exists nowhere: not in drafts (just deleted), not in annotate history (that stores versions of the source file, not annotations), not anywhere. The plan server does not have this hole because it persists a decision snapshot on approve/deny (`saveAnnotations` / `saveFinalSnapshot`); the annotate server had no equivalent. Adjacent work does not cover it either: #1091's `--result-file` applies only to strict `--gate --json` invocations, and #1143 preserves the draft on abandonment, not on a successful submit.

## The fix

New shared machinery, reusing the existing history storage layout rather than inventing a parallel scheme:

- `saveAnnotateSubmission()` in `packages/shared/storage.ts`: writes one markdown file per submit to `{DATA_DIR}/history/{project}/{slug}/submissions/{timestamp}.md`, right next to the file's annotate version history (`slug` is the same `deriveAnnotateHistorySlug` slug the version snapshots use). The `submissions/` subdirectory keeps records out of the numeric `NNN.md` version scans, and filenames carry a collision counter so rapid submits never overwrite.
- `persistAnnotateSubmission()` in `packages/shared/annotate-history.ts`: composes the record (source path, decision kind, timestamp, and the exported feedback text, which already embeds every annotation in human-readable form; raw annotations JSON only as a defensive fallback when the text is empty). Never throws; returns null on storage failure.

Both annotate servers (Bun `packages/server/annotate.ts` and Pi `apps/pi-extension/server/serverAnnotate.ts`, re-vendored via `vendor.sh`) wire it into `/api/feedback` and `/api/approve` after the decision settler wins and before `deleteDraft`. Ordering matters twice:

- persist happens after `decision.settle()` wins, so a 409 losing producer never writes a phantom record, and
- `deleteDraft` only runs when the record was written (or persistence was legitimately skipped). If the durable write fails, the draft stays behind as the only remaining copy of the reviewer's work; the decision itself still succeeds because persistence is an enhancement, never a gate.

`/api/approve` is not contentless: approve-with-notes carries `feedback`/`annotations`, so it persists under the same rule. A bare approve carries no user content and writes nothing. `/api/exit` is untouched.

## annotateHistory-disabled policy

`PLANNOTATOR_ANNOTATE_HISTORY=0` / `{ "annotateHistory": false }` means "do not write copies of annotated content to the data dir", and submitted feedback quotes that content (annotation `originalText` excerpts). So with history disabled, no submission record is written, and the submit path behaves exactly as before (draft deleted, response OK). The task's alternative of skipping the draft delete only when the decision promise has no live consumer is not implementable: the server cannot detect in-process that its caller stopped reading the decision. And keeping the draft on every opted-out submit would resurface already-delivered feedback on the next session for the same content, which contradicts the normal submit contract everywhere else (plan server included). #1143's precedent keeps the draft only when no decision was delivered (abandonment); here a decision was made, so the opt-out user gets the pre-existing, explicitly chosen stateless behavior. The failed-write path (history enabled but the data dir is unwritable) is the one place the draft is deliberately kept, since the durable record was expected and did not happen.

## Recovery discoverability

`plannotator sessions` lists live server processes from `~/.plannotator/sessions/{pid}.json`, so the new records are not discoverable from it (and per scope, no new CLI surface is added here). Records are plain markdown under `~/.plannotator/history/{project}/{slug}/submissions/` and sit next to the version snapshots for the same file.

## Tests

- Bun (`packages/server/annotate.test.ts`, new describe): record written and draft gone after `/api/feedback`; approve-with-notes persists while bare approve writes nothing; disabled history writes no content while still deleting the draft; a failed durable write keeps the draft.
- Pi (`apps/pi-extension/server/annotate-submission.test.ts`): mirrors the feedback, approve-with-notes, and disabled-history cases against the Node server.
- `bun test packages/server packages/shared`: 1362 pass, 0 fail (82 files).
- `bun test apps/pi-extension`: 186 pass, 0 fail (21 files).
- `bun run typecheck` (includes `vendor.sh`): clean.

Closes #678

### DIFF
--- a/apps/pi-extension/server/annotate-submission.test.ts
+++ b/apps/pi-extension/server/annotate-submission.test.ts
@@ -1,0 +1,190 @@
+/**
+ * Annotate server (Pi/Node): durable submit records (#678)
+ *
+ * Node mirror of packages/server/annotate.test.ts's "durable submit records"
+ * describe block. The decision promise's consumer (the invoking agent) can
+ * time out before the reviewer submits; the submit then settled the promise
+ * with nobody listening, deleted the draft, and the feedback existed nowhere.
+ * These tests pin the fix in serverAnnotate.ts: a durable record is written
+ * to history/{project}/{slug}/submissions/ BEFORE the draft is deleted, and
+ * the annotate-history opt-out suppresses the record while keeping the
+ * legacy submit behavior.
+ *
+ * History writes go to the real ~/.plannotator data dir (generated/storage
+ * caches its data directory at import time — see annotate-history.test.ts),
+ * so each test uses a unique project namespace cleaned up in afterAll.
+ */
+
+import { afterAll, afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { existsSync, mkdtempSync, readdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join, resolve } from "node:path";
+import { startAnnotateServer } from "./serverAnnotate.ts";
+import { deriveAnnotateHistorySlug } from "../generated/annotate-history.ts";
+import { getPlannotatorDataDir } from "../generated/data-dir.ts";
+
+describe("pi annotate server: durable submit records (#678)", () => {
+	let savedPort: string | undefined;
+	let savedRemote: string | undefined;
+	let savedHistoryFlag: string | undefined;
+
+	beforeEach(() => {
+		savedPort = process.env.PLANNOTATOR_PORT;
+		savedRemote = process.env.PLANNOTATOR_REMOTE;
+		savedHistoryFlag = process.env.PLANNOTATOR_ANNOTATE_HISTORY;
+		delete process.env.PLANNOTATOR_PORT;
+		process.env.PLANNOTATOR_REMOTE = "0";
+		// Force the toggle on unless a test explicitly flips it off — a real
+		// ~/.plannotator/config.json must never change the outcome.
+		process.env.PLANNOTATOR_ANNOTATE_HISTORY = "1";
+	});
+
+	afterEach(() => {
+		if (savedPort === undefined) delete process.env.PLANNOTATOR_PORT;
+		else process.env.PLANNOTATOR_PORT = savedPort;
+		if (savedRemote === undefined) delete process.env.PLANNOTATOR_REMOTE;
+		else process.env.PLANNOTATOR_REMOTE = savedRemote;
+		if (savedHistoryFlag === undefined) delete process.env.PLANNOTATOR_ANNOTATE_HISTORY;
+		else process.env.PLANNOTATOR_ANNOTATE_HISTORY = savedHistoryFlag;
+	});
+
+	const mintedProjects: string[] = [];
+	function uniqueProject(label: string): string {
+		const project = `_pi_annotate_submission_test_${label}_${Date.now()}_${Math.random().toString(36).slice(2, 8)}`;
+		mintedProjects.push(project);
+		return project;
+	}
+
+	afterAll(() => {
+		const historyDir = join(getPlannotatorDataDir(), "history");
+		for (const project of mintedProjects) {
+			rmSync(join(historyDir, project), { recursive: true, force: true });
+		}
+	});
+
+	function submissionsDir(project: string, docPath: string): string {
+		return join(
+			getPlannotatorDataDir(),
+			"history",
+			project,
+			deriveAnnotateHistorySlug(resolve(docPath)),
+			"submissions",
+		);
+	}
+
+	// The project name is baked into the markdown so every test gets a unique
+	// content-hashed draft key — drafts live in the real data dir and identical
+	// markdown across tests would collide on one draft file.
+	async function startServer(project: string, docPath: string) {
+		const markdown = `# Doc ${project}\n\nBody\n`;
+		writeFileSync(docPath, markdown, "utf-8");
+		return startAnnotateServer({
+			markdown,
+			filePath: docPath,
+			htmlContent: "<html></html>",
+			project,
+		});
+	}
+
+	test("feedback submit writes a durable record and only then deletes the draft", async () => {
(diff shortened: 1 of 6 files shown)


backnotprop:
**TLDR:** An adversarial verification pass tried to break every user-reachable path of this PR. The core mechanism held (no data loss, no availability regressions, exit-code contract intact, Bun/Pi parity confirmed), but it proved two unintended behavior changes, both now fixed in 095f7e53. 1552 tests pass.

*This verification and the fixes were AI-assisted.*

**Fixed:**
1. The durable record had no mode gate: annotate-last and URL sessions, which never touched the data dir before, would persist submitted feedback (quoting the agent's message or the fetched page) under history/, silently widening the documented annotateHistory contract. The record now shares the exact single-local-file eligibility gate the version history uses, so previously-stateless modes stay stateless.
2. A non-string feedback body on /api/feedback regressed from 200 to a thrown 500 (after the decision had already settled). Only reachable by third-party API clients, but now it degrades to the exact legacy behavior instead.

**Verified and could not be broken:** version scans ignore the submissions/ subdir in every listing path, 50 sub-millisecond submits produce 50 distinct records, concurrent submits write exactly one record for the winner, write failures never escape as errors on an otherwise-successful submit, the record lands on disk before the awaiting CLI can observe the decision, the strict-gate exit-code contract is untouched, and vendor.sh already covers both shared modules so the published Pi package resolves cleanly.

**Known and accepted:** a failed durable write keeps the draft, so a later identical session can show a restore banner holding already-submitted annotations. That is the intended recovery tradeoff.

PR #1243, opened 2026-08-09
feat(annotate): pinpoint-first raw-HTML sessions with element anchors and a minimal-first render

backnotprop:
**TLDR:** Raw-HTML annotate sessions now open as just the page (all chrome hidden on first run, last-used chrome remembered afterwards), default to Pinpoint, and get rebuilt pinpoint mechanics: overlay-based hover highlighting, click-to-pin straight into the comment popover, serialized CSS element anchors for reliable restoration, and numbered pin badges for element-only annotations. UI-only; no server changes. Markdown sessions are untouched.

### 1. Pinpoint is the HTML default
- `packages/ui/utils/inputMethod.ts` resolves the input method per surface. HTML sessions read/write their own cookie (`plannotator-input-method-html`) and default to Pinpoint when nothing is saved.
- Decision: the legacy shared key is ignored for HTML. It was only ever written from markdown sessions, so honoring it would let a markdown-era "drag" cookie silently suppress the new HTML default for every existing user. An explicit switch made inside an HTML session persists for HTML sessions only (mirrors the per-origin AI provider pattern).

### 2. Rebuilt pinpoint mechanics (bridge script)
All inside the sandboxed iframe bridge (`packages/ui/components/html-viewer/bridge-script.ts`); only validated, size-capped DTOs cross postMessage.

- Hover is now a dedicated fixed-position outline box + element label. The page's own elements are never touched (the old implementation wrote a class onto author DOM on every hover). Vim navigation keeps its class-based target tint.
- Hit-testing via `elementFromPoint` at the pointer, snapped to the nearest semantic target (Plannotator's existing block/inline/cell graph, so pins land on meaningful elements rather than bare wrappers).
- A rAF-coalesced reconcile pass (capture-phase scroll + resize + ResizeObserver) keeps the hover box glued under a stationary cursor while the page scrolls, tracks the pinned element, and repositions pin badges.
- Click-to-pin: capture-phase click suppresses page behavior, outlines the element in the pinned (accent) style while composing, and goes straight to the existing comment popover, skipping the intermediate toolbar. Crosshair cursor while pinpoint is active; Escape cancels the pending pin or drops the hover.
- Element anchors: each pin serializes a verified-unique CSS selector (semantic ladder: id, then identity attributes like `data-testid`/`aria-label`/`href`, then meaningful classes with generated-class filtering, then a `tag > tag:nth-of-type()` path, every rung proved unique with a real query) plus a tag + normalized text snapshot fingerprint. Restoration is anchor-first and fails closed: a weak positional selector must also match the text snapshot, otherwise it falls back to the existing document-wide text search. Annotations without anchors (older drafts, share links) restore exactly as before.
- Numbered pin badges: element pins that cannot take an inline `<mark>` (SVG nodes, drifted text) previously had no visual at all after commit. They now get a numbered badge anchored to the element box, re-acquired through the anchor after client-side re-renders, clickable to select the annotation.
- Data model: `Annotation.htmlAnchor` added additively (`packages/ui/types.ts`). Exported feedback, drafts (full JSON, carries the field automatically), and share links (compact tuples, restore via text search) keep working.

### 3. Minimal-first render with remembered chrome
- First-ever raw-HTML session opens with everything the "Hide tools" toggle controls hidden, plus the sidebar closed. The header "Show tools" affordance stays visible and Mod+B still opens the sidebar.
- From then on, sessions open with exactly the chrome the user last left (tools visibility + sidebar open state), persisted as a JSON cookie (`plannotator-html-chrome` in `packages/ui/utils/htmlChrome.ts`), following the cookie-persistence pattern from CLAUDE.md. A pre-restore render can never clobber the saved state (writer is gated on the restore having applied).

## What we took from each reference

**app-notes-extension:**
- The hover visual: a fixed-position border box sized to the element's live rect, `pointer-events: none`, never a style/class write on the page. Label pill with flip-below-when-clipped and horizontal clamping.
- Identity-gated hover updates (only restyle when the resolved element changes) and the rAF-coalesced reconcile worker fed by capture-phase scroll + resize that re-hit-tests at the last pointer position, so the outline follows the element under a stationary cursor during scroll.
- The anchor design: verify-as-you-generate selectors (every semantic rung proved unique with a real query before acceptance, `:nth-of-type` only when needed), generated-class filtering, and fail-closed restoration where strictness scales inversely with selector strength (weak selectors must also match the captured text snapshot; a shifted list never mis-anchors).
- Re-acquiring a pinned element through its anchor after the page re-renders under it.

**agentation:**
- The crosshair cursor affordance while pinpoint mode is active, via an attribute-scoped `cursor: crosshair !important` rule.
- The numbered marker badges: small circular badges, creation-order numbering, pop-in animation, hidden (not chased) when their element is offscreen or unrendered.
- Capture-phase click interception with `preventDefault` so links/buttons never fire while annotating, and the pin-first-then-comment flow where the click immediately opens the comment input.
- The hover enter animation (subtle scale + fade) and the `prefers-reduced-motion` opt-out.

Deliberately not copied: agentation's coordinate-only anchors (x% + document-y px, no selector; they cannot survive reflow) and app-notes' deepest-element-always targeting (a bare `<span>` or icon `<svg>` is a poor annotation target; Plannotator's semantic-target snap handles nested elements better for this use case).

## Two-runtime check
UI-only: changes live in `packages/ui` and `packages/editor`. No endpoints added or changed, so nothing to mirror in the Pi server. Both runtimes serve the same rebuilt HTML bundle. No new directories with Tailwind classes (bridge CSS is inline in the srcdoc injection), so no `@source` changes.

## Tests

`DOM_TESTS=1 bun test packages/ui packages/editor`:

```
 908 pass
 0 fail
 2922 expect() calls
Ran 908 tests across 94 files. [19.67s]
```

`bunx tsc --noEmit -p packages/ui/tsconfig.json`: clean, exit 0.

New DOM-gated coverage:
- `packages/ui/utils/inputMethod.test.ts`: pinpoint default resolution (HTML vs markdown vs saved prefs, garbage values, cookie round trips, no cross-surface leakage).
- `packages/ui/utils/htmlChrome.test.ts`: chrome-hidden first-run default, malformed cookie fallback, "user showed tools" and "user re-hid" states persisting across a fresh mount.
- `packages/editor/App.htmlHideTools.test.tsx`: full-App mounts covering the minimal first paint, the Show/Hide toggle, both persistence directions across remounts, and keyboard sidebar reachability while hidden.
- `packages/ui/components/html-viewer/htmlPinpointProtocol.test.tsx`: anchor DTO validation (size caps, malformed rejection), pinpoint selections opening the comment composer while drag selections keep the toolbar, and anchor propagation onto committed annotations.
- `packages/ui/components/html-viewer/srcdoc.test.ts`: bridge-level test against the real evaluated bridge script; a pinpoint click posts an anchored selection without mutating the hovered element, anchor-first find-and-mark, pin badge on text drift, and fail-closed stale anchors.

## Recording

A 17s Playwright headless recording drives the real built annotate UI (`bun run --cwd apps/review build && bun run build:hook`, then `plannotator annotate sample-page.html`) through: minimal first paint (no chrome), pinpoint hover highlighting across elements, click-to-pin, comment save, a second pin, and the Show tools reveal. See the follow-up comment for the video, or the local path in the review handoff.

### DIFF
--- a/packages/editor/App.htmlHideTools.test.tsx
+++ b/packages/editor/App.htmlHideTools.test.tsx
@@ -1,6 +1,11 @@
-import { afterEach, describe, expect, test } from "bun:test";
+import { afterAll, afterEach, describe, expect, test } from "bun:test";
 import React, { act } from "react";
 import { createRoot, type Root } from "react-dom/client";
+import {
+  resetStorageBackend,
+  setStorageBackend,
+  type StorageBackend,
+} from "@plannotator/ui/utils/storage";
 
 const hasDom = typeof document !== "undefined";
 
@@ -16,6 +21,22 @@ const originalFetch = globalThis.fetch;
 const originalEventSource = globalThis.EventSource;
 
 const RAW_HTML = "<h1>Rendered page</h1><p>Body copy.</p>";
+
+// In-memory storage backend (the codebase-standard persistence-test pattern):
+// keeps values across mounts within a test, so a remount simulates the next
+// session with the same cookies.
+const memory = new Map<string, string>();
+const memoryBackend: StorageBackend = {
+  getItem: (key) => memory.get(key) ?? null,
+  setItem: (key, value) => void memory.set(key, value),
+  removeItem: (key) => void memory.delete(key),
+};
+
+function seedAnnouncementsSeen(): void {
+  memory.set("plannotator-look-feel-announcement-seen", "2");
+  memory.set("plannotator-vim-mode-announcement-seen", "2");
+  memory.set("plannotator-plan-ai-announcement-seen", "1");
+}
 
 class SilentEventSource {
   static readonly CONNECTING = 0;
@@ -82,7 +103,7 @@ async function settle(): Promise<void> {
   });
 }
 
-async function mountHtmlAnnotate(): Promise<void> {
+async function mountHtmlAnnotate(expectButton: string): Promise<void> {
   globalThis.fetch = annotateFetch;
   // SAFETY: the App only uses EventSource's constructor, handlers, and close;
   // this test double implements those browser-facing members without I/O.
@@ -93,24 +114,53 @@ async function mountHtmlAnnotate(): Promise<void> {
   await act(async () => {
     root?.render(<App />);
   });
-  for (let attempt = 0; attempt < 20 && !findButton("Hide tools"); attempt += 1) {
+  for (let attempt = 0; attempt < 20 && !findButton(expectButton); attempt += 1) {
     await settle();
   }
 }
 
-afterEach(async () => {
+async function unmountHtmlAnnotate(): Promise<void> {
   if (root) await act(async () => root?.unmount());
   root = null;
   host?.remove();
   host = null;
+}
+
+afterEach(async () => {
+  await unmountHtmlAnnotate();
   globalThis.fetch = originalFetch;
   globalThis.EventSource = originalEventSource;
+  memory.clear();
+  resetStorageBackend();
   if (hasDom) document.body.replaceChildren();
 });
 
-describe.if(hasDom)("HTML annotate hide-tools", () => {
+afterAll(() => {
+  resetStorageBackend();
+});
+
+describe.if(hasDom)("HTML annotate chrome (minimal-first render + persistence)", () => {
+  test("first-ever HTML session opens minimal: all chrome hidden, Show tools discoverable", async () => {
+    setStorageBackend(memoryBackend);
+    seedAnnouncementsSeen();
+    await mountHtmlAnnotate("Show tools");
+
+    // Nothing the hide-tools toggle controls may be mounted on first paint.
+    expect(sidebarTabs()).toBeNull();
+    const show = findButton("Show tools");
+    if (!show) throw new Error('"Show tools" affordance did not render on first run');
+
+    await act(async () => show.click());
+    expect(sidebarTabs()).not.toBeNull();
+    expect(findButton("Hide tools")).not.toBeUndefined();
+  });
+
   test("hiding tools removes the collapsed sidebar tab flags, showing tools restores them", async () => {
-    await mountHtmlAnnotate();
+    setStorageBackend(memoryBackend);
+    seedAnnouncementsSeen();
+    // Simulate a returning user who previously showed tools.
+    memory.set("plannotator-html-chrome", JSON.stringify({ toolsHidden: false, sidebarOpen: false }));
+    await mountHtmlAnnotate("Hide tools");
 
     const strip = sidebarTabs();
     if (!strip) throw new Error("Collapsed sidebar tab flags did not render");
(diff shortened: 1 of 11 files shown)


backnotprop:
Headless Playwright recording (17s, 1360x850) demonstrating: minimal first paint with no chrome, pinpoint hover highlighting across elements, click-to-pin into the comment composer, two saved comments, and the Show tools reveal. Recorded against the real built annotate UI serving a sample dashboard page.

Local paths (not committed, kept out of git):
- /private/tmp/claude-501/-Users-ramos-plannotator-plannotator/34d56d5a-33a8-4878-92b2-3f4318214ae2/scratchpad/html-annotate-pinpoint-demo.mp4
- /private/tmp/claude-501/-Users-ramos-plannotator-plannotator/34d56d5a-33a8-4878-92b2-3f4318214ae2/scratchpad/html-annotate-pinpoint-demo.webm

Drag either file into this comment thread to embed it.

backnotprop:
**TLDR:** An adversarial review of this PR found two blocking defects (wrong-element anchor restoration on regenerated pages, and uncapped page-controlled selection text) plus three smaller ones. All five are fixed in 443983f7 and ccd4a791, re-verified adversarially, and the verdict at the current head is merge-safe. 911 tests pass.

*This review and the fixes were AI-assisted.*

The review attacked the trust boundary (hostile HTML inside the sandboxed iframe), anchor fail-closed correctness, the new cookies, and regressions, using a live harness against the real bridge script.

**Fixed:**
1. Anchors ending in behavioral attributes (role, href, aria-label) skipped the text check and could silently bind an annotation to the wrong element after a page regeneration, while suppressing the text-search fallback that used to get it right. Stability is now only the element's own #id or data-* rung, re-derived from the resolved element and compared whole, and the document-wide text search runs before the pin fallback. A missing or empty text snapshot is a rejection, not an exemption.
2. Selection text crossed the bridge uncapped: one pinpoint click on a large pre/table shipped a 600k-character string into drafts, feedback, and share URLs. Now truncated at 10k on both sides of the bridge.
3. The last-resort selector builder could return a selector already proven non-unique; it now returns no anchor, and those annotations restore by text search.
4. The minimal-first chrome never applied to a linked .html doc opened from a markdown session; restore moved to a surface-transition effect and was verified end-to-end.
5. Residual from re-review: the chrome writer ran once in the restore commit with pre-restore values; it now skips exactly that run, with a mutation-tested regression test.

**Verified safe:** postMessage source checks, anchor DTO validation and caps, no injection sink for htmlAnchor, share payloads drop anchors, cookies degrade to defaults on garbage, vim and teardown unchanged.

**Deliberately deferred:** pinpoint-only comment entry (product call), scroll-frame semantic-graph rebuild (perf), positional pin badge numbers, SVG camelCase selectors.

backnotprop:
Product decision on the deferred review question: pinpoint stays comment-only. A pinpoint click's job is to attach a note to an element; redline and quick-label remain available through drag selection. Working as intended, not a gap.

PR #1244, opened 2026-08-09
feat(ui): add Shift+1-4 shortcuts to switch annotation mode

### LINKED ISSUE #1242: [Feature Request] Keyboard shortcut / quick switcher for annotation modes (comment, redline, quick label, Ask AI)
## What I'd like

A keyboard-driven way to pick which kind of annotation I'm making, instead of reaching for the toolbar buttons.

Today, once the annotation toolbar is open on a selection, the keyboard covers some of it but not the choice itself:

- typing `A-Z` jumps straight into **Comment**
- `Alt+1-0` applies a **quick label**
- `Alt hold` / `Alt Alt` switches the **input method** (Select ↔ Pinpoint)

But the other actions on that toolbar — delete/redline, opening the quick-label picker, "looks good" — still need the mouse, and there's no single "switch to X" gesture. Same story on the code review side: toggling between **Comment**, **Suggested code**, and **Ask AI** in the comment editor is mouse-only (`Mod+Enter` submits, `Esc` cancels, but nothing selects the mode).

## Why

Mid-review I change my mind about what a selection needs — sometimes it's a comment, sometimes just a redline, sometimes a label, sometimes a question for the AI. Right now that means selection → move to the toolbar → click → back to typing, on every annotation. On a long plan or a big diff that adds up, and it breaks the flow that the existing type-to-comment shortcut sets up so nicely.

## Rough idea (very open to alternatives)

Either of these would solve it for me — I don't have a strong opinion on which:

- **A modifier per mode**, in the same spirit as the existing `Alt+1-0` labels — e.g. a key that goes straight to redline/delete, one that opens the quick-label picker, one that switches the code-review editor to Ask AI or Suggested code.
- **A small mode picker** on the open toolbar that you can cycle with a single key (Tab, or `Alt Alt`-style double-tap like the input-method switch already does), so the current mode is visible and one keystroke moves between comment / redline / quick label / Ask AI.

Whatever fits the existing shortcut registry best is fine by me — I mostly care that choosing the annotation type doesn't need the mouse.

Happy to be told this is already possible and I missed it, or that it overlaps with something already planned.

---

Plannotator 0.25.0, macOS, Claude Code.

galmadar:
Closes #1242.

Choosing the annotation mode in plan review was mouse-only — the four toolstrip buttons were the only way to move between **Markup**, **Comment**, **Redline** and **Label**. This adds `Shift+1-4`, following the toolstrip left to right.

## Why Shift+digits

`Alt+C` for comment would have been nicer to remember, but nearly every obvious key is already claimed:

| Candidate | Why not |
|---|---|
| bare letters | `annotationToolbar` binds `A-Z` to type-to-comment; vim also binds `c`/`d`/`m`/`t` |
| bare digits | quick-label picker (`1-0`), `Alt+1-0`, ImageAnnotator tools, vim `0` |
| `Mod+<digit>` | browser tab switching, not preventable |
| `Mod+Shift+<digit>` | `Cmd+Shift+3/4/5` are the macOS screenshot shortcuts, captured above the browser |
| `Mod+Shift+C/M/D` | DevTools, Chrome's profile switcher, bookmark-all-tabs |
| `Alt+<letter>` | `useInputMethodSwitch` flips Select↔Pinpoint on any `Alt` keydown, so every chord flickers the input method and writes storage twice |

`Shift+1-4` is free of OS and browser reservations on macOS, Windows and Linux, and can't double-fire with the bare-digit label picker: that binding declares no `Shift`, and the engine's `shiftMatches` (`core.ts:402`) rejects it once `shiftKey` is set.

That `Alt` behaviour looks like a pre-existing bug (`Alt+1-0` quick labels trigger the same flicker today), but it's independent of this change, so I've left it alone rather than widen the diff. Happy to open it separately if it's worth fixing.

## Implementation notes

- New scope `annotationMode.shortcuts.ts` under `plan-review/`, registered in `sharedPlanSurfaceShortcuts` so both the plan-review and annotate surfaces get it. Uses the existing `Annotations` section, so the docs page and section ordering are unchanged.
- No engine change needed: `getShortcutDigit` already falls back to `event.code`, so `Shift+2` matching token `2` works even though the layout reports `@`.
- Handlers write through the existing `handleEditorModeChange`, so the mode persists and every consumer re-renders from one call.
- The handlers share `canHandleDocumentChromeShortcut` and additionally gate on `toolstripVisible` — the five-condition check that decides whether the toolstrip renders, extracted into one memo now driving both the JSX and the guard. Without it the mode would change silently during plan diff, archive, markdown editing, goal setup, or with the HTML tools hidden, with no pill to report it. Suppression rather than a toast: in those states the mode is inapplicable anyway.
- `Alt` is excluded by the engine's own modifier matching (`requiresAlt === event.altKey`), so AltGr layouts can't fire these.
- Added the entry to `KeyboardShortcuts.tsx` by hand, since that panel isn't registry-driven.
- Added `aria-pressed` to the toolstrip buttons, so a keyboard-driven mode change is announced. Not `aria-label` — the labels collapse by `opacity`, not `display`, so they were already carrying the accessible name.

## Known limitation

In **HTML document mode** the content runs in a sandboxed iframe, and the bridge only forwards vim keys and one printable char — so these shortcuts don't fire while focus is inside the rendered document. Fixing it means a new bridge message plus plumbing through `useHtmlAnnotation` → `HtmlViewer` → `App`, which felt like it belonged in its own PR rather than doubling this one. Glad to follow up if you'd like it.

## Testing

`bun run typecheck` clean; `bun test` 3638 pass / 585 skip / 0 fail. Four tests in `packages/ui/shortcuts.test.ts` cover dispatch, the `event.code` fallback, inertness while `Alt` is held, the binds-once-on-every-plan-surface invariant, and that the bare-digit picker stays inert while `Shift` is held.

Verified by hand in `bun run dev:hook`: the active pill tracks each shortcut, each mode changes what a selection does, the mode survives a reload, and the shortcuts stay inert while typing in a comment.

---

### DIFF
--- a/packages/editor/App.tsx
+++ b/packages/editor/App.tsx
@@ -117,6 +117,7 @@ import { modKey } from '@plannotator/ui/utils/platform';
 import {
   annotateSidebarShortcuts,
   useAnnotateSidebarShortcuts,
+  useAnnotationModeShortcuts,
   useDocumentViewShortcuts,
   useDoubleTapShortcuts,
 } from '@plannotator/ui/shortcuts';
@@ -2654,6 +2655,39 @@ const App: React.FC = () => {
   // Alt/Option key: hold to temporarily switch, double-tap to toggle
   useInputMethodSwitch(effectiveInputMethod, handleInputMethodChange);
 
+  // Gates both the toolstrip's own render and its shortcuts, so a mode can never
+  // change with no visible pill to report it.
+  const toolstripVisible = useMemo(
+    () =>
+      !goalSetupMode && !isPlanDiffActive && !archive.archiveMode && !isEditingMarkdown && !htmlChromeHidden
+      && (!isCompactTouchLayout || !(annotateSource === 'folder' && !markdown && !linkedDocHook.isActive)),
+    [
+      annotateSource,
+      archive.archiveMode,
+      goalSetupMode,
+      htmlChromeHidden,
+      isCompactTouchLayout,
+      isEditingMarkdown,
+      isPlanDiffActive,
+      linkedDocHook.isActive,
+      markdown,
+    ],
+  );
+
+  const canHandleAnnotationModeShortcut = useCallback(
+    (event: KeyboardEvent) => toolstripVisible && canHandleDocumentChromeShortcut(event),
+    [canHandleDocumentChromeShortcut, toolstripVisible],
+  );
+
+  useAnnotationModeShortcuts({
+    handlers: {
+      selectMarkupMode: { when: canHandleAnnotationModeShortcut, handle: () => handleEditorModeChange('selection') },
+      selectCommentMode: { when: canHandleAnnotationModeShortcut, handle: () => handleEditorModeChange('comment') },
+      selectRedlineMode: { when: canHandleAnnotationModeShortcut, handle: () => handleEditorModeChange('redline') },
+      selectQuickLabelMode: { when: canHandleAnnotationModeShortcut, handle: () => handleEditorModeChange('quickLabel') },
+    },
+  });
+
   // Check if we're in API mode (served from Bun hook server)
   // Skip if we loaded from a shared URL
   useEffect(() => {
@@ -5083,8 +5117,7 @@ const App: React.FC = () => {
                   comment/markup mode). Hidden during plan diff, and on HTML surfaces
                   when the header's "Hide tools" toggle is on (leaving the rendered HTML
                   free of overlay controls). On HTML it floats top-left over the doc. */}
-              {!goalSetupMode && !isPlanDiffActive && !archive.archiveMode && !isEditingMarkdown && !htmlChromeHidden &&
-                (!isCompactTouchLayout || !(annotateSource === 'folder' && !markdown && !linkedDocHook.isActive)) && (
+              {toolstripVisible && (
                 <div
                   data-print-hide
                   className={isHtmlSurface
--- a/packages/editor/shortcuts.ts
+++ b/packages/editor/shortcuts.ts
@@ -1,4 +1,5 @@
 import {
+  annotationModeShortcuts,
   annotationPanelShortcuts,
   annotationToolbarShortcuts,
   annotateSidebarShortcuts,
@@ -82,6 +83,7 @@ const annotateEditorSettingsShortcuts = defineShortcutScope({
 const sharedPlanSurfaceShortcuts = [
   documentViewShortcuts,
   inputMethodShortcuts,
+  annotationModeShortcuts,
   annotationToolbarShortcuts,
   viewerShortcuts,
   vimSelectionShortcuts,
--- a/packages/ui/components/AnnotationToolstrip.tsx
+++ b/packages/ui/components/AnnotationToolstrip.tsx
@@ -321,6 +321,7 @@ const ToolstripButton: React.FC<{
       onClick={onClick}
       onMouseEnter={() => setHovered(true)}
       onMouseLeave={() => setHovered(false)}
+      aria-pressed={active}
       className={`relative flex items-center h-7 rounded-md overflow-hidden ${colorClass}`}
       style={{ width: currentWidth, transition }}
     >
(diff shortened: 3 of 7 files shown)


backnotprop:
TLDR: The implementation is exactly right, but the key choice breaks on macOS: Cmd+Shift+3 and Cmd+Shift+4 are the system screenshot shortcuts, so two of the four modes are unreachable there and take a screenshot of the plan instead. We propose Shift+1 through Shift+4. AI-assisted review.

First, what's good, because most of this PR is: it is the cleanest instance of the shortcut convention in the repo. Declarative scope in the right subfolder, composed onto both plan surfaces so it reaches the help modal and the auto-generated docs page, real behavior tests, no raw keydown listeners, no registry collisions, correct focus guards, and you even updated the hand-maintained in-app modal. Thank you for building it the way the system intends.

The blocker: on macOS the OS captures Cmd+Shift+3/4 above the browser (screenshots, on by default), so the keydown never reaches the page. A macOS reviewer following the shortcut sheet gets a shutter sound and a PNG on their Desktop, and the mode does not change. The Ctrl+Shift variant technically also matches Mod on mac, but nothing surfaces that, so it cannot be the answer.

Proposed replacement: Shift+1 through Shift+4. We verified: free of OS and browser reservations on macOS, Windows, and Linux; cannot double-fire with the bare-digit quick-label picker (that binding requires shift to be up); avoids the Alt family entirely (any bare Alt keydown flips the drag/pinpoint input method and would flicker the toolstrip); token-valid with zero engine changes; and it keeps your 1-4 toolstrip order mnemonic. Everything else in the PR survives unchanged except the test event fixtures.

Worth folding into the same pass:
- Rebase on main: we since landed the same guard extraction as canHandleDocumentChromeShortcut, so canHandlePlanShortcut is now redundant and the conflict resolves by using ours. Post-rebase your diff shrinks to the scope file, the index export, the surface array, the modal line, and the tests.
- The mode change is silent (but persists) when the toolstrip is hidden, e.g. during plan diff view. Gate the handler on toolstrip visibility or show a toast.
- Add the cross-surface registry test pattern already in shortcuts.test.ts (the binds-once-on-every-plan-surface loop) so the scope cannot silently drop off a surface later.
- On the AnnotationToolstrip change: keep aria-pressed (genuine improvement), but the aria-label is redundant since opacity-zero text stays in the accessibility tree, and the comment claiming otherwise should not merge as written.

We want this in the next release. If a rebase is a pain, say so and we can push the mechanical parts to your branch.

galmadar:
@backnotprop thanks — the macOS blocker is real and I missed it. I checked browser and app reservations when I picked the keys, not the OS layer above them. Taking the whole review; pushed as `b4b72dae`.

**`Shift+1-4`.** Verified your reasoning against the code before swapping: the bare-digit picker binds `1-0`, and `shiftMatches` in `core.ts:402` rejects a shift-less binding once `shiftKey` is set, so no double-fire. `getShortcutDigit` already falls back to `event.code`, so no engine change — `Shift+2` reporting `@` still matches token `2`.

**Rebased.** `canHandleDocumentChromeShortcut` is yours, mine is gone. Diff is now 7 files, +169/−2.

**Toolstrip gate.** Rather than duplicate the five-condition render check, I extracted it into a `toolstripVisible` memo and pointed both the JSX and the shortcut guard at it, so they can't drift. Suppression rather than a toast — during plan diff and markdown editing the mode is inapplicable anyway, and a toast for "the thing you can't see just changed" seemed worse than nothing happening. Say the word if you'd rather have the toast.

**ARIA.** You're right, `opacity: 0` keeps the label in the accessibility tree. Dropped the `aria-label` and the comment; kept `aria-pressed`.

**Tests.** Added the binds-once-on-every-plan-surface loop, plus one asserting the bare-digit picker stays inert while Shift is held. The surface test alone wouldn't catch that regression — `1-0` and `Shift+1` are different binding strings, so a claimant check by string equality reads them as unrelated.

`bun run typecheck` clean, `bun test` 3638 pass / 0 fail.

One thing I have **not** re-done since the swap: hands-on verification in a browser. The tests can't prove the one thing that actually matters here — that `Shift+3` reaches the page on macOS instead of firing a screenshot. I'll confirm that before you merge unless you get there first.

PR #1268, opened 2026-08-11
feat(review): add optional CallDiff call-flow analysis

backnotprop:
## Summary

- add optional Call flow analysis powered by a pinned CallDiff 0.4.1 runtime
- expose a snapshot-bound Dock and per-file Lens that share one result, plus a dedicated raw tree view and native Plannotator annotations
- make Semantic changes and Call flow independently configurable, with a one-time welcome dialog that follows existing review dialog conventions and credits the CallDiff creator
- add Bun and Pi API parity, managed installers and uninstall cleanup, user documentation, and a detailed integration ADR

## Runtime and safety

- runs CallDiff in an isolated Node 22 worker with 45 second, 512 MB, and 12 MB output limits
- computes exact Git snapshots through a temporary shared clone and synthetic commits without mutating the reviewed index or worktree
- pins the CallDiff source commit, integrity hashes, and 21 parser grammars; review-time analysis is offline
- binds every result to the current review snapshot so stale requests cannot populate the Dock or Lens

## Verification

- focused Call flow, config, dialog, install, and uninstall tests: 225 passed, 8 platform skips
- dialog DOM tests: 4 passed
- full repository suite: 3,186 passed and 457 skipped; two existing five-second API guard tests timed out only under full-suite load
- isolated API guard rerun: 9 passed
- bun run typecheck
- bun run build:review
- bun run build:hook
- managed CallDiff runtime smoke against a temporary Git repository
- Bun review API end-to-end smoke for advertise, analyze, snapshot binding, and disable behavior


backnotprop:
**TLDR:** Impressive integration and the review-time surface is genuinely well-defended (snapshot immutability, staleness 409s, cancellation, bounds, and Bun/Pi endpoint parity all held under adversarial tracing). Two findings are merge-blocking: the SHA-512 integrity check runs only after npm has already downloaded and executed the unverified archive, and patch-backed views fail for any changeset containing a binary file. Six more should-fix findings below, ranked. Three independent adversarial reviews (security/runtime, server correctness, UI/spec), findings self-refuted before reporting; the two blockers were independently re-verified against the branch. AI-assisted.

## Merge-blocking

**1. The SHA-512 pin does not protect the install (security, CONFIRMED).** `installCallFlowRuntime` (packages/shared/call-flow.ts:294+) writes a package.json pointing at the GitHub archive URL and runs `npm install` with no committed lockfile and no `--ignore-scripts`. On first install npm has nothing to verify against (TOFU), downloads the archive, and runs its lifecycle scripts as the user. `CALLDIFF_SOURCE_INTEGRITY` (:24) is only compared at :252, post-build, against the package-lock that npm itself generated from whatever bytes arrived, so a substituted archive gets code execution before verification and can trivially satisfy the later check by writing the expected constant into the lock. The 24 grammar/tooling deps are pinned by version spec only, no integrity hashes, and their node-gyp scripts run. Fix shape: download the tarball directly, verify SHA-512 against the pin FIRST, install from the verified local file with a shipped lockfile via `npm ci --ignore-scripts`, then explicit grammar rebuilds.

**2. Binary files break every patch-backed view (server, CONFIRMED).** The synthetic-commit path applies the visible patch with `git apply --cached --binary` (call-flow.ts:463), but the visible patches are generated in review-core.ts without `--binary`/`--full-index`, so binary changes arrive as "Binary files differ" stubs that `git apply` rejects. Any review touching an image makes since-base/uncommitted/staged/unstaged return `error` and sit in the 30s cooldown. The internal staged leg at call-flow.ts:541 already uses `--binary --full-index`, so only the visible-patch legs miss it.

## Should fix

**3. Poisoned cache on mid-run repo change (server, mechanism CONFIRMED).** The ok-cache write (call-flow.ts:691) happens before the endpoint's after-fingerprint check (review.ts:908-911). A repo change during the run 409s correctly, but the wrong result is already cached under the unchanged snapshot key; a change-then-revert cycle (reset --hard, stash pop) later serves it as fresh. Move the after-check into the service or skip caching when it fails.

**4. Dock/Lens offer Open/Comment on nodes the review cannot service (UI, CONFIRMED).** CallFlowTreeView.tsx:47-50 gates on node status/location, never on diff membership. Trees reach files the patch never touched: those rows render enabled buttons that silently no-op (App.tsx:531-533, 578-590). Worse, changed-but-out-of-hunk lines open the composer and create annotations that never render inline and, in PR mode, can fail the whole hosted submission when GitHub rejects the out-of-diff line. Gate the affordances on patch membership.

**5. `error` is a terminal state (UI, CONFIRMED).** useCallFlowAnalysis.ts:16-39 refetches only on snapshot/available change; one transient failure (45s timeout) blanks the feature until a view switch. No retry affordance exists, so the server's 30s failure cooldown is dead code from the client's perspective. Spec requires "explains the state and recovery action" for every state.

**6. Direct `/api/call-flow` on unsupported views returns `error` not `unsupported` (server, CONFIRMED, both runtimes).** The advert correctly says unsupported for All Files/jj/GitButler/P4, but the endpoint lacks the guard: createCallFlowSnapshotPlan throws and maps to `analysis-failed`, entering the failure cooldown. Advert and endpoint disagree on the spec's distinct domain states. Related: `/api/review-analysis` has no superseded guard (unlike `/api/diff/switch`), so a late response can re-advertise call flow as available on a newer unsupported view, and the alarming `!` badge shows instead of the calm unsupported copy.

**7. Bun/Pi drift in the PR-checkout fallback (server, CONFIRMED drift).** Bun's ensurePRLocalCwd falls back to the launch repo on pool failure (review.ts:872, 357-375) and analyzes there; Pi's ensurePRCallFlowCwd returns null and reports `unavailable` (serverReview.ts:722-731). Same session shape, different outcomes; Bun's fallback is only correct when the launch repo happens to contain the PR objects. Align on Pi's behavior.

**8. Relative `PLANNOTATOR_CALLDIFF_PATH` resolves against the reviewed repo (security, CONFIRMED path).** call-flow.ts:228-232 `resolve()`s a relative override against cwd and skips every integrity check. A dev with a relative override in their profile who reviews a hostile repo containing a matching directory executes its code. Require an absolute path.

## Worth knowing (not blocking)

- Test gaps where it matters: useCallFlowAnalysis (shared-result/abort/error-latch) has no tests; CallFlowService's mutex/cooldown/cache/getAdvert are untested; call-flow.test.ts covers 2 of the 9 snapshot plans. The spec's own verification list (stale rejection, cancellation, one-shared-result) is largely unpinned.
- The worker is unsandboxed: "no network during review" is an npm env hint, not enforcement; the 512MB flag caps V8 only, not Tree-sitter's native allocations. Defense-in-depth erosion given finding 1.
- Windows kill is not tree-safe (call-flow.ts:362-368): grandchildren can survive the timeout, and a cleanup rmSync throw escapes the finally as a 500 plus a leaked temp clone. Wrap cleanup in try/catch.
- getAdvert spawns `node --version` and re-reads package-lock.json synchronously on every diff/switch/PR-switch while enabled; cache it like the semantic-diff probe. CallFlowFileBadge runs a full up-to-5000-node traversal per badge per render, unmemoized.
- Optimistic toggle: if the POST fails after the cookie flips, the Dock misdiagnoses it as "runtime not available." failureCache is unbounded (minor).
- New-user first session now shows five sequential dialogs (guide intro, look-and-feel, review setup, analysis intro, edit mode). Follows the chain convention and never stacks, but it is a heavy first run.

## Verified clean under attack (for the record)

Snapshot materialization never touches the user's repo (temp shared clone, read-tree/commit-tree, cleanup on all paths); ref/option injection prevented (`--end-of-options`, hex validation); all 9 spec table mappings match the code including the subtle unstaged composition; mid-execution change detection is real (before/after fingerprints + snapshot rechecks, both runtimes); staleness/cancellation races converge; settings toggles are genuinely independent with deep-merge persistence; output bounds enforced and reject-not-truncate; uninstall preserves unknown vendor entries; skip flags honored across all three installers; vendor.sh and export maps cover the new shared modules; one shared analysis result powers Dock, Lens, and badges with no N-request storm from virtualized headers; dialog chain cannot stack; deliberate non-goals respected.

backnotprop:
Addressed in `a68a2cda`.

## Blockers

1. **Verify before execute:** replaced the generated `npm install` path with a repository-owned runtime manifest and committed lockfile. The installer now downloads the pinned CallDiff archive itself, bounds it to 20 MB, verifies the repository-pinned SHA-512 before writing/passing any bytes to npm, then runs `npm ci --ignore-scripts` from that verified local archive. Registry dependencies are integrity-locked. Native Tree-sitter/grammar lifecycle steps are invoked explicitly, followed by an explicit CallDiff `tsc` build. Installation happens in a sibling temp directory and is swapped in only after the package entry, grammar set, revision, and unchanged committed lock all validate.

2. **Binary patch-backed views:** left the visible/UI patch pipeline unchanged and added a dedicated CallDiff materialization patch path with `--binary --full-index` for `since-base`, `uncommitted`, `staged`, and `unstaged`, including strict binary-capable untracked-file patches. Added coverage for both changed tracked images and new untracked images. A real CallDiff smoke run over a TypeScript edit plus changed PNG completed successfully.

## Should-fix

3. Snapshot revalidation now runs inside `CallFlowService` before an `ok` result enters the success cache. A failed after-check returns `stale` and is never cached; change/revert poisoning is regression-tested.

4. Dock and Lens Open/Comment actions are gated on the complete node range being present in one hunk on the correct old/new side. Out-of-patch navigation is disabled with a direct tooltip and the native annotation action is omitted. The ordinary Plannotator annotation path remains the only comment path.

5. Error state now has **Retry analysis**. Retry remains subject to the server's bounded 30-second failure cooldown; after expiry it starts a fresh request. Client retry and request-abort behavior are covered.

6. Direct `/api/call-flow` now returns `unsupported` for unsupported views before snapshot planning/runtime execution. `/api/review-analysis` has monotonic request + view + PR-scope guards in Bun and Pi; persistence occurs only after the final guard, so superseded requests write nothing. Both runtime endpoints have overlap/unsupported tests.

7. Bun now matches Pi for CallDiff PR checkout failure: pool `ensure()` failure returns `checkout-unavailable`; it never falls back to the launch repository.

8. Relative `PLANNOTATOR_CALLDIFF_PATH` values are rejected. Only absolute overrides are accepted.

## Batched hardening

- CallFlowService tests now pin shared in-flight results, supersession/abort, cooldown, no-cache-on-stale, unsupported-before-probe, and bounded runtime-probe caching.
- Runtime probe is cached for 30 seconds per review service; success/failure caches are bounded.
- Temp clone/install cleanup is best-effort for Windows `EBUSY`; Windows cancellation kills the full child process tree.
- Call-flow badge tree filtering/traversal is memoized.
- Failed `/api/review-analysis` writes roll both analysis settings back to the last server-confirmed values.
- Pi vendoring ships the exact same committed runtime manifest + lockfile.

## Verification

- Clean isolated managed-runtime install: passed; verified archive removed after install and built entry present.
- Real CallDiff binary-change smoke analysis: passed.
- `bun run typecheck`: passed.
- Production builds: review, hook, and Pi passed.
- Focused regression set: 93 passed (including DOM retry/abort coverage).
- Full `bun test`: 3,198 passed; two unrelated 5-second API route tests timed out only under full-suite load. Immediate isolated rerun of that file: 9/9 passed.
- `git diff --check`: passed.

backnotprop:
**TLDR:** Verification of a68a2cda complete: all eight review findings are genuinely fixed in both runtimes, with regression tests that encode the failure modes. The integrity pin was independently verified against the live upstream artifact (downloaded and hashed it: match). 306 tests pass, 0 fail across the affected suites. Recommendation: merge. Five minor follow-ups noted, none blocking. AI-assisted.

Per-finding verdicts (every claim in the response comment was verified in code, none taken on trust):

1. **Integrity flow: FIXED.** Tarball is streamed, bounded at 20MB, SHA-512-checked against the repo-committed pin before a single byte reaches disk or npm. Committed lockfile (46 entries, all with sha512 integrity, zero drift vs the pinned grammar specs, pin equality CI-enforced), then `npm ci --ignore-scripts`, then explicit scoped rebuilds over already-verified bytes, then explicit tsc, post-install validation, atomic swap. The only pre-verification execution is `node --version`.
2. **Binary patches: FIXED.** Separate `--binary --full-index` materialization patch feeds the CallDiff apply path in both runtimes; visible patch byte-identical (default-false flag, test-asserted with both a changed tracked and a new untracked binary).
3. **Cache poisoning: FIXED.** Post-analysis snapshot verification gates the cache write; failed revalidation is never cached (test-pinned), both runtimes.
4. **Out-of-diff affordances: FIXED.** Nodes stay rendered (spec-compliant), Open disabled with tooltip, comment button omitted; correct old/new side hunk-range checking.
5. **Retry: FIXED** with an honestly-disclosed caveat: retries inside the 30s failure cooldown return the cached error, so the button is a visual no-op for up to 30s. A disabled state or countdown would be more honest (follow-up).
6. **Unsupported/superseded guards: FIXED, both runtimes**, exercised by dual-runtime endpoint tests booting both real servers.
7. **PR-checkout parity: FIXED.** Bun now mirrors Pi's checkout-unavailable behavior; the residual fallback can only resolve to the PR checkout, never the launch repo.
8. **Relative override: FIXED.** Non-absolute PLANNOTATOR_CALLDIFF_PATH rejected before any cwd resolution (test-pinned).

Non-blocking batch all present (EBUSY-tolerant cleanup, Windows full-tree kill, probe caching, badge memoization, settings revert with abort guard, Pi vendoring byte-identical). New follow-ups from the delta sweep, all minor: unbounded materialization-patch memory for very large binaries (a byte cap would be cheap insurance); patch generation and fingerprints run even on cache hits; a resolveRuntime rejection escapes as a raw 500; single-hunk membership is conservative for nodes spanning adjacent hunks; the cooldown-masked retry UX above. Soft spot: since-base and staged snapshot plans still lack direct unit tests (uncommitted and unstaged have them).

jsummerfield:
@backnotprop Thanks for this, excited to try it out but all the PRs I've tested it with fail the 12MB output limit. Would you be open to raising this, or making it configurable? I can send a PR if you agree.

PR #1269, opened 2026-08-11
fix(pi): never touch Pi's system prompt; phase framing as conversation messages (#922)

### LINKED ISSUE #922: Pi extension drops base system prompt, causing AGENTS.md to not be in the context
# What's the issue

When the plannotator extension is started, it replace the system prompt with it's own.
The default pi system prompt is not injected anymore. This means that precisions on how to navigate the repo, and available skill are not included.

backnotprop:
## TLDR

Fixes #922, both halves. The Pi extension now **never returns or modifies `systemPrompt`**: Pi's base prompt (AGENTS.md context, skills catalog, tools guidance, `--append-system-prompt` text, cwd) always reaches the model untouched, which fixes the missing-AGENTS.md report by construction. Phase framing is delivered **once per phase entry** as a hidden conversation message, and execution progress rides in small per-turn todo-status messages, so prompt-cache invalidation (WinPooh32's report) reduces to conversation-suffix appends plus one history adjustment per phase transition. This message-based approach was suggested by @Karrq on this PR; adopted outright.

**BREAKING (Pi users):**
- The plan-mode toggle command is renamed from `/plannotator` to `/plannotator-plan-mode`. No alias; the old name is gone. `Ctrl+Alt+P`, `pi --plan`, and `/plannotator-review` / `/plannotator-annotate` / `/plannotator-last` are unchanged.
- The `phases.*.systemPrompt` config key is retired. Phase templates now live in `phases.*.instructions` and are delivered as a phase-entry message, not a system prompt. Old `systemPrompt` keys are **ignored, with a session-start warning** naming the config file and the new key. Template variables (`${planFilePath}`, `${todoList}`, counts, `${phase}`) still work and render once at phase entry.

## Why

Two reports on #922:

- paullegranddc: starting the extension replaces Pi's system prompt, so AGENTS.md, repo navigation guidance, and available skills are missing. In Pi, a `before_agent_start` `{ systemPrompt }` return replaces the entire system prompt for the run, and AGENTS.md/skills only ever reach the model inside that prompt.
- WinPooh32: the prompt manipulation causes prompt-cache misses. Pi places Anthropic `cache_control` breakpoints on the system prompt and conversation tail; the cache is a strict prefix, so any system-prompt change re-processes the whole conversation uncached. The old executing template interpolated the live todo list, changing the system prompt as steps completed.

An earlier revision of this PR composed the base prompt into the replacement via a `${baseSystemPrompt}` variable. Karrq suggested the stronger invariant: leave the system prompt alone entirely and put the context in the conversation. That is what this revision does; the `${baseSystemPrompt}` variable is gone.

## How it works now

- `before_agent_start` never returns `systemPrompt`. On the **first prompt after a phase entry** it returns a hidden `plannotator-framing` message rendered from the phase `instructions` template (the 3.6k planning workflow is injected once, not per prompt). The delivery latch (`framingDelivered`) is persisted with the rest of the extension state; deny/resubmit rounds never leave the planning phase, so they inject nothing further.
- During execution, later prompts carry only a small `plannotator-context` todo-status snapshot (X/Y complete, remaining steps, and a one-line restatement of the `[DONE:n]` marker convention), appended at the conversation tail where the cache extends anyway. Executing `instructions` that do not reference `${todoList}` get the entry-time snapshot appended automatically so the first executing prompt always carries the checklist.
- The shipped executing instructions open by superseding the planning rules ("the planning phase is over... planning-phase restrictions no longer apply"), and a phase-aware context filter enforces coherence: while idle, everything Plannotator injected is filtered from model context (as before); during a phase, only the newest framing for the **current** phase survives, so stale planning rules cannot leak into execution and re-entered phases do not accumulate duplicate 3.6k framings. The filter is deterministic within a phase, so it never perturbs the cached prefix mid-phase.

### Latch resilience (review follow-ups)

The latch assumes the delivered framing stays in conversation history; two Pi mechanisms violate that, and both are now handled:

- **Compaction:** custom messages are ordinary compactable messages and auto-compaction defaults on, so a long session can summarize the framing away while the latch stays closed. The extension subscribes to `session_compact` (fired after compaction) and reopens the latch during an active phase; the next prompt re-delivers the framing. If the framing survived in the kept tail, the newest-framing-wins filter drops the old copy, so re-delivery never duplicates. Between a compaction and re-delivery, the todo message's `[DONE:n]` line keeps the completion protocol alive.
- **`/tree` branch navigation:** a branch switch changes the active path out from under the latch; the new path can carry a different phase or lack the framing that was delivered on the abandoned branch. The extension subscribes to `session_tree` and re-derives phase, latch, and checklist state from the new active path using the same restore logic `session_start` uses (extracted as `resyncPhaseFromSession`, now reading `getBranch()`, the root-to-leaf active path, instead of the whole append-only entry file). A path with no plannotator state at all means idle, which also hands back phase tools and restores the saved model/thinking state. Framing delivery and its `persistState` write are adjacent entries, so path-scoped state and the framing message travel together across branches.
- A missing bundled `plannotator.json` (packaging regression) now warns at session start instead of silently producing a rule-less planning phase.

## Files

- `apps/pi-extension/index.ts`: rewritten `before_agent_start` (message-only, once-per-entry latch `framingDelivered`, persisted in extension state), phase-aware context filter, `session_compact` / `session_tree` resilience handlers, shared `resyncPhaseFromSession` restore, command rename.
- `apps/pi-extension/config.ts`: `PhaseProfile.instructions` replaces `systemPrompt`; obsolete-key and missing-bundled-config warnings in config loading.
- `apps/pi-extension/plannotator.json`: shipped templates moved to `instructions`; executing template is a phase-entry message with an entry todo snapshot and the supersede line.
- `apps/pi-extension/README.md`: config docs rewritten for the message model; command rename.
- Marketing `installation.md` and the `plannotator-meets-pi` blog post: command rename.
- Tests: `phase-prompts.test.ts` (19 tests) drives the real extension through `session_start`, plan-mode toggles, and the `before_agent_start` / `context` / `session_compact` / `session_tree` events: no result ever carries `systemPrompt`, framing delivered exactly once per entry and re-delivered on re-entry and after compaction (exactly once), per-turn todo messages with changing content, resume-with-latch does not re-deliver, tree switches resync phase and latch from the new active path, the `persistState` payload is pinned on both sides of the latch, null instructions, custom templates, obsolete-key warning, and the idle/phase filter semantics. `config.test.ts` covers the `instructions` rename, shipped-template contract, and the obsolete `systemPrompt` warning. Mutation-verified: removing the once-only latch, returning a `systemPrompt`, breaking the filter anchor, disabling the compaction reset, hardcoding `framingDelivered: false` in `persistState`, and no-oping the tree resync each fail the suite.

## Validation

- `bun test` in `apps/pi-extension`: 211 pass, 0 fail
- `bun test tests/parity`: 7 pass, 0 fail
- `bun run typecheck`: clean

---

_This PR was written with AI assistance._


Karrq:
Hey, just wanted to say that I noticed this same issue today. I have a suggestion for more cache-friendliness: avoid changing the system prompt entirely and instead append the context as an extension of the user message (or on the subsequent request from phase change), reducing cache-busting to a suffix of the conversation.

backnotprop:
on it, about to ship

backnotprop:
Merged, and we went further than the original fix: your suggestion was adopted outright. The extension now never touches the system prompt in any phase; phase framing is delivered as a one-time conversation message at phase entry (with re-delivery on Pi compaction and tree navigation so long sessions stay framed), and the live todo status rides as small per-turn messages. Cache-busting is reduced to pure conversation-suffix appends, zero system-prompt invalidations. Thanks @Karrq for pushing it to the cleaner design. Ships in the next release. AI-assisted.

PR #1270, opened 2026-08-11
feat(review): make the CallDiff runtime a strictly opt-in, in-UI install

backnotprop:
## TL;DR

Call Flow is now a strictly opt-in, selective install. Plannotator installs nothing by default. On first use, it inspects the current review's changed file extensions and offers the pruned CallDiff core plus only the grammar packs that review needs. A Python-only review, for example, installs about 4.93 MiB measured instead of the previous 784 MB all-language runtime.

This keeps the UI-driven install from the original PR, but replaces the eager 21-grammar layout that diverged from [CallDiff's lazy language model](https://github.com/tanishqkancharla/calldiff).

## User experience

- Enabling **Call Flow** with no runtime installed shows one install action. The disclosure names the detected languages, gives the rounded installed-size total, and calls out Node.js 22+.
- The primary flow has no language picker: the current patch is authoritative, and the server installs core plus exactly its missing language packs.
- If a later review introduces another language, installed languages are still analyzed. The Dock shows a quiet `N files skipped: <language> support not installed (~size) · Install` state and reruns the same snapshot after installation.
- The Dock's **Languages** detail lists all supported families, installed state, measured per-pack size, and install-ahead actions. There is no per-pack removal in v1; uninstall still removes the complete managed runtime.
- The per-file Lens marks skipped-language files and opens the Dock. Installed-language Lens and Dock views continue sharing one analysis result, including full inferred entry trees and native Plannotator annotation actions.
- Pack errors remain visible beside the affected skipped files and language row with an explicit Retry action.

## Selective runtime and supply-chain boundary

- Core contains CallDiff 0.4.1 at the pinned upstream commit, Tree-sitter, and the JavaScript/TypeScript grammars.
- Every one of the 19 optional grammar packs has its own repository-committed manifest and lockfile with exact versions and registry integrity hashes.
- Pack tarballs are downloaded first, SHA-512 checked against the committed lock, and only verified bytes are exposed to a fresh private npm cache. `npm ci` then runs offline with `--ignore-scripts`; the pinned grammar is rebuilt explicitly afterward.
- Core and packs are pruned fail-closed before publication: generated parser sources, foreign-platform prebuilds, nested dependencies, node-gyp intermediates, TypeScript, and build-only packages are removed. Only the current-platform prebuild or final locally compiled `.node` addon survives.
- The pruned candidate must load through CallDiff before its atomic swap/lock marker. Validation and review workers both put an npm blocker first on `PATH`, set npm offline/ignore-scripts flags, and point `CALLDIFF_GRAMMAR_CACHE` only at Plannotator's managed store.
- Runtime preflight and the worker verify exact CallDiff, Tree-sitter, core-grammar, and relevant optional-grammar versions before importing CallDiff. Missing support is a server-authored preflight/partial-analysis state, not an upstream npm surprise.
- A core repair preserves only optional packs whose exact marker, version, and native payload still validate against the candidate core.

## Server and client behavior

Both Bun and Pi expose the same contract:

- `POST /api/call-flow/install` accepts `{ languageIds?: [...] }`. Omission uses the current review plan; explicit Languages-list requests prepend only core when needed. One coordinator deduplicates current/completed targets, queues new targets onto the active flight, performs Node preflight, and enforces the same-origin guard.
- `GET /api/call-flow/install-status` reports target-aware `downloading → verifying → installing-deps → building` progress. Transient non-2xx polls retain the running state and retry.
- `GET /api/review-analysis` is the read-only advert refresh. It observes but never increments the settings mutation epoch; refreshes on either side of an active toggle write yield as superseded. Install completion invalidates the runtime probe and successful/failure analysis caches, then retries advert reconciliation until the same session reruns analysis.
- Invalid `PLANNOTATOR_CALLDIFF_PATH` states never show a managed install funnel that cannot repair the override. Missing/old Node enters the funnel so the coordinator's distinct preflight recovery message remains reachable.

The Pi vendoring script includes the language catalog, pack locks, and all pack manifests/lockfiles.

## Measured footprint

Measured from real verified installs on macOS arm64, Node 24.15.0:

| Artifact | Before prune | Published |
| --- | ---: | ---: |
| Core | 94,408,476 bytes (90.03 MiB) | 4,649,257 bytes (4.43 MiB) |
| Python pack | 7,513,587 bytes (7.17 MiB) | 527,081 bytes (0.50 MiB) |
| Core + all 19 packs | — | 71,925,356 bytes (68.59 MiB) |

All 19 packs were installed through the final download → SHA-512 verify → offline `npm ci` → build/prune/load pipeline. Every pack resolved as installed, including locally compiled Kotlin and the previously 200 MB OCaml package.

## Review remediation folded in

- Read-only advert refresh cannot supersede a settings write or later overwrite it with stale capabilities.
- Advert reconciliation retries transient failures instead of leaving `Install complete. Starting…` stranded.
- Invalid overrides do not get a false install funnel.
- Failed Node preflight cannot leak old review targets into a later retry.
- Delayed duplicate POSTs cannot requeue a target already installing or completed in the active flight.
- Runtime validation cannot invoke CallDiff's lazy npm fallback.
- Candidate pruning is strict and verified before atomic publication.
- Optional-pack progress now corresponds to real download, verification, dependency installation, and build work.

## Verification

- Real isolated core install, core repair, Python/Go install, and final all-19-pack install: passed. Core repair preserved and revalidated installed packs.
- Targeted dual-runtime endpoint/coordinator tests: 30 passed.
- Targeted Call Flow DOM + native annotation tests: 14 passed.
- Installer tests: 121 passed, 8 Windows-gated skipped.
- Full repository suite: 3,256 passed, 468 skipped. The 14 failures are the same pre-existing environment-dependent semantic-diff/GitButler tests observed before this change; no Call Flow test failed.
- `bun run typecheck`: passed, including Pi vendoring/parity.
- `bun run build:pi`: passed.
- `bun run build:marketing`: passed.
- `git diff --check origin/main --`: passed.

Tests mock installation; CI never downloads or builds the real runtime. Real-install verification used isolated temporary data directories and did not touch `~/.plannotator`.


backnotprop:
CI fix pushed in d8e0659b: the install endpoint tests were overriding PLANNOTATOR_DATA_DIR at module-eval time without restoring it, which broke the Pi annotate-history unwritable-dir test and both durable-submit-record tests later in the same process (bun evaluates all test modules before running tests, and Pi's generated/storage.ts caches its data dir at import). A plain afterAll restore reproduces the same three failures with the mismatch inverted, so the env var is now only mutated inside tests and restored per-test, the advert tests' config writes are snapshot-restored in afterAll, and the process-global mock.module scope is documented. Regression proof: running the endpoint file plus both annotate files in one bun invocation now passes in either order.

PR #1272, opened 2026-08-12
fix(review): contain /api/call-flow analysis throws as JSON error responses

backnotprop:
**TLDR:** The v0.27.0 pre-release QA sweep found one confirmed major: a hard git failure during Call Flow patch materialization escaped the `/api/call-flow` handler. On Pi that unhandled rejection reaches the process-level handler and kills the user's entire session; on Bun it surfaces as a non-JSON 500 that breaks the client's quiet-failure UX. This wraps the handler in both runtimes so failures return the standard `{ status: "error", reason: "analysis-failed" }` JSON envelope the client already renders.

Reachable only with Call Flow enabled plus a hard VCS failure (for example an unreadable untracked file), but the failure mode on Pi is a dead session from a read-only analysis request, so it should not ship.

Verified: full typecheck across all seven projects, both call-flow endpoint suites (26 pass, covering Bun and Pi).

*AI-assisted (found by a 29-agent QA workflow, verified adversarially, fixed by Claude).*


backnotprop:
Added a second commit: the consent copy in the intro dialog and Settings was six sentences; it is now three. Same disclosed facts (what installs, size, Node 22+, languages-as-needed), no filler. Tests updated to pin the new strings.

PR #1274, opened 2026-08-12
ci: harden releases and add security scanning

backnotprop:
## Summary

This PR implements the first three phases of the Plannotator security rollout:

- hardens release and production deployment boundaries without requiring pull requests for owner changes to `main`;
- enables repository-native dependency ownership and security reporting;
- adds credential-isolated Gitleaks and zizmor monitoring with SARIF output.

### Test and deployment

- Pins Bun 1.3.14 and uses frozen-lockfile installs in test, release, and deployment jobs.
- Disables checkout credential persistence where Git credentials are not required.
- Adds a release-version consistency test across all release-coupled manifests.
- Keeps continuous deployment from `main`, but deployment now waits for the exact commit SHA to pass the full `Test` workflow before any production job runs.
- Production jobs continue using the existing `production` GitHub environment and existing AWS/Cloudflare deployment paths.

### Release and npm publishing

- Requires every publishing `v*` tag to point to a commit contained in `main` and match all release manifest versions.
- Separates credential-free package construction from the privileged npm publication job.
- PRs and ordinary manual dry runs build and pack npm artifacts but receive neither npm credentials nor OIDC permission.
- Real npm publication remains automatic for valid `v*` tags and is isolated behind the `npm-publish` environment using npm trusted publishing/OIDC.
- Removes `NPM_TOKEN` from the workflow. The existing token should be revoked only after the first successful OIDC-only release.
- Preserves the existing GitHub release and artifact-attestation behavior.
- Disables Bun and Node package-manager caching in release jobs to prevent untrusted cache state from influencing published artifacts. The tradeoff is a small increase in release download/setup time.

### Security scanning

- Adds Gitleaks 8.30.0 and zizmor 1.29.0, installed from checksum-verified release archives.
- Runs on pull requests to `main`, pushes to `main`, a weekly schedule, and manual dispatch.
- Gitleaks scans PR/push commit ranges and performs full-history scans weekly/manually.
- zizmor runs offline with strict collection against all workflows, local actions, and Dependabot configuration.
- Both scanners include runtime-generated detector-health fixtures and fail when the scanner/configuration is broken.
- Findings start in monitor mode and upload as separate SARIF categories; findings do not yet block the PR.
- Fork and Dependabot PRs run without repository secrets and intentionally skip SARIF upload because their GitHub tokens are read-only.
- Local baselines are clean: zero active Gitleaks findings and zero active zizmor findings.

### Dependencies and reporting

- Adds weekly Bun Dependabot updates, grouped minor/patch PRs, and a seven-day cooldown for routine releases. Security updates bypass the cooldown; it does not increase CI duration.
- Keeps Renovate as the sole owner of GitHub Actions updates.
- Adds a latest-release-only security support policy and private vulnerability reporting instructions.

## Live prerequisites already configured

The following repository/service settings were configured before this PR and are not introduced by the YAML itself:

- GitHub dependency graph, Dependabot alerts/security updates, secret scanning, push protection, and private vulnerability reporting;
- `production` environment restricted to `main`;
- `npm-publish` environment restricted to `v*` tags;
- npm trusted publishers for both published packages;
- protected `v*` tag ruleset;
- AWS trust restricted to the GitHub `production` environment and S3 versioning enabled.

`main` remains directly writable by the owner. The controls gate deployment and publishing rather than changing that operating practice.

## Validation performed

- `bun run check:release-version`
- `bun test scripts/check-release-version.test.ts`
- actionlint 1.7.12 across all workflow files
- zizmor 1.29.0, offline regular persona with strict collection: zero active findings
- Gitleaks 8.30.0 working-tree and full-history audits: zero unresolved findings
- synthetic positive and reviewed-allowlist scanner health checks
- `git diff --check`

## Follow-up after merge

- Verify the first live SARIF uploads and fork behavior.
- Verify the next normal tag release through npm trusted publishing, then delete the repository `NPM_TOKEN` and revoke the superseded npm token.
- Validate the first Dependabot Bun PR and security notification delivery.
- Introduce blocking for new findings only after the monitor-mode GitHub run is confirmed.
- Implement Semgrep/Trivy, SBOM/Grype, isolated ZAP, and governance in separate follow-up PRs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,23 @@
+version: 2
+
+# Dependabot owns Bun workspace updates. Renovate owns GitHub Actions.
+updates:
+  - package-ecosystem: bun
+    directory: /
+    schedule:
+      interval: weekly
+      day: monday
+      time: "05:00"
+      timezone: America/Los_Angeles
+    # Delay routine releases long enough for ecosystem problems to surface.
+    # Dependabot security updates do not observe this cooldown.
+    cooldown:
+      default-days: 7
+    open-pull-requests-limit: 5
+    groups:
+      bun-minor-and-patch:
+        patterns:
+          - "*"
+        update-types:
+          - minor
+          - patch
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -28,7 +28,64 @@ permissions:
   contents: read
 
 jobs:
+  # Deploy remains automatic, but only after the exact commit has passed the
+  # repository's full Test workflow. The workflows start together on a main
+  # push; this job waits for Test instead of allowing deployment to race it.
+  verify:
+    name: Require successful Test run
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    permissions:
+      actions: read
+      contents: read
+    steps:
+      - name: Wait for Test on this commit
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPOSITORY: ${{ github.repository }}
+          COMMIT_SHA: ${{ github.sha }}
+        shell: bash
+        run: |
+          deadline=$((SECONDS + 2400))
+          previous_state=""
+
+          while (( SECONDS < deadline )); do
+            run_json="$(gh api --method GET \
+              "repos/${REPOSITORY}/actions/workflows/test.yml/runs" \
+              -f head_sha="$COMMIT_SHA" \
+              -f event=push \
+              -f per_page=1)"
+
+            status="$(jq -r '.workflow_runs[0].status // "missing"' <<<"$run_json")"
+            conclusion="$(jq -r '.workflow_runs[0].conclusion // ""' <<<"$run_json")"
+            run_url="$(jq -r '.workflow_runs[0].html_url // ""' <<<"$run_json")"
+            state="${status}:${conclusion}"
+
+            if [[ "$state" != "$previous_state" ]]; then
+              if [[ -n "$run_url" ]]; then
+                echo "Test workflow for ${COMMIT_SHA}: ${state} (${run_url})"
+              else
+                echo "Test workflow for ${COMMIT_SHA}: ${state}"
+              fi
+              previous_state="$state"
+            fi
+
+            if [[ "$status" == "completed" ]]; then
+              if [[ "$conclusion" == "success" ]]; then
+                exit 0
+              fi
+              echo "Test workflow did not succeed for ${COMMIT_SHA}: ${conclusion} (${run_url})" >&2
+              exit 1
+            fi
+
+            sleep 15
+          done
+
+          echo "Timed out waiting for the Test workflow on ${COMMIT_SHA}." >&2
+          exit 1
+
   detect-changes:
+    needs: verify
     runs-on: ubuntu-latest
     outputs:
       marketing: ${{ steps.changes.outputs.marketing }}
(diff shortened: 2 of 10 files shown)


backnotprop (commented):
## TLDR

The workflow engineering here is genuinely solid: every `uses:` is SHA-pinned, both scanner checksums match upstream, the tag-ancestry and version-consistency gates are correctly plumbed, the credential split between `npm-package` and `npm-publish` is real, and the live repo config (environments, tag ruleset, PVR) matches what the PR body claims. The one thing that can hurt you is npm trusted publishing readiness: the PR removes `NPM_TOKEN` entirely, so if the npm-side trusted publishers are not actually configured when the next `v*` tag fires, `npm-publish` fails after the immutable GitHub release already exists, and the retry window is only 24 hours because of `retention-days: 1`. Recommendation: tag v0.27.0 before merging, or verify both trusted publishers on npmjs.com first.

## BLOCKER (process, not code): confirm npm trusted publishing before any v* tag runs this workflow

`.github/workflows/release.yml` removes `NPM_TOKEN` and makes `npm-publish` OIDC-only (`id-token: write`, `environment: npm-publish`, no checkout, publish from downloaded tarballs). The PR body lists "npm trusted publishers for both published packages" under "Live prerequisites already configured", but that claim is not verifiable from the repo, and my direction for this review is that trusted publishing has NOT yet been configured for `@plannotator/opencode` and `@plannotator/pi-extension`.

Failure mode if it is not configured when a tag fires:

1. `npm publish` attempts the OIDC exchange, gets nothing, falls back to the `.npmrc` that `setup-node` wrote (`_authToken=${NODE_AUTH_TOKEN}`, which setup-node populates with a placeholder), and fails with an auth error. This does fail closed: nothing can be published with a bogus token, and there is no `NPM_TOKEN` fallback anymore. Good.
2. But `release` (GitHub Release creation) runs in parallel on the `attest` chain, not behind `npm-publish`. The immutable release and the protected `v*` tag ruleset (I verified both are live: ruleset 20729661 blocks update and deletion on `refs/tags/v*`) mean you cannot delete and re-tag. The only recovery is re-running the failed `npm-publish` job, and that only works while the `npm-packages` artifact exists.

Before the first tag on this workflow, configure both packages on npmjs.com with: repository `backnotprop/plannotator`, workflow `release.yml`, environment `npm-publish` (the environment name is validated when specified, and the job does set it). Then a manual `workflow_dispatch` on an existing tag with dry-run=false, or simply the v0.27.1-era release, validates the path.

## SHOULD FIX

1. **`retention-days: 1` on the `npm-packages` artifact** (`release.yml`, `npm-package` job). This is the entire recovery window for a failed `npm-publish` on a release that cannot be re-tagged. The tarballs are a few MB. Bump to 7 or more; there is no secret material in them (they are the public packages).

2. **Deploy verify has no escape hatch** (`deploy.yml`, `verify` job). It polls `actions/workflows/test.yml/runs?head_sha=SHA&event=push`. If a main push carried `[skip ci]`, or the Test run was cancelled, no `event=push` success will ever exist for that SHA, and even a manual `workflow_dispatch` deploy hits the same 40-minute poll and fails, because the query is hard-coded to `event=push`. The only workaround would be editing the workflow on main. Add a `workflow_dispatch` boolean input (`skip-test-gate` or similar) that lets a manual dispatch bypass or re-trigger the gate, or accept any successful Test run for the SHA regardless of event.

3. **Gitleaks "commit range" scans are not actually range-scoped** (`security.yml`). `--log-opts="--all ${BASE}..${HEAD}"` passes `--all` alongside the range, and `git log --all A..B` walks all refs minus ancestors of A, not just the PR commits. This is the shape the gitleaks README suggests, and it over-scans rather than under-scans, so it is safe. But the step summary claims "pull request commit range" and the PR body claims "scans PR/push commit ranges", and neither is what runs: PR scans will surface findings from unrelated stale branches and take longer on a busy repo. Either drop `--all` in the range branches or relabel the scope honestly.

## CONSIDER

1. **Assert `npm --version` >= 11.5.1 in `npm-publish`.** Trusted publishing needs npm 11.5.1+. `node-version: 24` currently ships a new enough npm, but nothing pins that, and an OIDC failure caused by an old npm would present as a confusing auth error. A one-line preflight makes the failure legible.

2. **Drop `registry-url` from the `npm-package` job's setup-node.** The pack job never talks to the registry; the input just writes an `.npmrc` with an auth placeholder into a job you deliberately stripped of credentials. Only `npm-publish` needs it.

3. **Security concurrency group cancels the weekly deep scan.** `security-scanning-${{ github.event.pull_request.number || github.ref }}` puts schedule runs and main-push runs in the same `refs/heads/main` group with `cancel-in-progress: true`, so a push landing during the Saturday full-history scan cancels it and the range-scoped push scan replaces it. Include `github.event_name` in the group key.

4. **The version-check test exercises Bun, CI runs Node.** `runCheck` spawns `process.execPath`, which is `bun` under `bun test`, while `release-eligibility` and `check:release-version` run the script with `node`. For a script this small it hardly matters; noting it so the divergence is a choice, not an accident.

## VERIFIED-OK

- **Tag ancestry check is correct.** `fetch-depth: 0` plus an explicit `git fetch --no-tags origin +refs/heads/main:refs/remotes/origin/main` makes the check independent of checkout's ref layout; `rev-parse "${GITHUB_REF}^{commit}"` peels annotated tags; `merge-base --is-ancestor` is the right predicate. Runs as an independent job feeding `attest` and `npm-publish` via `needs`, so an off-main tag can neither attest, release, nor publish even if the build succeeds.
- **`check-release-version.mjs` manifest list exactly matches the canonical 7** in `.agents/skills/release/SKILL.md` (root, opencode-plugin, pi-extension, hook plugin.json, copilot plugin.json, openpackage.yml, packages/server), and correctly excludes `apps/vscode-extension`. All 7 are at 0.26.8 lockstep on main, so wiring the check into the required `test` job is safe. The test is a real behavior-guarding subprocess test with temp-dir fixtures and no module-scope global mutation.
- **Binary build, smoke matrix, and attestation are untouched.** All 12 attestation subjects unchanged; `attest` -> `release` ordering (provenance before release publish) preserved; job additions are purely additive to the graph.
- **Every `uses:` in all four workflows is SHA-pinned** (I grepped for non-40-hex refs: none).
- **Both scanner checksums verified against upstream**: gitleaks 8.30.0 linux_x64 `79a3ab...a66e` and zizmor 1.29.0 x86_64-gnu `dd96df...7839` match the published release artifacts.
- **No GITLEAKS_LICENSE needed**: the workflow uses the MIT-licensed gitleaks binary directly, not gitleaks-action (whose org licensing does not apply here).
- **Live config matches the PR body where I could check it**: `npm-publish` environment exists with tag policy `v*`, `production` with branch policy `main`, "Protect release tags" ruleset active on `refs/tags/v*`, private vulnerability reporting enabled (so the SECURITY.md link works), secret scanning and push protection enabled.
- **SARIF permissions are scoped per job** (`security-events: write` only on the two scanner jobs), and the fork/Dependabot upload skip matches the read-only-token reality.
- **`.gitleaks.toml` is minimal and safe**: one exact-match allowlist for the RFC 6455 test vector, scoped to `generic-api-key` with `regexTarget = "match"`, no path excludes, and the detector-health step proves the allowlist does not swallow real findings.
- **`persist-credentials: false` breaks nothing**: the repo is public, the later `git fetch` calls in deploy/security are unauthenticated-safe, and no step pushes.
- **`bun pm pack --filename ../../npm-packages/x.tgz` works on Bun 1.3.14** (tested locally), and `digest-mismatch: error` is a real download-artifact v8 input.
- **Bun 1.3.14 pin matches** release.yml (already pinned on main) and local dev (`bun --version` is 1.3.14); no `packageManager` field to conflict. Frozen lockfile failing on drift is the intended behavior.
- **Nothing touches the install-scripts S3 bucket**; the hand-sync comment block in deploy.yml is preserved verbatim.
- **No stale `DRY_RUN`/`NPM_TOKEN` references remain**, and since main has no branch protection with required checks, the `npm-publish` -> `npm-package`/`npm-publish` job split cannot orphan a required check name. `verify`, `detect-changes`, and all deploy job ids are additive or unchanged.
- **Deploy gate timing is comfortable**: recent Test runs complete in 3 to 5 minutes against a 40-minute poll budget, and the gate fails closed on failure, cancellation, or absence.
- **Dependabot config is coherent**: `bun` ecosystem against bun.lock (supported), weekly with cooldown and minor/patch grouping, and no `github-actions` ecosystem so it cannot fight Renovate, which `.github/renovate.json` confirms owns actions exclusively.

## Release interaction (v0.27.0)

**Recommendation: tag v0.27.0 before merging this PR**, unless you configure and positively verify both npm trusted publishers first.

Reasoning: v0.27.0 on the current main uses the known-good `NPM_TOKEN` path. Merging first means the very next tag bets the npm leg of an immutable, non-re-taggable release on npm-side configuration that (per your own account, contra the PR body) does not exist yet, with a 24-hour re-run window. Nothing else in this PR threatens the v0.27.0 flow: the ancestry gate passes for the release skill's push-main-then-tag order, the version check passes because the skill bumps all 7 files atomically, and the deploy/test changes are orthogonal.

Sequence I would run: (1) tag and ship v0.27.0 on the old flow; (2) configure trusted publishers for both packages on npmjs.com (workflow `release.yml`, environment `npm-publish`); (3) merge this PR, ideally with the retention-days bump; (4) validate OIDC publish on the next tag (or a dispatched dry-run=false on a fresh tag); (5) only then delete and revoke `NPM_TOKEN`, exactly as the PR's follow-up list says.

If you prefer to merge before tagging: do steps 2 and the retention bump first, and treat the first tag as the OIDC validation run with a same-day recovery plan (re-run `npm-publish` within 24 hours if it fails).

backnotprop:
**Correction to the review above.** The BLOCKER item was based on a wrong premise supplied by the review coordinator, not on anything observed on npm: the maintainer has confirmed trusted publishers ARE configured for both @plannotator/opencode and @plannotator/pi-extension and were verified manually. With that, the OIDC-only publish path is ready and there is no process blocker.

Still standing from the review, both small:

- Bump the `npm-packages` artifact `retention-days` from 1 to 7+. It is the retry window if any publish job needs a re-run against an immutable release.
- The deploy `verify` job has no manual override: a `[skip ci]` push to main or a cancelled Test run dead-ends `workflow_dispatch` deploys because the query filters on `event=push` only.

*AI-assisted (Claude, Fable 5) under maintainer direction.*

PR #1277, opened 2026-08-12
feat(review): refine Call Flow navigation and annotations

backnotprop:
## Summary

- Rework the Call Flow path view into collapsible entry sections with changed-path defaults, optional full context, file boundaries, and more legible source locations.
- Add structured Paths search across call labels and full file paths. Search walks complete inferred trees, reveals matching unchanged context without opening all context, highlights the active result, supports Enter / Shift+Enter and Cmd/Ctrl+F, and shares its control pattern with Raw search.
- Add one-click Expand all / Collapse all for entry paths while retaining bounded changed-path rendering by default.
- Replace per-row comment icons with native row selection: click starts a comment and Shift-click collects multiple inferred steps into one annotation. Every Call Flow row is commentable; in-hunk targets remain inline while out-of-hunk and source-less targets safely become file- or review-scoped feedback.
- Preserve complete Call Flow target context in local feedback and hosted PR submission.
- Add a searchable, color-classified raw CallDiff view with copy support, and move language management into the panel header.
- Refine the file sidebar with copy-all, persistent Viewed/Git-add visibility controls, left-aligned viewed progress, and a search field anchored directly beneath its trigger row.
- Share the composer-yield state machine with HTML annotation so Shift-click selection stays usable around the open composer.

## Verification

- `bun test` — 3,306 passed, 489 skipped, 0 failed
- Structured Call Flow + Raw search DOM suite — 6 passed, 0 failed
- `bun run typecheck`
- `bun run build:review`
- `bun run build:hook`
- `git diff --check`
- Live review QA on the actual branch result: 21/21 entry paths expanded and collapsed; structured search found and navigated 128 matches while unchanged context remained hidden; no browser warnings or errors.

The Call Flow raw view remains a dedicated surface; this does not add Pierre decorations to the ordinary file diff.


backnotprop (commented):
**TLDR:** The two churn-sensitive areas hold up: the #1273 panel chrome composition is intact (last-used memo, resolvePanelView wiring, single search UI, self-heal all verified), and the composerYield extraction is a clean re-export shim with zero logic drift. One blocker: the new raw-view Cmd+F capture listener stays live while the Call Flow tab is open but inactive, because dockview keeps panel portals mounted, so it hijacks Cmd/Ctrl+F from the whole review. Two smaller holes (a viewed-files trap, testing-rules violations in one new test) and a pinned-copy removal that needs explicit sign-off.

## BLOCKER

1. **Cmd+F is hijacked from a background Call Flow tab.** `packages/review-editor/dock/panels/ReviewCallFlowPanel.tsx:266-269` registers a window-level capture keydown that calls `stopImmediatePropagation()` on every Cmd/Ctrl+F while `CallFlowRawView` is mounted. Dockview renders panels through persistent React portals that are disposed only when the panel is closed, not when its tab is inactive (`dockview/dist/esm/react.js`, `ReactPart.dispose`), and the panel's `view: 'raw'` state survives tab switches. Repro: open Call Flow, switch to Raw, click back to a file tab leaving the Call Flow tab open, press Cmd+F. The sidebar diff search (bubble-phase handler, `packages/review-editor/App.tsx:1497`) never fires and the hidden raw search opens instead. Gate the listener on panel activity, `state.isCallFlowActive` is already in `ReviewState`, or on visibility (`rawPreRef.current?.offsetParent`). The raw test asserts the capture behavior only for the active-panel case, so this gap is untested.

## SHOULD FIX

2. **Disabling Viewed controls can strand hidden files.** `hideViewedFiles` is session state (`App.tsx:345`) and keeps filtering rows (`SectionsPanel.tsx:274`, `FileTreeNode.tsx:40`) after Viewed controls are turned off, but the eye toggle that clears it lives inside the `showViewedControls` gate (`PanelChrome.tsx:65`). Sequence: hide viewed files, then switch off Viewed controls in the new Tree controls popover; viewed files vanish with no visible affordance until the setting is re-enabled. Reset `hideViewedFiles` when the toggle turns the controls off, or keep the eye rendered while the filter is active.

3. **`PanelChrome.test.tsx` violates the testing rules #1272 just landed.** Line 84 pins the absence of a prose sentence ("Choose which per-file controls appear.") which guards nothing and reads as a leftover from an earlier iteration; lines 59-60 pin Tailwind classes (`pl-1`, `pr-2`), which is implementation echo. The sibling `FileTree.search-placement.test.tsx` shows the right pattern (structural `data-*` attributes). Drop these assertions; the switch/copy behavior assertions in the same test are fine.

## CONSIDER

4. **CopyDiffFooter removal erases a maintainer-pinned copy from #1273.** The footer whose tooltip "click to copy the entire diff set" was commented "Pinned copy: tooltip text specified verbatim by the maintainer" is deleted wholesale (no remaining references, verified), replaced by a "Copy all diffs" icon in the utility row with proper aria-label and an sr-only status region. The evolution is coherent and declared in the PR body, but since the old string was an explicit pin, this should be an acknowledged decision, not silent drift.

5. **Raw search and raw rendering are unbounded.** `findCallFlowRawMatches` (`callFlowPresentation.ts:47`) has no match cap or minimum query length; a one-character query over a large CallDiff output creates thousands of `<mark>` nodes per keystroke, and `CallFlowRawView` now materializes every raw line as a `<span>` where main rendered one text node. `content-visibility` covers entries but not the raw pane. A match cap (for example 1000, with a "capped" indicator) would bound the worst case.

6. **Composer anchor can go stale across a diff refresh.** `targetElements` (`CallFlowTreeView.tsx:384`) is never pruned and `draftTargets` survive a `trees` swap while `collapsedPaths` reset; an open composer can end up anchored to a detached row (rect collapses to 0,0). Clearing the draft, or re-resolving the anchor, on `trees` change would be tighter. Submission is safe either way since `resolveCallFlowAnnotationPlacement` re-validates against the current files at commit time.

7. **Dead shapes.** `annotationTargetForCallFlowNode` declares `| null` but never returns null (`CallFlowTreeView.tsx:27-45`), making the downstream `annotatable` null-check unreachable, and `annotationSelectionForCallFlowNode` is now a pass-through alias of `selectionForCallFlowNode`. Fine to keep for API symmetry, but worth tightening.

## VERIFIED-OK

- **#1273 composition intact.** `resolvePanelView`/`effectivePanelView` wiring survives (`App.tsx:2009`, `App.tsx:3885`), last-used recording fires on all toggle paths with commits correctly excluded (`App.tsx:400`), self-heal still passes `recordLastUsed: false` (`App.tsx:2410`), full-width `PanelViewToggle` keeps the header row. The old inline search inputs are removed from both panels and `PanelSearchField` renders exactly once, directly beneath `PanelControlsRow`, asserted structurally by `FileTree.search-placement.test.tsx`. No duplicate or competing search UI. Nav rows now stay visible during search in both panels; deliberate and consistent across FileTree and SectionsPanel.
- **composerYield sharing is clean.** `packages/ui/utils/composerYield.ts` is logic-identical to the original (same constants, same hysteresis); `packages/ui/components/html-viewer/composerYield.ts` is a one-line re-export shim; `HtmlViewer` and `composerYield.test.ts` consume the shim unchanged; the published `./utils/*` export map covers the new path, so no `@plannotator/ui` consumer breakage and no `packages/ui/README.md` seam impact. The `CommentPopover` click-outside change (shift-pointerdown preserved while target chips exist, `CommentPopover.tsx:228-247`) also reaches HTML annotate, but it is the correct multi-select semantics on that surface too.
- **Call Flow annotation targeting is sound.** Rename normalization via `oldPath` (`callFlowAnnotations.ts:42`), per-side in-hunk validation through `isLineRangeInPatch`, out-of-hunk demotion to file scope and source-less to general scope, with every target retained in selection order. A target file absent from the diff degrades to file scope, and navigation routes call-flow-scoped feedback back to the Call Flow panel instead of issuing a dead scroll (`App.tsx:2655`). Hosted PR submission posts inline comments only for line scope and carries full step context in the review body otherwise (all three shapes covered by `ReviewSubmissionDialog.test.ts`). No server counterpart is needed: targets ride inside annotation objects and exported markdown, nothing reads a server field that does not exist, so the two-runtime law is not triggered. Workspace/multi-repo mode is moot since Call Flow is git-session only.
- **Shift-click selection composes with the existing systems.** Rows are real buttons (keyboard reachable, `aria-pressed`, sr-only live count); Escape and stray-key capture come from `CommentPopover`; window blur and Shift keyup both release the yield state, so no stuck click-through composer; per-instance chip keys (`treeInstanceId` prefix) keep the lens and dock composers from cross-talking; the lens popover correctly refuses to close while a draft is active and closes when the draft resolves (`CallFlowFileBadge.tsx`).
- **test.yml is complete.** All five new DOM test files are registered in `DOM_TESTS`; the remaining new tests (callFlowAnnotations, callFlowPresentation, exportFeedback, ReviewSubmissionDialog, CallFlowTreeView static markup) are node-safe. I ran the new and modified suites against the branch: 77 pass, 0 fail.
- **settings.ts follows convention.** Both new entries are cookie-only (`serverKey: undefined`) matching the persisted-chrome pattern, invalid cookie values fall through to the default, and no invalid state is persistable.
- **types.ts is backward compatible.** `CallFlowAnnotationTarget` is a new export, `callFlowTargets` is optional on `CodeAnnotation`; existing drafts and share links are unaffected (code review has no share payload, and drafts tolerate the extra field in both directions).
- **CSS is contained.** No new directories, so no `@source` additions needed; new chrome lives in already-scanned paths; coarse-pointer hit targets and `prefers-reduced-motion` are extended to the new controls; the raw-line kind classifier only colors first-column markers so operators and path hyphens stay neutral (tested).
- **Dead code from the replaced per-row comment icons is fully gone.** No remaining `CopyDiffFooter`, `onCommentNode`, or `call-flow-comment` references outside the intentional negative assertion in `CallFlowTreeView.test.ts`.

backnotprop (commented):
**Delta review of `a9da0e62..5cce1a7b` (5 commits since the previous review).**

**TLDR:** The lens work itself is well built and well tested, but the Cmd+F blocker is **NOT RESOLVED, it widened**: the ungated window capture listener now exists in the Paths view as well as Raw, so a backgrounded Call Flow tab hijacks Cmd/Ctrl+F in more states than before, and two mounted instances now race for the same keystroke. One new CI gap: the test file covering the claimed lens Cmd+F behavior is not registered in `DOM_TESTS`, so it silently skips in CI. The viewed-files trap and the PanelChrome test pins are untouched.

## BLOCKER: NOT RESOLVED

- The remediation added refocus-with-selected-query, but never gated the listener. `CallFlowTreeView.tsx:489-503` (new, Paths view) and `CallFlowRawView.tsx:97-110` both register a window capture keydown with `stopImmediatePropagation()` whenever mounted; there is no `state.isCallFlowActive` or visibility check anywhere. Dockview portals persist for inactive tabs, so the repro from the previous review still holds and now needs one step less: open the Call Flow tab, switch to any file tab, press Cmd+F; the hidden panel swallows it and the sidebar diff search (`App.tsx:1497`, bubble phase) never fires. Before this delta the hijack required the Raw view; now the default Paths view does it too.
- New contention: with the dock tab open in the background and a lens open, both instances hold capture listeners and the first-registered one wins, which is normally the dock's. Cmd+F then opens or refocuses the HIDDEN dock search, not the active lens, contradicting the commit's stated behavior. Registration order also reshuffles whenever either instance's `searchOpen` flips (the effect re-registers on `[searchOpen]`), so which instance wins is unstable. `CallFlowFileBadge.interaction.test.tsx` mounts the lens alone, so it cannot catch this.
- Fix shape: each instance must bail unless it owns the keystroke, for example dock instances check `state.isCallFlowActive` plus root-element visibility (`offsetParent`), and the dock yields while a lens popover is open.

## Status of previous findings

- Viewed-files trap (hide-viewed left active when Viewed controls are switched off): **still open**, untouched by the delta.
- `PanelChrome.test.tsx` prose-absence pin (line 84) and `pl-1`/`pr-2` class pins (lines 59-60): **still open**, untouched.
- CopyDiffFooter removal (maintainer-pinned tooltip from #1273): **unchanged by the delta**; sign-off still pending, that decision is the maintainer's.
- Unbounded raw-search marks: **still open**; raw lines are now one `<button>` each, which raises DOM weight further on huge outputs.
- Stale composer anchor: **partially improved**; `CallFlowRawView` prunes its element map via ref callbacks, but `CallFlowTreeView.tsx:384` still never prunes and drafts still survive a `trees` swap.
- Unreachable `| null` return on `annotationTargetForCallFlowNode`: still present (trivial).

## New findings in the delta

- **SHOULD FIX:** `CallFlowFileBadge.interaction.test.tsx` is DOM-gated but missing from `DOM_TESTS` in `.github/workflows/test.yml` (the other new suites are listed at lines 99-100). In the non-DOM CI leg every case is `skipIf(!hasDom)`, so the suite that covers the remediation's headline behavior never actually runs in CI.
- **SHOULD FIX:** `parseCallDiffWorkerResult` (`packages/shared/call-flow-types.ts:243` onward) now throws when an entry's ascii is empty or is not an in-order substring of the response-level ascii. Previously the per-entry ascii was ignored entirely (the old fixture literally said "also omitted"), so nothing historically guaranteed that property of the pinned CallDiff 0.4.1 output. A mismatch now fails the ENTIRE analysis to protect a lens display slice. Degrade instead: fall back to a single unsectioned raw view rather than throwing.
- **CONSIDER:** raw lines set `user-select: text` but any click, including the mouseup that ends a text-selection drag inside a line, opens the composer; and raw-view multi-select passes no `yieldState`, so the composer does not fade out of the way of a Shift-click the way the tree view does.
- **CONSIDER:** lens Copy joins per-entry slices with blank lines, so it omits response-level header and diagnostic lines that the dock Copy includes.

## Verified in the delta

Sticky lens header and search toolbars are correctly scoped to the popover scroll container with coherent z-layers (dock toolbar is not sticky, matching the lens-only claim). The `CallFlowRawView`/`CallFlowSearchControls` extraction removes the dock/lens duplication cleanly. `CallFlowTree.raw`/`rawLineStart` typing is additive, raw-line targets flow through the existing placement machine as general-scope feedback (no server change needed), and export formats them as data (`raw CallDiff line N`). Delta suites pass here: 74 pass, 0 fail (DOM-gated cases verified separately). The two 5-second server-route timeouts they reported touch no file in this PR, consistent with pre-existing single-process suite flake; worth confirming they reproduce on main before attributing.

backnotprop:
Addressed the delta review in `81261c6a`:

- Cmd/Ctrl+F ownership is now explicit and shared across Paths/Raw. Retained Dockview portals only register while Call Flow is the active panel; an open Lens owns the shortcut over the Dock.
- Registered `CallFlowFileBadge.interaction.test.tsx` in the CI DOM test list.
- Per-entry raw alignment now degrades independently: a missing/misaligned entry slice no longer aborts structured analysis. Out-of-order alignment is covered.
- Disabling Viewed controls now clears an active hide-viewed filter before removing its only affordance.
- Reworked `PanelChrome.test.tsx` to assert behavior instead of Tailwind classes or prose absence.
- Left `CopyDiffFooter` unchanged pending maintainer sign-off.

Verification:
- Focused DOM + parser tests: 21 passed
- `bun run typecheck`: passed (including Pi vendoring/check)
- `bun run build:review && bun run build:hook`: passed
- Full `bun test`: 3,311 passed, 493 skipped, 0 failed
- Live production-bundle repro: backgrounded Paths and Raw tabs both yield Cmd+F to review search; a visible Lens focuses its own search instead of the retained Dock portal.

backnotprop (commented):
**Final verification of `81261c6a`.**

**TLDR:** All findings are resolved. The Cmd+F blocker is fixed the right way: ownership-gated listeners keyed to real dockview active-panel state, with lens precedence decided at keystroke time rather than by registration order. Merge-ready from this review's perspective.

- **BLOCKER (Cmd+F hijack): RESOLVED.** `useCallFlowFindShortcut` registers no listener at all unless `active`; both dock surfaces pass `state.isCallFlowActive` (the dockview active-panel signal, `App.tsx:256`/`1035`, carried in `ReviewState`), so a retained background portal holds no listener. Lens surfaces key on popover `open`. Precedence is a DOM sentinel (`[data-call-flow-lens="true"]`) checked inside the dock handler per keystroke, which is immune to the registration-order race from the previous round: the dock defers without stopping propagation no matter which listener runs first. Shift/Alt chords now pass through, and `visibilityState` is checked. The only residual path I could construct (lens somehow open while its host tab backgrounds) requires a keyboard-only panel switch mid-draft in a hover popover that otherwise self-closes; negligible.
- **DOM_TESTS: RESOLVED.** `CallFlowFileBadge.interaction.test.tsx` is registered (`test.yml:100`), and the new dock test exercises the regression class the earlier suite could not: `isCallFlowActive: false` mount asserts Cmd+F escapes untouched, a lens sentinel asserts dock deference, sentinel removal asserts the dock claims it. Verified locally: all DOM-gated suites pass under `DOM_TESTS=1` (13 pass, 0 fail) plus the node-safe suites (19 pass, 0 fail).
- **Parser: RESOLVED.** Unalignable, empty, or oversized entry ascii degrades to a raw-less `CallFlowTree` (explicit `CallFlowTreeWithRaw | CallFlowTreeWithoutRaw` union) instead of throwing; the lens skips unaligned sections, the dock raw view is unaffected, alignment is order-independent, and both behaviors are tested. No new throw paths. The shared file reaches Pi via `vendor.sh`, so both runtimes pick it up. One cosmetic nit: identical duplicate entry slices share the first occurrence's line numbers.
- **Viewed-controls trap: RESOLVED, correctly scoped.** The filter clears only on the controls' off-transition (`PanelChrome.tsx:273-277`); the eye toggle path is untouched, so a deliberate hide-viewed with controls enabled is never clobbered, and re-enabling the controls does not resurrect the filter.
- **Test quality: RESOLVED.** Class pins and the prose-absence assertion are gone; the harness is stateful and asserts the observable outcome (label flips, filter actually cleared across a disable/enable cycle).
- **CopyDiffFooter: closed by maintainer decision.**

Remaining open items are the previously noted CONSIDERs only (uncapped raw-search marks, unpruned tree-view anchor map, lens copy omitting response-level header lines), none blocking.

PR #1278, opened 2026-08-12
fix(review): restore the prior diff when leaving the Commits view

backnotprop:
## TLDR

Clicking a commit in the Commits panel hijacked the single session-global diff and there was no way back: switching to Tree kept the commit diff, and reloading the page did not escape either. This makes the Commits view self-contained: leaving it restores the diff the session was reviewing before, and a reload that lands on a stranded commit diff snaps back to the session default. Client-only; the server is unchanged.

## The bug

The maintainer reported that after clicking a commit in the Commits panel, that commit's diff becomes the global active diff and gets stuck: returning to Tree still shows the last-clicked commit's changeset instead of the review the user started from, with no obvious way back.

Root cause, from code reading plus a live sandboxed API repro:

- The review server keeps ONE active diff per session. `POST /api/diff/switch` with `commit:<sha>` mutates it for every surface, and `GET /api/diff` serves it afterwards, so a page reload (or a second tab) also lands on the commit diff. Reload was not an escape.
- On the client, merely entering the Commits view already switches the diff (HEAD auto-select in `useCommitsView`), and every commit click switches it again. Nothing recorded the pre-commit diff, and `handlePanelViewSelect('tree')` only changed the panel view, so leaving Commits could not restore anything.
- This was a known accepted edge: PR #994 shipped the feature with "toggling Commits to Tree keeps a commit diff active, where the tree's diff-type dropdown marks nothing."

## The fix

All in `packages/review-editor` (shared by both server runtimes' bundles):

- **Memo**: `preCommitDiffRef` captures the pre-commit diff (full diff type including any worktree prefix, plus the selected base) in `handleSelectCommit` on the first non-commit to commit switch. The HEAD auto-select routes through the same handler, so one capture site covers both entry paths. Walking further commits never overwrites it.
- **Restore**: leaving Commits for Tree restores the memo via `fetchDiffSwitch(memo.diffType, memo.base)`, deliberately not `handleDiffSwitch` (which would re-compose the current worktree prefix over an already-composed one). With no memo (post-reload; refs do not survive) it falls back to the session default using the exact resolution `handleWorktreeSwitch` already applies when it abandons a commit diff. Exiting while the entry's commit switch is still in flight also restores: the memo is captured synchronously before that fetch, so memo plus loading means a commit diff is inbound, and issuing the restore now supersedes it server-side via the existing epoch guard.
- **Clear invariant**: `fetchDiffSwitch` clears the memo whenever the APPLIED `data.diffType` is not commit-family. The superseded early-return never reaches the clear, so a stale response cannot drop a memo a newer commit switch still needs. Failed switches never reach it either, so the commit diff stays up with the normal `diffError`, no retry loop, memo retained.
- **Reload un-trap**: a one-shot load effect snaps a served commit-family diff back to the session default when the resolved opening panel view is not Commits (it never is; the view is deliberately not persisted). The one-shot ref burns on the first settled load regardless, so commit diffs the user opens later in the session are never snapped.
- The decision logic lives in a pure module, `utils/commitViewRestore.ts` (`isCommitDiffType`, `resolveCommitExitDiff`), unit-tested directly.

## What deliberately did NOT change

- The server keeps its single-active-diff session model; no endpoint changes.
- Git status remains a since-base reset (`handleSwitchToSections`), because that view IS the since-base comparison; it now also clears the memo through the applied-switch invariant.
- Annotations, drafts, Ask AI context, and the staleness probe still key off the active diff; cross-tab behavior via the snapshot-echo freshness probe is unchanged.
- The last-used panel view memo still records Tree when the user clicks Tree to leave Commits; a deliberate Tree click is a Tree preference.

## Tests

- New `utils/commitViewRestore.test.ts`: `isCommitDiffType` agreement with the App/server commit-family parse (plain, worktree-composed, colons in worktree paths, the 4 to 64 hex sha rule, non-commit modes), and `resolveCommitExitDiff` (memo wins verbatim; fallback prefers the configured default only when offered, else first option, else uncommitted; worktree prefix composition).
- `bun run typecheck`: green.
- `bun test packages/review-editor packages/ui`: 851 pass, 0 fail.
- Unmaintained review-editor tsc profile diffed against main: no new errors beyond the profile's known `bun:test` module noise, which already applies to all 47 existing test files and now also names the new test file.

*AI-assisted: implemented and tested by Claude under maintainer direction.*


backnotprop (commented):
## TLDR

**Approve with two SHOULD FIXes, neither of which is a code-behavior defect.** I traced every capture / clear / restore path in the PR head against the current `origin/main` (which now carries #1277) and could not find a way to fire a wrong or stale diff switch through the normal UI. The memo invariants hold, the regex agrees with both the App-inline worktree parse and the server's `parseCommitDiffType` on every producible diff type, and the composition with #1277 is clean semantically as well as textually. The two things I want changed are a test that cannot fail on the branch it claims to cover, and the missing `AGENTS.md` / `CLAUDE.md` sentence for a new session-level invariant.

Verification performed on a throwaway `git worktree` holding `origin/main` + `git merge --no-commit origin/fix/commits-view-restore-diff` (worktree removed afterwards):

- `bun test packages/review-editor/utils/commitViewRestore.test.ts` → 6 pass, 0 fail
- `bun test packages/review-editor` → 266 pass, 98 skip, 0 fail (55 files)
- `DOM_TESTS=1 bun test` on the panel-view files #1277 added (`PanelChrome.test.tsx`, `FileTree.search-placement.test.tsx`, `FileTreeNode.controls.test.tsx`) plus `resolvePanelView.test.ts` → 6 pass, 0 fail
- `tsc --noEmit -p packages/review-editor/tsconfig.json` on the merge preview → zero errors attributable to `App.tsx` or the new files (only the profile's pre-existing `bun:test` module noise). Note that the repo's `bun run typecheck` script does **not** include `packages/review-editor`, so "typecheck green" in the PR body proves nothing about this diff; the targeted run above is the real evidence.

---

## SHOULD FIX

### 1. `resolveCommitExitDiff`'s `preferredDefault` branch is untested; the test that names it cannot fail on it

`packages/review-editor/utils/commitViewRestore.test.ts:46-51` is titled "no memo: preferred default used only when the session offers it", but its fixture sets `preferredDefault: 'since-base'` while `diffOptions[0].id` is also `'since-base'` (`commitViewRestore.test.ts:35-39`). Every one of the four `resolveCommitExitDiff` assertions therefore returns the same value whether the implementation honors `preferredDefault` or ignores it entirely.

Proved by mutation on the merge preview. Replacing `packages/review-editor/utils/commitViewRestore.ts:45-47`

```ts
const resolved = diffOptions.some((o) => o.id === preferredDefault)
  ? (preferredDefault as string)
  : (diffOptions[0]?.id ?? 'uncommitted');
```

with the whole branch deleted (`const resolved = diffOptions[0]?.id ?? 'uncommitted';`) still gives **6 pass, 0 fail**. Under AGENTS.md's "name the failure it catches" rule, that assertion currently guards nothing, and the regression it is supposed to catch (a reload un-trap that snaps to the wrong diff because it silently stopped honoring `defaultDiffType`) is exactly the one worth guarding: it is also the rule `handleWorktreeSwitch` (`App.tsx:2474-2480`) hand-mirrors, so drift between the two is a live risk.

Minimal fix: make the preferred value differ from the first option, for example `preferredDefault: 'uncommitted'` against `diffOptions: [{ id: 'since-base' }, { id: 'uncommitted' }]`, expecting `'uncommitted'`. One extra assertion, and the branch is then covered.

### 2. The new session invariant is not documented in `AGENTS.md` / `CLAUDE.md`

`AGENTS.md:217` and `CLAUDE.md:217` carry the detailed contract for the `Git status | Tree | Commits` toggle, right down to "the Commits view is session-only and never an opening view" and which writer may touch `reviewPanelViewLastUsed`. This PR adds three durable invariants to that same contract:

1. leaving Commits for Tree restores the pre-Commits diff (memo captured on the first non-commit to commit switch, never overwritten while walking the rail);
2. any applied non-commit switch clears the memo;
3. a load served a commit-family diff snaps once to the session default because the opening view is never Commits.

Given how strictly this repo keeps that paragraph authoritative (and that the next agent reading it will otherwise re-derive the old "toggling Commits to Tree keeps a commit diff active" edge from #994), please add a sentence or two. Both files carry identical text, so both need the edit.

---

## CONSIDER

### 3. The reload un-trap mutates the session-global diff for every other connected surface

`App.tsx:2419-2431` fires `POST /api/diff/switch` on load whenever the served diff is commit-family. The server holds one active diff per session, so a **second** surface (a second browser tab, the VS Code panel, a remote-mode viewer opening the advertised URL) yanks the first viewer's commit diff out from under it: viewer A keeps rendering its commit locally, then its snapshot-echo freshness probe reports stale and shows "Diff out of date, Refresh". Clicking Refresh re-switches the global diff back to A's commit, which then makes B stale. Ping-pong is user-driven rather than automatic, so nothing corrupts, but the notice is confusing precisely in the scenario the un-trap targets (someone actively walking commits).

Before this PR a second surface loading was read-only with respect to the active diff. If the trade is deliberate, say so in the PR body; a cheap alternative is to make the reload snap conditional on there being no other connected client, though the review server has no client tracking today, so "accept and document" is a legitimate answer.

### 4. Wire reordering during the in-flight exit leaves a state with no in-app restore path

`App.tsx:2337-2361`: exiting while the HEAD auto-select switch is still in flight issues the restore, relying on the server's arrival-order epoch (`packages/server/review.ts:2010`, captured before any `await`, so the ordering is genuinely arrival order and not body-completion order). That is the right mechanism, and it makes the commit switch return `{ superseded: true }` in the normal case.

If the two requests are reordered on the wire (separate connections from the pool), the restore is superseded instead and the client lands on Tree with a commit diff and `panelView === 'tree'`. The exit branch requires `panelView === 'commits'`, so it can never fire again; the only escape is picking something in the diff dropdown. The same shape occurs when the restore switch simply fails (`fetchDiffSwitch` returns false at `App.tsx:2255-2258`), where the code comment says "the next exit is the recovery path", but reaching a next exit means re-entering Commits first. Both are recoverable and low probability, so this is a CONSIDER rather than a blocker; a cheap hardening would be to re-run the same snap logic whenever an applied diff is commit-family while `panelView !== 'commits'` (a generalization of the load-time effect rather than a one-shot).

### 5. Two concurrent switches are possible on a single load

The un-trap effect (`App.tsx:2420-2431`) and the persisted-pair self-heal (`App.tsx:2446-2459`) both key off the first settled load and both can issue a switch: the un-trap to `defaultDiffType`, the heal to `since-base`. On a session that is served a commit diff **and** has a conflicted persisted pair, both fire; the heal is declared second so it wins the epoch and the result is correct, but two full diff recomputes go out for one load. Ordering the un-trap behind "no heal pending", or simply having the un-trap skip when the heal will run, would remove the redundant server work.

### 6. The abandoned commit snapshot still costs a server-side Call Flow analysis

`useCallFlowAnalysis` (`packages/review-editor/hooks/useCallFlowAnalysis.ts:24-46`) re-runs on every `snapshotId` change and aborts only the **client** fetch. With Call Flow enabled, a load that hits the un-trap starts a CallDiff analysis for the commit snapshot the session is about to abandon, then another for the restored one. Bounded and only in the stuck-state path, but CallDiff is the expensive analysis in this app, so it is worth knowing. No change required if that is acceptable.

### 7. A fourth parse of the commit family now exists in the client

The client already answers "is this a commit diff" three ways: the inline worktree parse in `App.tsx:1117-1136` (`/^commit:[0-9a-f]{4,64}$/i` on the tail), `commitShaFromMode` in `utils/exportFeedback.ts:40-42` (prefix only, applied to the already-parsed base), and `handleWorktreeSwitch`'s `activeDiffBase.startsWith('commit:')` at `App.tsx:2474`. `isCommitDiffType` (`utils/commitViewRestore.ts:15-21`) is the fourth, and the test comment at `commitViewRestore.test.ts:4-6` asserts agreement by restating expectations rather than by construction.

I checked the agreement by hand and it holds for every producible diff type, including the worktree-with-colons forms (`worktree:C:/work/tree:commit:abcdef12` true; `worktree:/tmp/a:commit:deadbeef:uncommitted` false in both, since only the anchored tail counts). The only divergence I could construct is `foo:commit:abcdef`, which the App parse reads as a non-commit type while `isCommitDiffType` returns true, and neither the client nor the server can produce it.

Two of the three new call sites (`App.tsx:2345`, `App.tsx:2389`) have `activeDiffBase` in scope and could use the existing derived value, which would make the agreement structural instead of asserted. The `fetchDiffSwitch` clear at `App.tsx:2177` genuinely needs a raw-string helper because it reads `data.diffType` off the response, so extracting the App-inline parse into this new util and building `isCommitDiffType` on it would be the cleaner end state. Not required for this PR.

### 8. The Tree click that leaves Commits still records `reviewPanelViewLastUsed = 'tree'`

`selectPanelView` (`App.tsx:394-399`) records Tree for any non-Commits selection, so a user whose persisted and last-used view is Git status, who dips into Commits and clicks Tree to get back to their diff, silently changes what their **next** review opens on. That behavior predates this PR (#1273), but this PR makes the Tree click the primary escape from Commits, so the drift becomes much more likely. The PR body calls it deliberate ("a deliberate Tree click is a Tree preference"), which is defensible; flagging it so the decision is conscious. Restoring the view the session entered Commits from, rather than hardcoding Tree, would be the alternative.

---

## VERIFIED-OK

Traces I ran and found correct:

- **Enter Commits, click several commits, exit to Tree.** Capture is gated on `!isCommitDiffType(diffType)` (`App.tsx:2389`), so only the first non-commit to commit transition writes the memo; further rail clicks leave it alone. The re-click early return (`App.tsx:2377-2380`) sits before the capture and is harmless because `diffType` is already commit-family there.
- **HEAD auto-select on entry.** `useCommitsView`'s auto-select routes through the same `handleSelectCommit` (`hooks/useCommitsView.ts` `onOpenCommit`), so entry and user click share one capture site, and `if (isLoadingDiff) return` in that effect means the auto-select never races another in-flight switch.
- **Exit to Git status.** Handled before the restore branch (`App.tsx:2333-2336`), so no double switch; `handleSwitchToSections` to `since-base` is an applied non-commit switch and clears the memo through `App.tsx:2177`.
- **Diff-switch failure.** The `throw` path skips the clear, so a failed switch keeps the memo and no retry loop exists. Confirmed the clear sits after the `superseded` early return (`App.tsx:2169` then `2177`), so a stale body cannot drop a memo a newer commit switch still needs.
- **Superseded responses.** Server side confirmed: `const switchEpoch = ++diffSwitchEpoch;` is captured before `await req.json()` (`packages/server/review.ts:2010`) and all three exit points re-check it (`review.ts:2040`, `2103`, `2160`), so the client's last-response-wins assumption is backed by real arrival-order semantics.
- **Reload.** `snappedCommitDiffOnLoad` burns on the first settled load regardless of what it observed (`App.tsx:2421-2422`), so a commit diff opened later in the session is never snapped. `diffType`, `diffData`, `gitContext` and `configStore.init()` all land in the same batch inside the `/api/diff` handler (`App.tsx:1606-1638`), so the effect never reads a half-populated load. The `panelView === 'commits'` guard is genuinely unreachable: both `reviewPanelView` and `reviewPanelViewLastUsed` reject `'commits'` at the cookie reader (`packages/ui/config/settings.ts:200-228`), and `selectPanelView` never records it.
- **Worktree sessions.** Restore deliberately uses `fetchDiffSwitch` with the memo's full diff type rather than `handleDiffSwitch`, avoiding double prefix composition; the no-memo fallback composes `worktree:<path>:<resolved>` and matches `handleWorktreeSwitch`'s resolution rule byte for byte (`utils/commitViewRestore.ts:45-51` vs `App.tsx:2474-2483`). A worktree switch out of a commit diff is itself an applied non-commit switch, so it also clears the memo.
- **Base picker mid-Commits.** Not reachable: `FileTree` and `SectionsPanel` (which own the base and diff pickers) render only when `!showCommitsPanel` (`App.tsx:3839`), and `CommitsPanel` exposes only the view toggle plus rail actions (`App.tsx:3821-3835`). Every panel exit routes through `handlePanelViewSelect`, including `FileTree`'s `onSwitchToTree` (`components/FileTree.tsx:393-397`) and `CommitsPanel`'s toggle, so there is no bypass of the restore branch.
- **Dangling memo after a failed capture.** A capture whose switch then fails leaves a memo with no commit diff active. I could not find a path where that memo is later read against a mismatched diff: it is only read while `panelView === 'commits'`, and every route into the commit family from a non-commit diff recaptures first. The one route that reaches a commit diff without `handleSelectCommit` is the page load, where the memo is null by construction.
- **Composition with #1277.** `origin/main`'s `App.tsx` changes (`handleAddCallFlowAnnotation`, the Call Flow branch in `handleNavigateToAnnotation`, the viewed/stage control toggles and their `FileTree` / `SectionsPanel` props) touch no part of the diff-switch, panel-view or commits machinery. Merge preview confirms all four edit sites land intact (`isCommitDiffType` at merged lines 2210 / 2378 / 2422 / 2456, `preCommitDiffRef` at 440), and the merged tree typechecks and tests clean.
- **CI.** The `test` job runs a bare `bun test`, so the new file is picked up without a workflow edit; #1277's explicit DOM list needs no addition since this test is not DOM-gated.
- **Test rules.** No pinned prose, no round-trip prop assertions, no module-scope global mutation, no real data dir touched. Only finding 1 above falls short.

---

## Merge readiness

Ready once finding 1 is addressed (one fixture value plus one assertion) and finding 2's documentation sentence lands. Everything else is optional and can be answered in the thread.

PR #1279, opened 2026-08-12
feat(editor): focus-mode shortcut to toggle both sidebars

### LINKED ISSUE #1276: [FEATURE REQUEST] A keybind to enable focus mode in annotations
A good feature is a keybind that would enable focus mode in annotation. it would just close the 2 sidebars on left and right right away. This would benefit keyboard-first users tremendously instead of reaching for both sidebars every time

backnotprop:
**TLDR:** Closes #1276. One shortcut, Mod+., toggles focus mode in plan review and annotate: first press closes both the left sidebar and the annotation panel, second press restores the exact prior arrangement. Verified end to end in headless Chrome via CDP, including the typing guard and asymmetric open states.

## Why Mod+.

A full audit of every binding across both plan surfaces (registry scopes plus ad-hoc handlers, all states) found Mod+. unclaimed everywhere on these surfaces. Mod+B is already the annotate Contents toggle and means bold inside editable fields; Mod+\ requires AltGr on German and Nordic layouts, so it silently fails there. Mod+. is unshifted on every layout, unclaimed by browsers and OSes, and matches this codebase's existing idiom: code review already uses Mod+. to collapse its sidebar. A registry test now asserts the binding has exactly one claimant per surface.

## What changed

- New `document-view` scope with a `toggleFocusMode` action, composed into both `planReviewSurface` and `annotateSurface`; the marketing shortcuts docs page picks it up automatically.
- The handler drives the existing focus view-mode machinery (snapshot and restore already existed behind the document card's Focus button); the keyboard entry point was the only missing piece. A press while wide mode is active restores instead of swapping, since re-hiding already-hidden panels reads as a dead key. Pure decision logic extracted to `resolveFocusShortcutAction` with unit tests.
- Focus button tooltip now shows the binding; the shortcuts help modal gains a View section.
- Raw-HTML annotate surfaces are deliberately excluded: they persist their own chrome state and do not render the Focus control.

## Verification

- `bun run typecheck` clean; `bun test packages/ui packages/editor` 674 pass, 0 fail.
- Headless Chrome via raw CDP against real sandboxed servers: screencast video plus before/during/after screenshots for annotate and plan review, the typing guard (press inside a comment editor does nothing), and the help modal entry.

*AI-assisted: implemented and verified by Claude under maintainer direction.*

### DIFF
--- a/packages/editor/App.tsx
+++ b/packages/editor/App.tsx
@@ -111,10 +111,12 @@ import { observeActionsLabelMode } from './actionsLabelMode';
 // same env var on the server side so V2/V3 stay paired.
 import { DEMO_PLAN_CONTENT as DEFAULT_DEMO_PLAN_CONTENT } from './demoPlan';
 import { DIFF_DEMO_PLAN_CONTENT } from './demoPlanDiffDemo';
-import { canUseAnnotateWideMode, resolveWideModeExitLayout, type WideModeLayoutSnapshot, type WideModeType } from '@plannotator/ui/utils/wideMode';
+import { canUseAnnotateWideMode, resolveFocusShortcutAction, resolveWideModeExitLayout, type WideModeLayoutSnapshot, type WideModeType } from '@plannotator/ui/utils/wideMode';
+import { modKey } from '@plannotator/ui/utils/platform';
 import {
   annotateSidebarShortcuts,
   useAnnotateSidebarShortcuts,
+  useDocumentViewShortcuts,
   useDoubleTapShortcuts,
 } from '@plannotator/ui/shortcuts';
 const USE_DIFF_DEMO =
@@ -984,8 +986,11 @@ const App: React.FC = () => {
     [projectRoot, uiPrefs]
   );
 
-  const canHandleAnnotateSidebarShortcut = useCallback((event: KeyboardEvent) => {
-    if (!annotateMode || archive.archiveMode || goalSetupMode) return false;
+  // Shared gate for the chrome-level keyboard commands (sidebars, focus mode):
+  // never while a dialog, an overlay, a submission, or a text field owns the
+  // keystroke. Annotate-only commands layer their own conditions on top.
+  const canHandleDocumentChromeShortcut = useCallback((event: KeyboardEvent) => {
+    if (archive.archiveMode || goalSetupMode) return false;
     if (event.defaultPrevented) return false;
     if (document.querySelector('[data-plannotator-confirm-dialog="true"]')) return false;
     if (showExport || showImport || showFeedbackPrompt || showClaudeCodeWarning ||
@@ -997,7 +1002,6 @@ const App: React.FC = () => {
     const tag = target?.tagName;
     return tag !== 'INPUT' && tag !== 'TEXTAREA' && !target?.isContentEditable;
   }, [
-    annotateMode,
     archive.archiveMode,
     goalSetupMode,
     showExport,
@@ -1015,6 +1019,36 @@ const App: React.FC = () => {
     isExiting,
     isEditingMarkdown,
   ]);
+
+  const canHandleAnnotateSidebarShortcut = useCallback(
+    (event: KeyboardEvent) => annotateMode && canHandleDocumentChromeShortcut(event),
+    [annotateMode, canHandleDocumentChromeShortcut],
+  );
+
+  // Focus mode from the keyboard. Mirrors the document card's `Focus` control —
+  // including its availability — so the shortcut can never park the layout in a
+  // state with no visible way back. HTML surfaces are excluded because they own
+  // their own persisted chrome and never render that control.
+  const handleToggleFocusMode = useCallback(() => {
+    const action = resolveFocusShortcutAction({
+      canUseWideMode: canUseWideMode && !isHtmlSurface,
+      wideModeType,
+    });
+    if (action === 'enter-focus') enterViewMode('focus');
+    else if (action === 'exit') exitWideMode();
+  }, [canUseWideMode, enterViewMode, exitWideMode, isHtmlSurface, wideModeType]);
+
+  useDocumentViewShortcuts({
+    handlers: {
+      toggleFocusMode: {
+        when: (event) =>
+          canHandleDocumentChromeShortcut(event)
+          && !isHtmlSurface
+          && (canUseWideMode || wideModeType !== null),
+        handle: handleToggleFocusMode,
+      },
+    },
+  });
 
   useAnnotateSidebarShortcuts({
     handlers: {
@@ -4627,7 +4661,7 @@ const App: React.FC = () => {
                           <Tooltip
                             side="top"
                             align="end"
-                            content={type === 'wide' ? 'Hide panels and expand document width' : 'Hide panels, keep document width'}
+                            content={type === 'wide' ? 'Hide panels and expand document width' : `Hide panels, keep document width (${modKey}+.)`}
                           >
                             <button
                               type="button"
--- a/packages/editor/App.tsx
+++ b/packages/editor/App.tsx
@@ -1027,8 +1027,10 @@ const App: React.FC = () => {
 
   // Focus mode from the keyboard. Mirrors the document card's `Focus` control —
   // including its availability — so the shortcut can never park the layout in a
-  // state with no visible way back. HTML surfaces are excluded because they own
-  // their own persisted chrome and never render that control.
+  // state with no visible way back. HTML surfaces cannot ENTER focus mode (they
+  // own their own persisted chrome and never render that control), but exit stays
+  // available everywhere: a linked-doc navigation can flip the surface to HTML
+  // while focus mode is active, and without the exit path that layout is stuck.
   const handleToggleFocusMode = useCallback(() => {
     const action = resolveFocusShortcutAction({
       canUseWideMode: canUseWideMode && !isHtmlSurface,
@@ -1043,8 +1045,7 @@ const App: React.FC = () => {
       toggleFocusMode: {
         when: (event) =>
           canHandleDocumentChromeShortcut(event)
-          && !isHtmlSurface
-          && (canUseWideMode || wideModeType !== null),
+          && ((canUseWideMode && !isHtmlSurface) || wideModeType !== null),
         handle: handleToggleFocusMode,
       },
     },
--- a/packages/editor/shortcuts.ts
+++ b/packages/editor/shortcuts.ts
@@ -6,6 +6,7 @@ import {
   createShortcutRegistry,
   createShortcutScopeHook,
   defineShortcutScope,
+  documentViewShortcuts,
   goalSetupShortcuts,
   imageAnnotatorShortcuts,
   inputMethodShortcuts,
@@ -79,6 +80,7 @@ const annotateEditorSettingsShortcuts = defineShortcutScope({
 });
 
 const sharedPlanSurfaceShortcuts = [
+  documentViewShortcuts,
   inputMethodShortcuts,
   annotationToolbarShortcuts,
   viewerShortcuts,
--- a/packages/ui/components/KeyboardShortcuts.tsx
+++ b/packages/ui/components/KeyboardShortcuts.tsx
@@ -84,6 +84,13 @@ const inputMethodShortcuts: ShortcutSection = {
   ],
 };
 
+const documentViewShortcuts: ShortcutSection = {
+  title: 'View',
+  shortcuts: [
+    { keys: [modKey, '.'], desc: 'Toggle focus mode', hint: 'Collapses the Contents sidebar and the right-hand panel together; press again to restore whatever was open before.' },
+  ],
+};
+
 const annotationShortcuts: ShortcutSection = {
   title: 'Annotations',
   shortcuts: [
@@ -108,6 +115,7 @@ const imageAnnotatorShortcuts: ShortcutSection = {
 };
 
 const sharedPlanEditorShortcuts: ShortcutSection[] = [
+  documentViewShortcuts,
   inputMethodShortcuts,
   annotationShortcuts,
   imageAnnotatorShortcuts,
--- a/packages/ui/components/KeyboardShortcuts.tsx
+++ b/packages/ui/components/KeyboardShortcuts.tsx
@@ -87,7 +87,7 @@ const inputMethodShortcuts: ShortcutSection = {
 const documentViewShortcuts: ShortcutSection = {
   title: 'View',
   shortcuts: [
-    { keys: [modKey, '.'], desc: 'Toggle focus mode', hint: 'Collapses the Contents sidebar and the right-hand panel together; press again to restore whatever was open before.' },
+    { keys: [modKey, '.'], desc: 'Toggle focus mode', hint: 'Collapses the Contents sidebar and the right-hand panel together; press again to restore whatever was open before. Markdown documents only; HTML pages keep their own layout.' },
   ],
 };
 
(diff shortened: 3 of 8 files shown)


backnotprop (commented):
## TLDR

**Verdict: approve with follow-ups. No blockers.** `Mod+.` really is unclaimed on both plan surfaces, the guard refactor is behavior-preserving, and the enter/exit/restore machine is sound because every manual-open path already tears wide mode down with `restore: false`. A merge preview against current `main` (which now carries #1277) is clean: `bun test packages/ui packages/editor` 674 pass / 0 fail, `bun run typecheck` exit 0, and zero file overlap with #1277.

Five things worth a follow-up, none of which need to hold the merge: the help modal advertises the shortcut on raw-HTML annotate sessions where it is deliberately inert; the `!isHtmlSurface` term in the `when` guard is stricter than the resolver and blocks the one escape path from a reachable stuck state; the VS Code webview forwards the same chord up to VS Code's keybinding service; focus mode silently closes the Agent TUI and never restores it; and `.` is not shift-tolerant in the binding matcher, so the binding is unreachable on AZERTY.

---

## SHOULD FIX

### 1. The help modal advertises focus mode on raw-HTML annotate sessions, where it does nothing

`packages/ui/components/KeyboardShortcuts.tsx:87-92` adds the View section to `sharedPlanEditorShortcuts`, which `annotateShortcuts` composes (`:175-179`). The mode prop that selects that list is `packages/editor/components/AppHeader.tsx:353`:

```
mode={annotateMode ? 'annotate' : 'plan'}
```

There is no `renderAs` awareness. So a raw-HTML annotate session opens Settings, reads "Toggle focus mode / ⌘ ." and gets a dead key, because `packages/editor/App.tsx:1046` gates the handler on `!isHtmlSurface`. The exclusion is the right call, but the help UI now makes a claim that is false on exactly the surface the PR deliberately excluded.

Cheapest fix is a sentence in the hint ("Not available on raw HTML documents"), which also flows into the generated docs page since the registry hint is the Notes column. Gating the section on the surface would be better but needs a new prop.

The same caveat applies to `/docs/reference/keyboard-shortcuts`, which renders the annotate surface's View section with no such note.

---

## CONSIDER

### 2. `!isHtmlSurface` in `when` is stricter than the resolver, and blocks the only exit from a reachable stuck state

`packages/editor/App.tsx:1043-1050`:

```
when: (event) =>
  canHandleDocumentChromeShortcut(event)
  && !isHtmlSurface
  && (canUseWideMode || wideModeType !== null),
```

`resolveFocusShortcutAction` already returns `'exit'` for any non-null `wideModeType` regardless of availability (`packages/ui/utils/wideMode.ts:45`), and `wideMode.test.ts` has a dedicated case for it ("still restores when availability drops mid-focus"). The `when` guard's unconditional `!isHtmlSurface` means that case can never be reached through the HTML path.

That matters because the state is reachable. Enter focus mode on a markdown annotate document, then click a link to a `.html` doc: `useLinkedDoc` calls `setRenderAs(docRenderAs)` (`packages/ui/hooks/useLinkedDoc.ts:350`) while `wideModeType` stays `'focus'`. Nothing exits wide mode on that transition, the surface-transition chrome effect bails out early on `wideModeType !== null` (`packages/editor/App.tsx:1660`), and the Focus/Wide control is not rendered on HTML surfaces (`packages/editor/App.tsx:4651`, `&& !isHtmlSurface`). The user is left on an HTML page with no sidebar, no panel, no Focus control, and now no shortcut. The only way back is navigating back to the markdown document.

Reaching this through the Files sidebar is not possible (opening the sidebar routes through `openSidebarTab`, which exits wide mode), so it is specifically the in-document link path. It is pre-existing for the button, but the code comment right above the handler (`packages/editor/App.tsx:1029-1031`) says the mirroring exists "so the shortcut can never park the layout in a state with no visible way back", and here it declines to be the way back. One line:

```
&& ((canUseWideMode && !isHtmlSurface) || wideModeType !== null)
```

That also makes the tested availability-drop case reachable in production instead of test-only.

### 3. The VS Code webview forwards `Mod+.` up to VS Code in addition to handling it

The plan editor also runs inside a VS Code webview (`apps/vscode-extension/src/panel-manager.ts:23-33`). The injected proxy forwards nearly every keydown up to the webview document so VS Code's own keybinding service can resolve it (`apps/vscode-extension/src/cookie-proxy.ts:224-227`, re-dispatched at `panel-manager.ts:85-89`), excluding only the clipboard and undo chords (`cookie-proxy.ts:206-222`). So in that host, `Mod+.` is handled by the app **and** handed to VS Code.

In practice this is probably benign today: VS Code's default `editor.action.quickFix` on `Ctrl+.` is scoped to `textInputFocus` and will not fire from a webview, and the extension's own `ctrl+shift+.` / `cmd+shift+.` bindings (`apps/vscode-extension/package.json:53-59`) cannot collide because the matcher requires `requiresShift === event.shiftKey`. But any user keybinding on a bare `ctrl+.` with a loose `when` clause will now double-fire alongside focus mode. Worth a line in the PR notes at minimum, or an addition to the proxy's exclusion list.

### 4. Focus mode closes the Agent TUI and never restores it

`enterViewMode` hides the agent terminal (`packages/editor/App.tsx:958`), but `WideModeLayoutSnapshot` records only `sidebarIsOpen` / `sidebarTab` / `panelOpen` (`packages/ui/utils/wideMode.ts:5-9`), so exiting restores the two sidebars and leaves the terminal closed. The PR body says the second press "restores the exact prior arrangement"; that holds for the two sidebars, not for the Agent TUI.

Pre-existing behavior of the Focus button, but a keyboard binding makes it far easier to hit by accident: `Shift Shift` opens the terminal, `Mod+.` closes it with no undo. Either add `agentTerminalOpen` to the snapshot, or soften the PR/help wording.

### 5. `.` is not shift-tolerant, so the binding is unreachable on AZERTY

`matchesShortcutBinding` requires `requiresShift === event.shiftKey` (`packages/ui/shortcuts/core.ts:402-407`), and `.` is not in `SHIFTED_LITERAL_TOKENS` (`packages/ui/shortcuts/core.ts:79`, currently `{ '{', '}', '?', '$' }`). On French and Belgian AZERTY the period is `Shift+;`, so `Ctrl+.` arrives with `shiftKey: true` and never matches. That is the same class of silent failure the PR body cites when rejecting `Mod+\` for German and Nordic layouts, so the "unshifted on every layout" claim in the rationale is not quite right.

The fix is one token, and the existing `event.key === keyToken` term keeps it safe (US `Shift+.` yields `>`, not `.`, so it still will not match):

```
const SHIFTED_LITERAL_TOKENS = new Set(['{', '}', '?', '$', '.']);
```

Worth doing at the engine level rather than here, since it also fixes the existing code-review `Mod+.` binding (`packages/review-editor/shortcuts.ts:70-74`).

Two related matcher facts, both fine but worth knowing: `requiresMod !== (event.metaKey || event.ctrlKey)` (`core.ts:406`) does not distinguish Meta from Control, so on macOS both `Cmd+.` and `Ctrl+.` enter focus mode; and `matchesKeyToken` falls through to `key === token`, so the numpad decimal key matches too.

### 6. The hint string now lives verbatim in two hand-maintained places

`packages/ui/shortcuts/plan-review/documentView.shortcuts.ts:19` and `packages/ui/components/KeyboardShortcuts.tsx:90` carry the same 152-character hint character-for-character, and nothing asserts parity. Only the vim block in that component is registry-generated (`KeyboardShortcuts.tsx:124-140`); the rest is hand-maintained, so the PR is following the local convention rather than breaking it. Still, the View section is a one-scope, one-action section, which makes it the cheapest possible candidate to generate the way `vimShortcuts` is generated. Optional, and arguably a separate cleanup.

### 7. Tooltip formatting nit

`packages/editor/App.tsx:4664` renders `(${modKey}+.)`, which is `(⌘+.)` on mac; the platform convention is `⌘.` with no separator (the help modal's `Kbd` renderer already does this correctly by rendering separate keycaps). Also, the Wide control's tooltip carries no shortcut hint even though `Mod+.` exits wide mode too, which is a small asymmetry given the resolver deliberately treats `wide` as exitable.

---

## VERIFIED-OK

### `Mod+.` is genuinely unclaimed on these surfaces

I audited this independently of the PR description, across the registry, the ad-hoc listeners and the third-party keymaps.

**Registry:** the only two bindings in the repo that involve `.` are `packages/ui/shortcuts/plan-review/documentView.shortcuts.ts:17` (this PR) and `packages/review-editor/shortcuts.ts:72` ("Toggle review sidebar", section Layout). Different apps, different registries, no shared listener, so no collision, and the idiom is consistent rather than conflicting. No registry scope binds a bare `.` either.

**Ad-hoc listeners:** all 25 global `keydown` listeners across `packages/editor` and `packages/ui` were checked. The only one anywhere in the repo that compares against `.` is `packages/review-editor/App.tsx:1543`, which is the code-review surface. The handlers that forward arbitrary printable keys all bail on modifiers first, so none can intercept `Mod+.`: `AnnotationToolbar.tsx:141`, `CommentPopover.tsx:266`, `FloatingQuickLabelPicker.tsx:78`, and both html-viewer bridge handlers (`bridge-script.ts:3793` and `:4064`). The rest are `Escape`-only, or gated on `Enter` / `S` / `P`.

**Third-party keymaps:** CodeMirror (`@codemirror/commands`, `search`, `autocomplete`, `view`, `lang-markdown`) declares no `Mod-.` binding; `@plannotator/atomic-editor` binds only `Mod-b/i/e/k`, `Mod-Shift-x`, `Ctrl-Space`; `@base-ui/react` does not handle `Period`. Monaco, ProseMirror, tiptap and Radix are not in the plan/annotate dependency graph, and dockview is code-review only. Nothing to collide with.

**Agent terminal:** with the annotate agent terminal focused, `Mod+.` does nothing at all, and neither does `Mod+B`. `@plannotator/webtui`'s shortcut layers claim only `c/v/k` and the zoom keys, and xterm leaves keyCode 190 unmapped so it does not `preventDefault`; the event does reach `window`, but xterm's hidden helper `<textarea>` holds focus so the chrome guard rejects it at `packages/editor/App.tsx:1003`. That is consistent with how the pre-existing sidebar shortcuts already behave, so it is a documentation nuance rather than a regression, but it is worth knowing the PR's "unclaimed in every state" audit has this hole.

**One latent trap, not caused by this PR:** `packages/ui/components/AnnotationPanel.tsx:618` calls `e.stopPropagation()` unconditionally on the card-footer slot, which under React 19's root listener would kill the `window` listener for any key including `Mod+.`. It is dormant because `renderCardFooter` has no caller today, but it will bite whichever host fills that slot first.

### The dispatcher does not reject editable targets, and the PR supplies that itself

`dispatchShortcutEvent` (`packages/ui/shortcuts/runtime.ts:71-101`) has no target-tag check and no cross-scope arbitration; the TODO at `:64-70` says so explicitly and requires callers to guard with `when`. The PR does: `canHandleDocumentChromeShortcut` rejects `INPUT` / `TEXTAREA` / `isContentEditable` (`packages/editor/App.tsx:1001-1003`) plus dialogs, overlays, submission states and markdown editing. Importantly `when` is evaluated **before** `preventDefault` (`runtime.ts:85-90`), so `Ctrl+.` typed into a comment editor is not swallowed. The xterm agent terminal focuses a helper `textarea`, so it is covered by the same tag check.

### Guard refactor preserves the existing `Mod+B` / `Mod+Shift+B` / `Shift Shift` behavior

Old predicate opened with `if (!annotateMode || archive.archiveMode || goalSetupMode) return false;`. New form is `annotateMode && canHandleDocumentChromeShortcut(event)` (`packages/editor/App.tsx:1023-1026`) with every remaining condition unchanged and no side effects anywhere in the chain, so short-circuit reordering is unobservable. All three consumers (`:1056`, `:1060`, `:1071`) keep their extra terms. Dependency arrays are right: `annotateMode` drops out of the base callback because it is no longer referenced, and appears on the wrapper.

### Enter / exit / restore state machine

The invariant that makes this safe is that `wideModeType !== null` implies both panels are hidden. Every manual-open path while a view mode is active tears it down with `restore: false` rather than opening on top of it: `openSidebarTab` (`:613`), `toggleSidebarTab` (`:621`), `handleAnnotationPanelToggle` (`:629`), `handleAIChatToggle` (`:655`), `openAgentTerminal` (`:697`), `openAIChat` (`:3664`). So there is no state where a manual open leaves a stale snapshot for a later `Mod+.` to restore incorrectly. `enterViewMode` only snapshots when `wideModeType === null` (`:951`), so a `wide` -> `focus` swap and the shortcut's exit share one snapshot, and `resolveWideModeExitLayout`'s null-snapshot fallback (`packages/ui/utils/wideMode.ts:63-67`) closes both, which is the correct degenerate answer.

The `'none'` branch is unreachable while `when` passes (`when` already requires `canUseWideMode || wideModeType !== null`, and the resolver only returns `'none'` when both are falsy), so the shortcut never `preventDefault`s a keystroke it will not act on. Archive mode, goal setup and plan-diff view all correctly fall through to the browser: the first two via the chrome guard, the third because `canUseWideMode` goes false and the auto-exit effect (`:972-976`) has already cleared `wideModeType`.

### Shared-scope choice is right

The Focus/Wide control is not gated on `annotateMode` (`packages/editor/App.tsx:4651`) and `canUseAnnotateWideMode` is surface-agnostic, so plan review renders the same control over the same two panels: left TOC/Versions/Archive sidebar, right annotations/AI panel. Issue #1276 asked only for annotate; extending to plan review matches the existing control's availability rather than inventing new behavior, and the registry test asserts it on both surfaces.

### Registry test quality

`packages/ui/shortcuts.test.ts:116-131` names the failure it catches (a second claimant double-firing, which is accurate given `runtime.ts:64-70`) and asserts uniqueness rather than snapshotting a list. One honest limitation to keep in mind: it can only see registry claimants, so a future ad-hoc `window` keydown listener would still collide invisibly. The section-title `toEqual` extensions at `:74-95` are structural, not prose snapshots, so they are within the Testing Rules.

`wideMode.test.ts:39-77` is fine: three of the five cases name a real regression in comments (wide must exit not swap, availability dropping mid-focus). The other two are the remaining truth-table rows of a two-line function, which is close to the "implementation echo" line, but a pure function's return value is the observable outcome, so I would not change them.

### Marketing docs auto-generation

Picked up automatically, no marketing change needed. `apps/marketing/src/lib/shortcutReference.ts:8-16` iterates `allSurfaces` and calls `listRegistryShortcutSections`, and nothing there hardcodes scope ids or section titles, so composing into the existing surfaces is enough. Section order comes from first appearance across scope order (`packages/ui/shortcuts/core.ts:241-257`), giving `Actions, View, Input Method, ...` for plan review and `Actions, Sidebar, View, ...` for annotate, exactly what the test asserts. `ShortcutReference.astro` calls `formatShortcutBindingsText` with no platform argument, so the docs render `Cmd/Ctrl+.` with no build-host dependence, and `.` passes registry validation because it is whitelisted in `NAMED_TOKENS` (`core.ts:64`), so no build-time throw. Anchors `plan-review-view` and `annotate-mode-view` are distinct. The 152-character hint will be the tallest Notes cell on the page but wraps normally.

### No collision with #1277

`git show --name-only 8e7b5ce3` intersected with this PR's file list is empty. #1277 did not touch `packages/ui/shortcuts/index.ts`, `packages/ui/components/KeyboardShortcuts.tsx`, `packages/editor/App.tsx` or `packages/editor/shortcuts.ts`. Merge preview of `pr/1279` + `origin/main` applied with no conflicts, `bun test packages/ui packages/editor` gave 674 pass / 387 skip / 0 fail, and `bun run typecheck` exited 0.

---

PR #1280, opened 2026-08-12
feat(remote): tailnet auto-advertise, ready QR code, and a first-class --tailscale mode

backnotprop:
**TLDR:** Three small pieces that make Plannotator sessions reachable from another device over a tailnet with zero env-var soup: `PLANNOTATOR_URL_HOST=auto` detects the tailnet host for advertised URLs, remote-ready messages now print a scannable terminal QR code, and a first-class `--tailscale` flag publishes a loopback-bound review/annotate session over `tailscale serve` with HTTPS and checked cleanup.

*(Updated after two reviews: the external review and the internal security review. See the review-response comment for the per-finding mapping.)*

## The two user asks

1. *"id love to be able to use... Plannotator Review skill and being able to actually open that UI on my iPad to provide review notes, without actually getting to my mac"* - run `plannotator review --tailscale` on the mac, scan the QR from the iPad, review over HTTPS. No remote mode, no port forwarding, no hand-typed URL.
2. *"the problem is have multiple vpses, so its quite hard to ssh into each and map back... sometimes you just wanna get that diff out of the vm back to your 'dummy' machine"* - each VPS runs `--tailscale` (or remote mode with `urlHost: "auto"`); every session advertises its own reachable tailnet URL on a distinct random port, so simultaneous sessions across VPSes never collide and nothing needs SSH mapping. This productizes the community gist POC (`/plannotator-review-remote`) that users were maintaining by hand.

## What each piece does

- **`urlHost: "auto"`** (`PLANNOTATOR_URL_HOST=auto` or config `{ "urlHost": "auto" }`): in a remote session, the advertised host resolves once per process from `tailscale status --json` (`Self.DNSName`, trailing dot stripped) with a fallback to the single `tailscale ip -4` CGNAT (100.64.0.0/10) address, validated strictly. Detection failure warns once on stderr and falls back to `localhost`. Local sessions never spawn the CLI. The cached resolution lives in the shared vendored module, used identically by the Bun runtime and the Pi mirror.
- **Ready QR**: `handleServerReady` renders a compact unicode QR (zero-dependency `uqr`) under the remote-ready URL line, TTY-gated, and only when the advertised host is overridden away from localhost (a QR of a localhost URL scans to nowhere).
- **`--tailscale`** on `plannotator review`, `annotate`, and `annotate-last`/`last` (other subcommands reject the flag with a clear error; all four help texts document it): keeps the server loopback-bound, then orchestrates `tailscale serve --bg --https=<port> http://127.0.0.1:<port>`, extracts the HTTPS URL matching the session port, and advertises it with the QR. Publishing failures print an actionable message and exit 1 instead of leaving the loopback server hanging (CLI missing, daemon down or logged out, unparsable status output, no URL for our port). The conflict check inspects both background (`TCP`) and foreground (`Foreground.*.TCP`) serve mappings, refuses to steal either, never touches other ports, and fails closed on unrecognizable `serve status` output. Plan mode is hook-driven with no user flags, so it deliberately has no `--tailscale`.

## Cleanup lifetime (honest version)

Mappings the process creates are cleaned up on normal exit and on SIGINT/SIGTERM/SIGHUP (all routed through `process.exit` so exit handlers run). A failed `off` retries once, then warns with the exact manual command, and a port is only forgotten after a successful `off`. A hard kill (SIGKILL) or reboot can leave the mapping - `tailscale serve --bg` state persists by design - and the remedy is always the same documented command: `tailscale serve --https=<port> off`.

## Security: what did NOT change

- **Binding posture is unchanged.** `--tailscale` binds loopback only; reachability and TLS come from `tailscale serve` proxying 127.0.0.1. Combined with `PLANNOTATOR_REMOTE`/SSH detection, `--tailscale` wins and forces local mode with a stderr notice, because the wide `0.0.0.0` bind would only broaden exposure. Remote mode's bind behavior is untouched.
- **`urlHost` stays strictly display-only**, including `auto`: detection changes what is printed, never which interface is bound. `--tailscale` runs suppress `urlHost` entirely (the advertised URL comes from serve).
- **Agent terminal is off by default under `--tailscale`**, gated behind the same existing `PLANNOTATOR_AGENT_TERMINAL_REMOTE=1` opt-in remote mode uses: the PTY token is not an auth boundary against network peers, so tailnet reachability implies terminal reachability. No new setting was invented.
- **No funnel, no public exposure, ever.** Only `tailscale serve` (tailnet-only) is orchestrated; no secrets ride in advertised URLs.

## Tests

- `packages/shared/tailscale.test.ts`: status-JSON DNSName parsing, strict CGNAT IPv4 validation, port-matched serve-URL extraction (including the 443 default and other-port rejection), background AND foreground conflict detection with fail-closed malformed-output fixtures, exact serve/teardown argv construction, injected-runner detection fallbacks.
- `packages/server/tailscale-serve.test.ts`: publish + teardown round trip, refusal to steal background and foreground mappings, fail-closed status handling, port-mismatch self-teardown, off-retry and manual-command warning with the port kept for exit-time retry, missing-CLI hint. All runners injected; module and process state restored via a test reset seam.
- Full `packages/shared` + `packages/server` suites: 1492 pass, 0 fail. `apps/hook` + Pi `network.test.ts`: 227 pass. `bun run typecheck` (all packages incl. Pi vendored output): green.
- **Live tailnet verification: skipped** (no `tailscale` CLI on the build machine). Compiled-CLI verification against PATH shims instead: missing-binary startup exits 1 with the install hint in under a second (no hang), foreground-conflict status aborts with the exact message, SIGTERM and SIGHUP both issue `serve --https=<port> off`, a persistently failing `off` produces two attempts plus the manual-command warning, and `archive --tailscale` is rejected.

## Pi parity

Pi mirrors the `urlHost: "auto"` logic through the shared vendored module (now a single implementation, not a copy). Two deliberate divergences: no QR on Pi (its ready surface is an in-chat notification, not a TTY stream) and no `--tailscale` on Pi (sessions start inside the extension; there is no CLI process owning serve lifecycle and signal-routed cleanup).

One judgment call worth flagging: annotate strict-gate client leases stay enabled under `--tailscale` (it is a local session). The lease's 30s reconnect grace plus SSE heartbeats should tolerate the serve proxy fine, but if field reports show proxy drops reading as abandonment, the lease can be gated off for `--tailscale` in one line.

*AI-assisted: implemented and tested by Claude under maintainer direction.*


backnotprop (commented):
## Security review of #1280

**TLDR: Request changes.** The core security claims hold up under adversarial reading and live testing: the bind never widens, `urlHost=auto` stays display-only, only `serve` (never funnel) is orchestrated, pre-existing mappings are refused rather than stolen, and cleanup fires on SIGINT/SIGTERM/normal exit/uncaught errors. But one claim is empirically false: a `--tailscale` precondition failure does NOT stop the session. It prints the error and then hangs the CLI forever with the loopback server still up (reproduced against the compiled binary on this machine, where tailscale is not installed). That is a blocker. Two should-fixes: SIGHUP leaks the serve mapping (reproduced), and the annotate agent terminal skips the gate that remote mode applies to the exact same exposure class.

All findings verified against a merge preview of `feat/tailscale-remote` (6b47278b) onto `main` (8e7b5ce3). Test suites: 1486 pass / 0 fail (`packages/shared` + `packages/server`, after `vendor.sh`), Pi `network.test.ts` 31 pass, `bun run typecheck` green, compile smoke of the CLI binary green (built and executed here).

---

### BLOCKER

**1. Every `--tailscale` failure path hangs the session instead of aborting.**

`handleTailscaleReady` (`apps/hook/server/index.ts:246`) claims "Throwing here propagates through onReady, which stops the just-started server." It does not. `onReady` is typed `(url, isRemote, port) => void` and invoked without await inside `startAnnotateServer` (`packages/server/annotate.ts:989`) and `startReviewServer` (`packages/server/review.ts:3204`), while the CLI top level is awaiting that same start function. In that position Bun reports the rejection but keeps running (verified in isolation and end to end).

Reproduced with the compiled binary, no tailscale on PATH:

```
$ plannotator annotate t.md --tailscale
error: --tailscale: `tailscale` CLI not found on PATH. ...
      at enableTailscaleServe ...
[process still alive 8s later; server listening; killed manually]
```

The actionable error text does print, but the process then sits at `waitForDecision` forever. Under the `/plannotator-review` skill (bang-prefix bash) this hangs the whole prompt. Every precondition failure takes this path: CLI missing, daemon down or logged out, port conflict, unparsable serve output.

Fix is small: wrap the tailscale branch in each of the three `onReady` callbacks (`apps/hook/server/index.ts:982`, `:1200`, `:1442`):

```ts
if (tailscaleFlag) {
  try {
    await handleTailscaleReady(port);
  } catch (err) {
    console.error(err instanceof Error ? err.message : String(err));
    process.exit(1);
  }
  return;
}
```

`process.exit` runs the exit handlers, so a mapping created before a later failure is still torn down. Please add a test that pins the abort (the current suites only cover `enableTailscaleServe` throwing, not what the CLI does with the throw).

### SHOULD FIX

**2. SIGHUP leaks the serve mapping, leaving a standing tailnet-to-loopback proxy.**

`apps/hook/server/index.ts:425-426` routes only SIGINT and SIGTERM through `process.exit`. SIGHUP (closing the terminal window, the most common way a forgotten session dies) takes the default death and skips exit handlers. Reproduced with a PATH shim: after `kill -HUP`, the shim log shows `serve status --json` and `serve --bg --https=<port> ...` but no `serve --https=<port> off`.

The leaked mapping is not just clutter: it is a persistent tailnet HTTPS listener proxying to `127.0.0.1:<port>`, and since ports are random ephemeral ports, whatever unrelated local process later binds that port becomes silently reachable from the tailnet. One line fixes it, next to the existing pair:

```ts
process.once("SIGHUP", () => process.exit(129));
```

(This also benefits the pre-existing worktree-warmup cleanup that motivated the SIGINT/SIGTERM routing.) SIGKILL and hard crashes still leak; that residual is fair to accept and is already covered by the "best effort" comment in `cleanupAllServeMappings`.

**3. The annotate agent terminal skips the gate remote mode applies to the same exposure.**

`createBunAgentTerminalBridge` disables the PTY in remote sessions unless `PLANNOTATOR_AGENT_TERMINAL_REMOTE=1` (`packages/server/agent-terminal.ts:48`), precisely because a beyond-loopback session hands the terminal to the network. `--tailscale` creates the same beyond-machine reachability, but forces `PLANNOTATOR_REMOTE=0`, so `isRemoteSession()` is false and the terminal is on by default.

The token is not an authentication boundary against network peers: the `wsPath` (18 random bytes) is served to every client in the `/api/plan` capability payload (`packages/server/annotate.ts:503`), and `isAllowedOrigin` returns true for requests with no Origin header (`packages/server/agent-terminal.ts:325-327`), so any non-browser client that can reach the port can fetch the token and drive an interactive agent CLI in the project cwd. On a single-user personal tailnet that is the user's own device, as the PR states. On shared tailnets, work tailnets, or default allow-all ACLs, "tailnet peer" is not "the user", and the terminal is effectively code execution as the user.

Concrete recommendation, before merge: treat `--tailscale` as beyond-loopback for this one gate. Reuse the existing opt-in rather than inventing a new knob, e.g. have the CLI set an internal marker env (or thread a flag into `startAnnotateServer`) when `tailscaleFlag` is set, and extend the check at `agent-terminal.ts:48` to `(isRemoteSession() || isTailscaleSession()) && !isAgentTerminalRemoteEnabled()`, with the disabled-reason message mentioning `PLANNOTATOR_AGENT_TERMINAL_REMOTE=1`. That preserves the iPad review flow (the review UI itself is unaffected) and keeps the highest-impact surface opt-in, consistent with the posture remote mode already established. A future hardening option worth a comment: `tailscale serve` injects `Tailscale-User-Login` identity headers on proxied requests, which would allow restricting the PTY to the node owner without any env var.

**4. Verify the extracted serve URL's port before advertising it.**

`extractServeHttpsUrl` (`packages/shared/tailscale.ts:163`) takes the first `https://` match in stdout+stderr. Whether `tailscale serve --bg` prints only the new mapping or the whole serve config is version-dependent output shape, not a contract. If the output ever includes another pre-existing mapping first (e.g. one on 443), the advertised URL and QR would point at the wrong service. Since `--https=<port>` always uses an explicit non-443 port, the check is one line: after parsing, require `new URL(candidate).port === String(port)`, and on mismatch take the same self-teardown path as the no-URL case. Cheap insurance against CLI drift.

### CONSIDER

**5. `serveStatusHasPort` fails open in two corners.** (a) Exit-0 with unparsable non-empty JSON returns false, so the refusal is bypassed and `serve --bg` would clobber, then exit teardown would delete, a mapping the user owned. Treating unparsable non-empty status as a conflict fails closed. (b) The check only reads top-level `TCP`; a foreground `tailscale serve` running in another terminal nests under `Foreground.<sessionId>.TCP` and would be missed. Both are unlikely on a random high port; noting for completeness (`packages/shared/tailscale.ts:145-158`).

**6. `--tailscale` silently no-ops on subcommands where it is not wired.** The flag is parsed and stripped globally (`apps/hook/server/index.ts:227`) and `PLANNOTATOR_REMOTE` is forced to `0`, but only `review`/`annotate`/`annotate-last` check `tailscaleFlag`. `plannotator archive --tailscale`, `setup-goal`, or a hand-invoked bridge subcommand (`copilot-last`, `opencode-review`) strips the flag, forces local mode, publishes nothing, and opens the local browser. An "unsupported with this subcommand" error would be cheaper to debug than silence.

**7. QR prints for remote sessions advertising plain `localhost`.** In the port-forward case (`shared-handlers.ts:255-259`, no urlHost override) the QR encodes `http://localhost:<port>`, which is useless on another device. Gating `writeUrlQr` on `isUrlHostOverridden()` in that branch would keep the QR meaningful; harmless either way.

**8. Confusing warning under `--tailscale` + configured `urlHost`.** A user with `{ "urlHost": "auto" }` in config (their VPS setup) running `--tailscale` on a laptop gets the local-session warning advising `PLANNOTATOR_REMOTE=1` (`packages/server/remote.ts:203-207`), which is the opposite of what `--tailscale` wants. Cosmetic; a `--tailscale`-aware suppression or wording tweak would do.

**9. No test covers the `auto` integration at the `buildAdvertisedUrl` layer.** `resolveAutoHost` calls `detectTailnetHost()` without runner injection, so `remote.ts`/Pi `network.ts` auto behavior (remote-only resolution, once-per-process warning, localhost fallback) is asserted only by code reading. The shared parsers are well tested. Optional, but an injectable seam would let the display-only invariant be pinned where it lives.

**10. Client lease over the serve proxy.** The PR flags that strict-gate leases stay enabled under `--tailscale` and that proxy drops could read as abandonment. Agreed with the wait-for-field-reports stance; the 30s grace plus 5s heartbeats should survive tailscale's local proxy.

### VERIFIED-OK

- **Bind never widens.** `getServerHostname()` is the only bind source for all Bun servers (`packages/server/{index.ts:206,annotate.ts:442,review.ts:1541,goal-setup.ts:104}`), keyed solely on `isRemoteSession()`, which reads env fresh on every call (no memoization), so forcing `PLANNOTATOR_REMOTE="0"` at `index.ts:234` before any bind decision is effective. Verified live: `PLANNOTATOR_REMOTE=1` + `--tailscale` printed the loopback notice and bound a random local port.
- **`urlHost` stays display-only, including `auto`.** No code path feeds `resolveUrlHost`/`resolveAutoHost` output into a bind; local sessions return `localhost` before the auto branch is reached (`remote.ts:201-210`), so a local session never spawns the tailscale CLI; detection failure caches, warns once, and falls back to localhost.
- **No funnel, ever.** The only argv shapes constructed are `serve --bg --https=<port> http://127.0.0.1:<port>` and `serve --https=<port> off`, pinned by tests (`packages/shared/tailscale.test.ts` "serve command construction").
- **No steal/clobber on the honest path.** Pre-existing TCP mapping on the chosen port aborts before any `serve` call, with nothing torn down (test-pinned in `tailscale-serve.test.ts`); teardown only ever targets ports in `activePorts`, which only this process's successful publishes populate.
- **Cleanup on the claimed exit paths.** Signal routing through `process.exit(130/143)` is real (`index.ts:425-426`); verified empirically that Bun runs `process.on("exit")` handlers on `process.exit`, uncaught exception, and unhandled rejection; live shim run showed exactly `serve --https=<port> off` on SIGTERM. The unparsable-URL path tears its own port down before throwing (`tailscale-serve.ts:78-82`).
- **CLI edge robustness.** `spawnSync("tailscale", args)` with argv array (no shell), `windowsHide`, and 3s/10s timeouts; ENOENT and ETIMEDOUT mapped to actionable messages. CGNAT validation is correct for 100.64.0.0/10 (first octet 100, second 64-127, leading-zero and out-of-range rejection); `ip -4` requires exactly one address; `Self.DNSName` is trimmed, dot-stripped, and validated through the existing fail-closed `isValidUrlHost` (whitespace/scheme/port injection rejected, test-pinned). Logged-out daemon, malformed JSON, and IPv6-only tailnets all degrade to the once-per-process warning plus localhost; a display setting never breaks a launch.
- **QR degradation.** TTY-gated (`qr.ts:13`), try/caught, additive after the URL line; piped stderr and hook transcripts keep plain text.
- **uqr dependency.** `uqr@0.1.3`, unjs org, MIT, genuinely zero dependencies (lockfile entry `{}` and installed package.json confirm), pure JS dual ESM/CJS. Bundles to ~21.5KB and compiles into the single binary; I rebuilt the compiled CLI from the merge preview and exercised the QR path live.
- **Pi parity.** `apps/pi-extension/server/network.ts` mirrors `remote.ts` line for line on the auto path; `vendor.sh` vendors `tailscale.ts` and the generated import resolves (typecheck covers it). The declared divergences are clean absences: zero references to `uqr`, `qr`, or serve orchestration anywhere in the Pi tree. No new HTTP endpoints, so the two-runtime server-API law is not implicated.
- **Overbuild check.** Proportionate to spec: one shared parser module, one small orchestrator, no new config knobs, no funnel/identity machinery, tests target real regressions (argv drift, steal refusal, cleanup). The exported `disableTailscaleServe` is test-plus-symmetry rather than dead weight.

### Security summary

The design is sound: reachability comes from tailscaled's own authenticated proxy while the app keeps its strictest binding posture, which is strictly better than remote mode's raw `0.0.0.0`. What `--tailscale` does change is the audience of an unauthenticated session UI: every endpoint (diff read, `/api/git-add` staging, feedback, Ask AI) becomes reachable by whatever the user's tailnet ACLs permit, and that is the accepted premise of the feature. The one place the PR is more permissive than the project's own precedent is the annotate agent terminal (finding 3): remote mode default-gates it for exactly this exposure class, `--tailscale` does not. Gate it, fix the failure-path hang (finding 1) and the SIGHUP leak (finding 2), and the security story is consistent end to end.

**Merge readiness:** not yet. Finding 1 is a hard blocker (every misconfiguration hangs the agent), findings 2 and 3 are small and should land in this PR; everything else can follow.

backnotprop:
Thanks to both reviewers - the external review and the internal security review. All three P1s, the additional internal findings, and the P2s are addressed in `edd4e1bc`. Per-finding mapping:

**P1: startup failures hung the session** (found by both reviews) - fixed with one coherent mechanism: `startReviewServer` / `startAnnotateServer` now await async ready handlers and stop the server on rejection (`packages/server/review.ts:3224`, `packages/server/annotate.ts:1012`, matching the plan server's existing pattern), and the CLI's `--tailscale` ready path resolves publishing failures itself with a clean actionable message and `process.exit(1)` (`apps/hook/server/index.ts:268`), so the bang-prefix skill can never be blocked by a hanging loopback server. Verified with the compiled CLI: missing `tailscale` binary now exits 1 with the install hint in under a second on all three command paths.

**P1: teardown failures silently leaked the mapping** - `runServeOff` (`packages/server/tailscale-serve.ts:41`) checks the CLI result, retries once, and warns with the exact manual command (`tailscale serve --https=<port> off`); a port is only forgotten after a successful off, so exit-time cleanup retries anything still registered. No reaper or ownership system was added.

**P1: foreground serve mappings were invisible to the conflict check** (found by both reviews) - `checkServeStatusPort` (`packages/shared/tailscale.ts:158`) now inspects `Foreground.<sessionId>.TCP` alongside top-level `TCP`, and fails CLOSED: unrecognizable or malformed status output aborts with a clear error instead of assuming the port is free. Fixtures cover foreground configs, malformed JSON, and the honest empty forms (`{}`, `null`).

**Internal: SIGHUP leaked the mapping** - terminal close now routes through `process.exit(129)` next to the existing SIGINT/SIGTERM routing (`apps/hook/server/index.ts:451`); shim-verified that SIGHUP issues the off command.

**Internal: agent terminal exposure under --tailscale** - the annotate agent terminal is now off by default in `--tailscale` sessions, gated behind the existing `PLANNOTATOR_AGENT_TERMINAL_REMOTE=1` opt-in (`packages/server/agent-terminal.ts:61`, threaded as `tailnetPublished` from the CLI). No new setting; AGENTS.md updated.

**Internal: extracted serve URL not verified against the requested port** - `extractServeHttpsUrl` now takes the expected port and skips https URLs for other ports (`packages/shared/tailscale.ts:199`), so version-dependent serve output cannot advertise another mapping's URL; mismatch takes our own mapping back down and errors.

**P2s**: tailscale-serve tests restore module and process state via `resetTailscaleServeForTests` in `afterEach` (no leaked exit listener); the duplicated `resolveAutoHost` collapsed into shared `resolveAutoHostCached` (`packages/shared/tailscale.ts:225`), used by both runtimes; `--tailscale` appears in `review` / `annotate` / `annotate-last` / top-level help and is rejected with a clear error on unsupported subcommands (`apps/hook/server/index.ts:228`). Trivial CONSIDERs applied: the QR renders only for URLs reachable off-machine (never localhost), and `--tailscale` runs suppress `urlHost` so the misleading local-session warning cannot fire.

**Honesty change on the cleanup-lifetime claim**: AGENTS.md and the PR body no longer say cleanup is guaranteed on every exit path. The accurate statement is: cleaned up on normal exit and SIGINT/SIGTERM/SIGHUP, with retry and a manual-command warning on failure; `tailscale serve --bg` mappings survive a hard kill (SIGKILL) and reboots by design, and the documented remedy is `tailscale serve --https=<port> off`.

Tests: 1492 pass in `packages/shared` + `packages/server`, 227 in `apps/hook` + Pi network tests, full typecheck green. Live tailnet verification remains skipped (no `tailscale` on the build machine); all of the above was verified end-to-end against PATH shims with the compiled CLI.

*AI-assisted: fixes by Claude under maintainer direction.*

backnotprop (commented):
## Delta review of edd4e1bc

**TLDR: All findings from my review are RESOLVED, verified empirically against a rebuilt binary, not just by reading the diff. No new blockers. This is mergeable from my side.**

Verification basis: merge preview of `edd4e1bc` onto current `main` (1bf90a23). `bun test packages/shared packages/server`: 1494 pass / 0 fail. Pi `network.test.ts`: 31 pass. `bun run typecheck` (incl. Pi vendored output): green. Compiled the CLI binary from the preview and re-ran my original repros.

### Per-finding status

**1. BLOCKER (startup hang): RESOLVED.** Re-ran the exact repro that hung before (no tailscale on PATH): `annotate t.md --tailscale` now exits `1` in under a second with the clean one-line install hint, no stack trace, no listening server left behind. Same for `review --tailscale` in a git repo. The mechanism is coherent in layers: `handleTailscaleReady` catches and does message + `process.exit(1)` (`apps/hook/server/index.ts:260-268`), which runs exit handlers (session unregister, worktree cleanup, serve cleanup); independently, both servers now await async `onReady` and `stop()` + rethrow on rejection (`packages/server/annotate.ts:1006-1016`, `review.ts:3218-3228`, with `stop` hoisted so the rejection path reuses the full teardown incl. `onCleanup`), so any future async onReady user cannot recreate the hang. All three command paths route through the same `handleTailscaleReady`; the other subcommands can no longer carry the flag at all (see 6). No remaining unhandled-rejection path found.

**2. SIGHUP leak: RESOLVED.** `process.once("SIGHUP", () => process.exit(129))` at `index.ts:451`, placed with the SIGINT/SIGTERM pair ahead of any server start. Verified with the PATH shim: `kill -HUP` on a published session now issues exactly `serve --https=<port> off` before death. Placement composes with the teardown retry: exit-time cleanup goes through the same `runServeOff` retry + warn path.

**3. Agent terminal gate: RESOLVED.** `tailnetPublished` threads from `tailscaleFlag` into both CLI `startAnnotateServer` sites (`index.ts:1224`, `:1467`) and gates in `createBunAgentTerminalBridge` behind the same `PLANNOTATOR_AGENT_TERMINAL_REMOTE` opt-in remote mode uses (`agent-terminal.ts:61-67`). Verified live against the shim-published session: without the opt-in, `/api/plan` reports the terminal disabled with the new tailscale-specific message; with `PLANNOTATOR_AGENT_TERMINAL_REMOTE=1` the tailnet gate clears (session then proceeds to the normal runtime checks). Airtight as far as I can trace: the PTY only exists in annotate servers, both reachable annotate call sites pass the flag, review/plan servers have no PTY, and every other subcommand rejects `--tailscale`. AGENTS.md rows for the env var and the tailnet section both match the behavior.

**4. URL/port verification: RESOLVED.** `extractServeHttpsUrl(output, expectedPort)` iterates all matches and requires `Number(url.port || "443") === expectedPort` (`packages/shared/tailscale.ts:186-199`); 443-default handling is correct and test-pinned ("port 443 URLs carry no explicit port"). Mismatch takes the self-teardown path and the teardown is asserted in the updated serve test ("rejects serve output whose only https URL is for a different port" checks the trailing `off`).

**5. External P1s: RESOLVED.** Teardown: `runServeOff` is one attempt + one retry, success-only forget, warn with the exact `tailscale serve --https=<port> off` command on persistent failure, and `disableTailscaleServe` leaves the port registered so exit-time cleanup retries; all test-pinned including the succeed-on-retry-silently case. Status parsing: `checkServeStatusPort` returns `free|conflict|malformed`, inspects `Foreground.<sid>.TCP`, and fails closed on unrecognized shapes. On the false-positive question: `null`, `{}`, and `TCP: null` all classify as `free` (test-pinned), so legitimate empty configs cannot trip the malformed refusal; only genuinely unparsable or wrong-shaped output does (empty stdout, non-JSON, arrays, non-object TCP). The refusal path is clean: no `--bg`, no `off` issued (asserted).

**6. P2s: RESOLVED.** `resetTailscaleServeForTests` clears ports, restores the real runner, and `removeListener`s the exit hook, wired via `afterEach`. `resolveAutoHostCached` is deduped into `packages/shared/tailscale.ts` with an injectable `detect` seam, consumed identically by `remote.ts` and the Pi mirror; vendored output typechecks under `apps/pi-extension/tsconfig.json`. `--tailscale` appears in top-level and per-subcommand help (test updated), and unsupported subcommands reject with a clear error and exit `1` (verified live: `plannotator archive --tailscale`).

### New changes in this commit

- **QR only when the advertised host is overridden** (`shared-handlers.ts:250-262`): remote users with a real `urlHost` keep the QR; port-forward users lose only a QR of `localhost`, which scanned to nowhere. The `--tailscale` path prints its own QR of the serve URL, unaffected. AGENTS.md updated. No side effect on plain remote users beyond dropping the useless QR.
- **`PLANNOTATOR_URL_HOST=""` force under `--tailscale`** (`index.ts:243-248`): runs only inside the `tailscaleFlag` branch, so remote-mode users are untouched. It uses the documented empty-but-set semantics to suppress a config-file `urlHost`, which kills the misleading "set PLANNOTATOR_REMOTE=1" warning I flagged (my CONSIDER 8). Child processes inherit the suppression, which is consistent: spawned agent jobs already pin `127.0.0.1` API URLs.
- Minor residue: the injectable `detect` seam on `resolveAutoHostCached` has no test exercising it yet (my CONSIDER 9 is only structurally addressed). Not merge-relevant.

### Verdict

All BLOCKER and SHOULD FIX items resolved and empirically confirmed; the external P1/P2 fixes are sound and test-pinned; the two new behavior changes are safe. No new findings that would hold this up. **Merge-ready from the security review's perspective.**

PR #1292, opened 2026-08-13
Default new Code Review users to Tree

backnotprop:
## Summary

- select Tree first in the initial Review setup dialog
- initialize genuinely new reviewers to Tree while keeping All changes as the coupled diff default
- preserve returning reviewers’ persisted and last-used panel choices
- add focused regression coverage for new and returning users

## Verification

- bun test packages/review-editor/utils/reviewSetup.test.ts packages/ui/config/reviewPanelViewLastUsed.test.ts packages/review-editor/utils/resolvePanelView.test.ts
- bun run typecheck
- bun run build:review && bun run build:hook && bun run build:opencode


backnotprop (commented):
## TLDR

**Approve with nits, merge after CI is green.** No blockers. The discrimination gate is the right one (`plannotator-review-setup-seen`, unchanged from v0.22.0), the seed goes through the coupled setter correctly, and it is strictly *less* destructive than what is on main: the old first-run reset called `setReviewPanelView('sections')`, which force-wrote `diffOptions.defaultDiffType = 'since-base'` into `~/.plannotator/config.json` over whatever the user had configured. The new seed writes Tree only, which is cookie-only, so a first run now performs **zero** server config writes. That is the real "preserve returning reviewers" win, and it is the one thing the tests do not cover.

Two things to fix before or right after merge: the user-facing docs still say Git status is the default, and the seed's last-used write is both unnecessary for the cohort it claims to serve and the only place a returning reviewer can still be silently flipped.

---

## Cohort matrix

`needsReviewSetup()` reads `plannotator-review-setup-seen` (`packages/review-editor/utils/reviewSetup.ts:8-12`), a per-cookie-jar key that has existed since #990 / v0.22.0, i.e. since the sections view itself. So "genuinely new" means "this browser jar has never completed a plain-git review with a resolvable base". Tracing every cohort:

| Cohort | `seen` | Outcome under this PR | vs `main` |
|---|---|---|---|
| Truly new, no cookies | absent | Seeded `reviewPanelView=tree` + `lastUsed=tree`; `defaultDiffType` untouched, resolves to the `since-base` default (`packages/ui/config/settings.ts:257`). Dialog opens with Tree selected. | Intended change |
| Pre-v0.22 user, never saw the dialog, has `diffOptions.defaultDiffType: "uncommitted"` in config.json | absent | Seeded Tree; their diff default is **preserved** (tree + uncommitted is a valid coupled pair) | Strictly better. `main` force-wrote `since-base` into config.json here |
| v0.22+ user who saw the dialog and kept Git status | present | Untouched (`reviewSetup.ts:27` early return) | Unchanged |
| v0.22+ user who chose Tree | present | Untouched | Unchanged |
| Memo present, persisted view cookie cleared | present | Untouched. Note all three keys live in one jar, so "cleared one cookie" is not a realistic split | Unchanged |
| Cookie jar lost to host scoping (`PLANNOTATOR_URL_HOST`, tailnet hostname, VS Code webview origin) | absent | Seeded Tree in the new jar; `lastUsed` overwritten to `tree`. `defaultDiffType` still arrives from config.json via `configStore.init()` at `App.tsx:1609`, which runs **before** the seed | Better than `main`, which flipped them to sections *and* rewrote config.json |
| Unseen but holding a `lastUsed` memo (v0.22-era user who kept closing the tab without clicking "Got it", back when `seen` was marked only on dismiss, and who had toggled to Git status) | absent | Memo overwritten to `tree`, so they are silently flipped | Same flip on `main`, opposite direction |

**Answer: nobody who ever completed the dialog is flipped.** The only silent flips are in jars where `seen` is genuinely absent, and in every one of those the dialog re-opens, so the flip is visible and recoverable. See the CONSIDER below for the one-line change that removes even those.

---

## VERIFIED-OK

**Coupling integrity is intact and the seed is write-minimal.** `setReviewPanelView('tree', ...)` cannot trigger the sections snap: the `defaultDiffType` write is gated on `view === 'sections'` (`packages/ui/config/reviewView.ts:40-42`). Both `reviewPanelView` and `reviewPanelViewLastUsed` are declared `serverKey: undefined` (`packages/ui/config/settings.ts:205`, `:228`), so `ConfigStore.set()` skips the `pendingServerWrites` queue entirely (`packages/ui/config/configStore.ts:226-229`). Net effect: a first run now issues **no** `POST /api/config`, where `main` issued one that stamped `since-base` over any pre-existing user config. Worth calling out in the release notes, it is the most user-visible part of this change.

**Ordering is correct.** `configStore.init(data.serverConfig)` at `App.tsx:1609` applies config.json before `initializeReviewSetup()` at `App.tsx:1686`, so the seed can never race the server value. The `&&` chain still short-circuits on all five guards before the side effect fires, so the seed cannot mark `seen` on a workspace / PR / jj / p4 / no-resolvable-base session. The callback has `[]` deps and runs once per load, so the side-effecting predicate cannot double-fire in production.

**Self-heal interaction is sane.** The heal is keyed on `persistedPanelView !== 'sections'` returning early (`App.tsx:2483`), so a seeded new user (persisted `tree`) never enters it, and `reviewSetupIsFirstRun.current` still guards the same window (`App.tsx:2482`). Tree + any diff is a valid pair by construction, so there is nothing for the heal to repair on the new default.

**Dialog preselection is driven by config, not array order.** `selected = panelView === opt.key` (`ReviewSetupDialog.tsx:100`) reads `useConfigValue('reviewPanelView')`, so reordering `VIEW_OPTIONS` is purely visual and the seeded `tree` is what lights the "Selected" pill. Choosing Git status still routes through `chooseView` to `setReviewPanelView('sections')` (`ReviewSetupDialog.tsx:72`), which forces `since-base`, syncs the memo, and flips the live panel immediately because `panelView` is derived from reactive config (`App.tsx:396`). The dismiss handler then applies the diff (`App.tsx:4323-4326`). Round trip is correct in both directions.

**The AGENTS.md toggle-order edit is a genuine correction, not a new claim.** `PanelViewToggle.tsx:36-38` renders `tree`, then `sections`, then `commits`. The old doc text (`Git status | Tree | Commits`) was stale, presumably since #1273 reworked the toggle. Good catch.

**Composition with the rest of the cycle holds.** #1273's memo layering is respected (the seed syncs it, so nothing is shadowed). #1278 / #1288's `resolveCommitExitDiff` reads `configStore.get('defaultDiffType')` (`App.tsx:2389`, `:2459`), which this PR never writes, so the Commits detour and the reload un-trap are unaffected. The `commit:<sha>` restore path now runs mostly from Tree rather than sections, which is the cheaper of the two exits.

---

## SHOULD FIX

**1. User-facing docs still say Git status is the default.** `apps/marketing/src/content/docs/commands/code-review.md:150` reads "**Git status** (default)". That is now false for every new reviewer, which is exactly the audience that reads the docs page. Same file line 122 still carries the old `Git status | Tree | Commits` toggle order that this PR just corrected in AGENTS.md, so the two are now inconsistent with each other. Line 84 ("the first time you open a review, a setup dialog lets you choose") is still accurate and needs no change.

**2. The comment on the memo write states an invariant that is false for the cohort it names.** `reviewSetup.ts:29-30`: "The shared setter also records Tree as last-used, so accepting the dialog opens this first review in Tree." For a genuinely new reviewer the memo is `null`, so resolution `session ?? lastUsed ?? persisted` (`App.tsx:396`) already lands on the persisted `tree` without any memo write. The memo write is load-bearing only for the *unseen-but-has-a-memo* cohort, which is the one cohort we are arguably not entitled to overwrite. In a file whose entire job is the coupled-pair invariants, a comment that justifies a write with the wrong reason is worth fixing even if the code stays.

**3. AGENTS.md overstates the seed.** Both `AGENTS.md:222` ("initializes genuinely new reviewers to Tree + since-base once") and `App.tsx:1669-1670` ("initialized to Tree + All changes") claim a `since-base` write that does not happen. `since-base` is inherited, not written, and for the pre-v0.22-with-config.json cohort it will be something else. Suggest "initializes genuinely new reviewers to Tree, leaving the diff default alone (`since-base` unless the user configured otherwise)". The distinction matters because *not writing it* is the whole preservation guarantee.

---

## CONSIDER

**4. `{ recordLastUsed: false }` would make the preservation guarantee total.** Passing it at `reviewSetup.ts:31` gives truly-new reviewers exactly the same Tree opening (persisted `tree`, memo `null`), while leaving an unseen returning reviewer's real memo intact so their session opens on what they were actually using. It also keeps the memo's meaning honest: today, after this PR, a `reviewPanelViewLastUsed` cookie no longer implies "the user touched the toggle", which quietly removes the only signal that distinguishes a reviewer with an opinion from one without. It is the same "this is a non-choice, not a user action" reasoning the self-heal already uses (`App.tsx:2489`). One wrinkle to weigh: the dialog would then show Tree selected while the panel behind it renders Git status for that cohort. If that mismatch is unacceptable, the alternative precise gate is to treat a non-null memo as evidence of a returning reviewer and mark `seen` without seeding. Either way, note that presence of `reviewPanelView` itself cannot be used as the signal, because `ensureLoaded()` seeds every missing default into the cookie jar (`configStore.ts:109-111`), so that key exists for everyone.

**5. The tests cover the gate but not the guarantee.** Per the repo's testing rules, name the failure each test catches:
- `reviewSetup.test.ts:34` asserts `defaultDiffType === 'since-base'` on an empty backend. That is asserting the declared default (`settings.ts:257`), not any behavior of `initializeReviewSetup`. It passes even if the function is emptied.
- `reviewSetup.test.ts:41` fixtures the persisted view as `'tree'`, the same value the seed would write, so the primary assertion at `:48` is a no-op under the regression it is meant to catch. Only `:47` and `:49` actually fail if the gate is removed. Fixture it as `'sections'` and the test becomes load-bearing on all three assertions.
- The missing test is the one the PR title promises: **unseen reviewer with a non-`since-base` `defaultDiffType` keeps it**. Seed the backend with `'plannotator-default-diff-type': 'uncommitted'` and no `seen` key, call `initializeReviewSetup(store)`, assert `defaultDiffType` is still `'uncommitted'`. That single test is what catches a future edit flipping the seed back to `setReviewPanelView('sections')` and silently rewriting config.json again. It is also worth asserting the store queued no server write (the existing `store.setServerSync(() => {})` at `:20` already gives you the seam, just capture the payloads instead of dropping them).

**6. Dialog copy now recommends the option badged "Classic".** `ReviewSetupDialog.tsx:85` recommends Tree, while the same card still carries `tag: 'Classic'` and the Git status card keeps `tag: 'New'` (`:47-59`), under a header that opens with "A simpler review, closer to what you'd see on GitHub". The sentence is not untruthful, but recommending the classic option while badging the other as new reads as a mixed signal on the very screen meant to set a default. Either drop the badges or reword the intro.

**7. Option order is now inconsistent across surfaces.** The dialog lists Tree then Git status (`ReviewSetupDialog.tsx:45-60`), matching `PanelViewToggle.tsx:36-38`. Settings still lists Git status then Tree (`packages/ui/components/Settings.tsx:378-381`). Cheap to align while you are here.

**8. Side-effecting predicate in a boolean chain.** `initializeReviewSetup()` at `App.tsx:1686` both mutates and answers "should I show the dialog". It is safe here (single-shot callback, idempotent on the second call), but the name reads like a command and the return value reads like a query. `const isFirstRun = initializeReviewSetup();` on its own line above the `if` would keep the same behavior and make the mutation visible at the call site.

---

## CI

One red leg: `test` in the **Release** workflow (run 31662595296) fails on `bun agent terminal bridge > node sidecar rebuilds spawn options from the server-owned launch plan`. That test is in `packages/server/agent-terminal.test.ts`, touches nothing in this PR, passes locally on `main` (5 pass / 0 fail), and the identically-named `test` job in the **Test** workflow passed for the same commit. Almost certainly an environment flake in that leg. Re-run before merging rather than merging red. Everything else (gitleaks, zizmor, install.ps1, install.cmd) is green.

Local note: I could not reproduce the new test run in a detached worktree because the symlinked isolated-linker `node_modules` splits `@plannotator/ui/utils/storage` and the relative `../utils/storage` into two module instances, so `setStorageBackend` only reaches one of them. Not a defect in the test, but a reminder that `reviewSetup.test.ts:2` imports through the package alias while the sibling `packages/ui/config/reviewPanelViewLastUsed.test.ts:2` imports relatively. Matching the sibling's relative style would make the test robust to that resolution difference.

---

backnotprop:
Addressed the review nits in 0a6ff551:

- returning-user coverage now persists Sections, the opposite of the Tree seed, while preserving a distinct Tree last-used memo
- added an unseen-reviewer case proving an existing Uncommitted diff default survives initialization
- corrected AGENTS.md and code comments to say Tree is cookie-only and the resolved diff default is inherited, not written

The stale Git status (default) claim exists only in the deprecated apps/marketing docs in this checkout; there is no Mintlify source here, so I left it for the current docs follow-up rather than widening this PR.

backnotprop (commented):
## Delta review (0a6ff551)

TLDR: All three fixes verified. Merge-ready once the unrelated Windows smoke-binaries leg is re-run.

**RESOLVED: returning-user fixture is now a real guard.** The fixture persists `reviewPanelView=sections`, a value that differs from the seed's `tree`. With `seen=true` the function early-returns, so the persisted `sections` survives and the `toBe('sections')` assertion holds. Drop the `needsReviewSetup()` guard and the seed would overwrite it to `tree`, failing the assertion. No longer a no-op.

**RESOLVED: uncommitted-survives test exercises the unseen init path and guards the config-stamping revert.** Backend seeded with `defaultDiffType=uncommitted` and no `seen` key; `initializeReviewSetup(store)` returns `true` (unseen path taken), then asserts `defaultDiffType` is still `uncommitted`. A revert to `setReviewPanelView('sections')` would force `since-base` through the coupled setter and fail that assertion. The observable "diff default unchanged" is the correct proxy for "no server config write" here, since `reviewPanelView` is `serverKey: undefined` and only the sections coupling would have queued one.

**RESOLVED: wording is accurate in both places.** AGENTS.md now says the initializer "seeds the cookie-only Tree choice ... inherits the resolved `defaultDiffType` without a server config write" and "initializes an unseen reviewer's panel to Tree once while preserving the resolved diff default". The `reviewSetup.ts` comment matches: "Selecting Tree preserves whichever defaultDiffType the store resolved ... without writing server-backed config." Both are now true to the code.

**NOTED: marketing doc** (`code-review.md:150` "Git status (default)", `:122` stale toggle order) deferred to the Mintlify follow-up, as stated. Fine to land separately.

CI: both `test` jobs pass (Test and Release workflows), typecheck passes, gitleaks / zizmor / install legs green. One red leg, `smoke-binaries (windows-latest)`, fails inside the agent-terminal WebTUI runtime smoke ("install-runtime agent-terminal failed"). It touches nothing in this PR (no server or runtime changes) and is a known Windows infra flake shape. Re-run that leg rather than merge red.

Verdict: approve. All review items addressed; the two new tests are load-bearing.

PR #1294, opened 2026-08-13
ci(security): add Semgrep CE and Trivy monitoring

backnotprop:
## Outcome

Adds Semgrep Community Edition source SAST and standalone Trivy dependency/filesystem/IaC scanning to the existing separate Security Scanning workflow. Both jobs run on the workflow's existing pull request, `main` push, weekly schedule, and manual-dispatch triggers.

This is monitor mode: findings are reported, summarized, retained, and uploaded as SARIF, but findings do not fail the jobs. Scanner installation, checksum/digest, rules, parser, timeout, database freshness, checks collection, dependency catalog, detector health, or report validation failures do fail. Release and deploy workflows are unchanged.

No Semgrep/Aqua account, GitHub App, third-party token, or service account is required. Semgrep receives no source or findings. Trivy secret scanning is disabled because Gitleaks remains authoritative.

## Semgrep CE

- Covers tracked TypeScript/JavaScript across the Bun and Pi/Node servers, React/DOM UI, AI/provider code, Cloudflare services, and other application sources. Generic rules cover shell, PowerShell, batch, and Dockerfile installer patterns without relying on Semgrep's experimental Bash parser.
- Uses Semgrep `1.172.0` from the official non-root image pinned to `sha256:d1012a3bf2acf47721216fbf7ff12d4c2971cc7f9c7b77cf6c6e9dcf006bd487`; the job verifies the digest, non-root user, version, CE request, OSS engine identity, and SARIF driver identity.
- Uses 15 Apache-licensed, repository-owned rules in `.semgrep.yml`. The scan is offline (`--network none`, read-only source mount, `--oss-only`, metrics/version checks off) and does not use `semgrep ci`, the Registry, `SEMGREP_APP_TOKEN`, or Semgrep Cloud.
- The official Semgrep rules repository was reviewed at `40b8c63f75dc7c22c8a77482d73bfb864b146f7e`, but its rules were not copied because the Semgrep Rules License restricts redistribution. Repository-owned rules avoid a mutable network ruleset and keep rule review/versioning local.
- CE limitations are explicit: intrafile analysis only, with no Pro cross-file/cross-function dataflow, reachability, hosted policy, or hosted fingerprinting.
- Runtime-created fixtures prove Node `exec`, Bun shell execution, and remote-script pipeline detections at exact lines. A deliberately malformed runtime ruleset must fail.
- Six behavior-preserving syntax changes (type-only imports, an equivalent `RegExp` constructor, and the JSX `&amp;` entity) allow strict parsing of active TS/TSX source. The large installer remains active coverage through generic rules rather than being excluded.
- Exclusions are limited to ignored/generated outputs and the Pi vendor mirror. Canonical sources remain covered; tracked `build/shiki-wasm-stub.ts` is not broadly excluded.

Semgrep release: https://github.com/semgrep/semgrep/releases/tag/v1.172.0

### Initial Semgrep baseline

Initial strict local scan: 1,082 targets, 15 rules, 28 findings (21 `ERROR`, 7 `WARNING`), zero parser/config/timeout errors, approximately 100% parsed lines. The final trusted CI checkout reports 30 findings (23 `ERROR`, 7 `WARNING`) across the same 1,082 targets with zero engine errors; that retained CI report is the current baseline. Both server implementations, UI, Cloudflare, and installers were asserted in `paths.scanned`.

Every current result was reviewed and grouped without suppressions:

- 17 request-to-filesystem audit hits in the final CI baseline: one developer mock, one randomized/extension-validated upload path, and 15 Bun/Pi reference/image/document paths with containment/validation preceding the reported sink. Retained for follow-up validation review rather than declared exploitable.
- 2 request-to-network audit hits: the OpenCode installed-smoke registry proxy fixture and VS Code's loopback cookie proxy. Both have constrained intended upstreams; retain for explicit allowlist review.
- 1 Pi `execSync` hit: current callers pass fixed Git subcommands, but the string-command helper should be converted to argv execution in a focused follow-up.
- 7 DOM/React HTML sink hits: two pass through the repository HTML sanitizer; the others are syntax highlighting, KaTeX (`trust: false`), Mermaid (`securityLevel: strict`), and Viz-generated SVG render boundaries. Retain the generated-SVG boundaries for dedicated XSS review.
- 3 shell hits are inert `curl | bash` examples inside installer help text, not executed commands.

No application finding is suppressed or fixed in this scanner PR.

## Trivy

- Uses standalone Trivy `0.73.0`, not `trivy-action` or `setup-trivy`. The Linux x64 archive is pinned to SHA-256 `2edd39da482bb4e9831962487b68f68e3928ec3137794757f54d00383d79547b` and the exact version is asserted.
- Release source/tag commit is `40c73e5d6166dcc0346a1ab4e94499d1572854e4`. The exact Linux archive was independently verified before this PR with GitHub attestation verification against Aqua's reusable release workflow, tag, source digest, and GitHub-hosted runners.
- The March 2026 incident affected malicious Trivy `0.69.4`, Docker `0.69.5`/`0.69.6`, and replaced mutable `trivy-action`/`setup-trivy` tags. Direct v0.73.0 installation avoids those Action paths and is also newer than subsequent scanner fixes in 0.71.x/0.72.x.
- The official `ghcr.io/aquasecurity/trivy-db:2` tag is resolved anonymously to an immutable digest per run, then downloaded by digest. Schema 2, nonempty DB, no more than 48-hour age, and no more than five-minute future skew are enforced. The digest and update time appear in the summary.
- Official Trivy checks v2.2.0 are pinned to OCI digest `sha256:891abb1e1dc95429e6ad6768a8c8164dacf6d7ff742b940d1d709e69b5855cc6`; cached policy metadata must match it so a failed download cannot silently fall back to embedded checks.
- Version checks and telemetry are disabled. Intended scanner network is limited to public GitHub release download and public GHCR DB/checks retrieval.
- Vulnerability/misconfiguration scanning covers the committed text `bun.lock`, 20 optional CallDiff npm lockfiles, and filesystem/IaC configuration. Generated/install outputs are excluded. `tests/manual/ssh/Dockerfile` remains scanned and is labeled as a non-production manual fixture.
- Development dependencies are included. A separate informational license scan runs after `bun install --frozen-lockfile --ignore-scripts`; it describes the runner/install graph and is not claimed to be a deployment inventory.
- Runtime-only `lodash@4.17.20` lockfile and Dockerfile fixtures must produce the configured finding exit, CVE-2021-23337, and DS-0002 at line 2. A deliberately invalid report exercises the report failure guard.

Incident/advisory and release evidence:

- https://github.com/aquasecurity/trivy/security/advisories/GHSA-69fq-xp46-6x23
- https://github.com/aquasecurity/trivy/discussions/10462
- https://github.com/aquasecurity/trivy/releases/tag/v0.73.0
- https://github.com/aquasecurity/trivy/security/advisories/GHSA-mcj4-mphf-j9ff
- https://github.com/aquasecurity/trivy/security/advisories/GHSA-q3fv-x8vg-qqm4

### Bun catalog evidence

- Clean checkout: text `bun.lock` parsed directly as 1,368 packages; 60 vulnerabilities.
- After frozen, script-free install: the same 1,368 packages and 60 vulnerabilities. Install materialization does not create the dependency result.
- With the 20 CallDiff npm locks: 1,470 catalog records.
- Sentinels: direct `dompurify@3.4.8`, indirect `hono@4.12.25`, and direct development `bun-types@1.3.14`.
- Without `--include-dev-deps`, Bun coverage falls to 743 packages and 57 vulnerabilities; the enabled scan includes 625 additional dev/test-classified packages.
- Installed-license evidence grows from 102 lockfile records before install to 1,095 in the initial local run (993 associated with `bun.lock`) and 1,102 in the final Linux CI run (1,000 associated with `bun.lock`). The small platform/install-graph variance reinforces that this is informational runner evidence, not a deployment inventory.

### Initial Trivy baseline and high/critical disposition

Vulnerabilities: 60 total — 0 critical, 22 high, 29 medium, 9 low. All 22 high findings currently have fixes in the DB and are left for a focused dependency follow-up:

- Direct marketing build dependency `astro@5.18.2`: CVE-2026-50146, CVE-2026-54299.
- Potential runtime paths: `brace-expansion@5.0.6` (CVE-2026-13149, CVE-2026-14257, CVE-2026-69152), `fast-uri@3.1.2` (CVE-2026-13676, CVE-2026-16221, CVE-2026-18446), `ip-address@10.2.0` (CVE-2026-69192), and `undici@8.3.0` (CVE-2026-12151, CVE-2026-13697, CVE-2026-9675, CVE-2026-9697).
- Dev/packaging-only evidence: `form-data@4.0.5` CVE-2026-12143 and `linkify-it@5.0.1` CVE-2026-59887.
- Astro/Vite/build/tooling chains: `js-yaml@4.2.0` CVE-2026-59869 and GHSA-5p4m-2wfm-xmqj; `nanoid@3.3.12` CVE-2026-67213 and CVE-2026-67214; `postcss@8.5.15` GHSA-r28c-9q8g-f849; `sharp@0.34.5` GHSA-f88m-g3jw-g9cj; `svgo@4.0.1` GHSA-2p49-hgcm-8545.

Misconfigurations: 4 failures (2 high, 1 medium, 1 low), all confined to the non-production manual SSH Dockerfile. High DS-0002 (root user) and DS-0029 (`apt-get` without `--no-install-recommends`) are intentional fixture characteristics, not production image findings.

License reporting: one Trivy critical/forbidden WTFPL opinion through dev-only VS Code tooling and one high/restricted LGPL-3 opinion through optional Astro/libvips build tooling. These are classifier opinions, not vulnerabilities or proof of shipped components.

## Review hardening

- Failed Semgrep/Trivy validation now retains partial machine reports and named logs under `always()`; incomplete SARIF is an artifact for diagnosis, not published as a complete code-scanning analysis.
- Semgrep rule count is derived from SARIF with a safety floor, coverage uses required repository-surface prefixes rather than five filenames, and every jq guard reports the failed invariant.
- Trivy compares cataloged npm targets with the tracked in-scope lockfile set derived from Git, uses broad meaningful-coverage floors, and asserts direct/transitive/development sentinel names and relationships without pinning their routine-update versions.
- Immutable Semgrep image, Trivy archive/DB, and checks retrieval has bounded retries. The checks reference is digest-only v2.2.0 content and therefore never follows the floating `2` tag at runtime.

## GitHub safety and reports

- Global permission remains `contents: read`. Each SARIF job narrowly adds `actions: read` and `security-events: write`; there is no OIDC, environment, npm, publishing, deployment, or repository-write credential.
- Checkout Actions are full-SHA pinned with `persist-credentials: false`. All added Actions are full-SHA pinned; scanner versions/digests are separately pinned.
- The workflow remains `pull_request`, never `pull_request_target`. Fork and Dependabot PRs execute meaningful scans with no repository secrets; SARIF upload is skipped only where GitHub does not allow the write.
- Semgrep and Trivy report artifacts run under `always()`: complete or partial JSON/SARIF plus scanner, detector, and configuration diagnostics are retained for 7 days even when a scanner or assertion fails. Code-scanning SARIF remains limited to fully validated reports in trusted contexts so an incomplete scan cannot appear authoritative. Scanner caches, databases, source, credentials, and fixture source are not retained.
- Existing Gitleaks and zizmor jobs are unchanged in authority and behavior. Cancellation remains a cancelled workflow, never a clean report.

## Validation

- Final hardened Security Scanning run [31664237033](https://github.com/backnotprop/plannotator/actions/runs/31664237033) against commit `af06b9e0`: Semgrep CE, Trivy, Gitleaks, and zizmor all pass; expanded diagnostic artifacts and both validated SARIF uploads completed successfully.
- Final PR status: 22 successful checks, 4 expected release/publish skips, zero failures, cancellations, or pending checks.
- Exact pinned Semgrep full scan and all detector/config failure assertions: pass.
- Exact pinned Trivy clean/post-install comparisons, full scan, DB/checks freshness/identity, detector exit, catalog and report assertions: pass.
- SARIF 2.1.0, tool/version/rules/results metadata assertions: pass.
- `bun run typecheck`: pass.
- Focused affected tests: 122 pass, 10 DOM-environment skips, 0 fail.
- `actionlint` 1.7.12 against all workflows: pass.
- zizmor 1.29.0 offline regular strict collection: 0 findings.
- `git diff --check` and generated-report/cache/credential audit: pass.
- Full `bun test` reaches one unrelated, reproducible pre-existing Pi GitButler assertion mismatch (`semanticDiff` now contains `enabled: false`); no scanner work touches that behavior, so it is not absorbed here.

## Promotion plan and residual risk

After baseline owners and follow-up triage are agreed, a separate approval should promote only new high-confidence Semgrep high/critical findings and new fixable critical Trivy vulnerabilities to blocking. This PR does not change publishing/deployment gates.

Residual decisions: assign owners for the Semgrep audit groups and dependency chains; decide whether generated SVG render boundaries need additional sanitization; and retain or replace the current Aqua publisher trust for DB/check OCI artifacts if Aqua publishes verifiable attestations/signatures. Scanner version/checksum/image digest and repository rules are deliberately manual review updates; the vulnerability DB intentionally changes per run but is digest-pinned and freshness-checked within that run.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -23,9 +23,9 @@ env:
   GITLEAKS_LINUX_X64_SHA256: 79a3ab579b53f71efd634f3aaf7e04a0fa0cf206b7ed434638d1547a2470a66e
   SEMGREP_VERSION: 1.172.0
   SEMGREP_IMAGE: semgrep/semgrep:1.172.0-nonroot@sha256:d1012a3bf2acf47721216fbf7ff12d4c2971cc7f9c7b77cf6c6e9dcf006bd487
-  SEMGREP_RULE_COUNT: "15"
   TRIVY_VERSION: 0.73.0
   TRIVY_LINUX_X64_SHA256: 2edd39da482bb4e9831962487b68f68e3928ec3137794757f54d00383d79547b
+  # v2.2.0 content; the digest-only runtime reference never follows a moving tag.
   TRIVY_CHECKS_REFERENCE: ghcr.io/aquasecurity/trivy-checks@sha256:891abb1e1dc95429e6ad6768a8c8164dacf6d7ff742b940d1d709e69b5855cc6
   TRIVY_DB_MAX_AGE_SECONDS: "172800"
   TRIVY_DISABLE_TELEMETRY: "true"
@@ -183,7 +183,17 @@ jobs:
         shell: bash
         run: |
           set -euo pipefail
-          docker pull "$SEMGREP_IMAGE"
+          for attempt in 1 2 3; do
+            if docker pull "$SEMGREP_IMAGE"; then
+              break
+            fi
+            if (( attempt == 3 )); then
+              echo 'Semgrep image pull failed after three attempts.' >&2
+              exit 1
+            fi
+            echo "Semgrep image pull failed; retrying (attempt $((attempt + 1))/3)." >&2
+            sleep $((attempt * 5))
+          done
 
           repo_digest=$(docker image inspect --format '{{index .RepoDigests 0}}' "$SEMGREP_IMAGE")
           expected_digest=${SEMGREP_IMAGE##*@}
@@ -209,7 +219,7 @@ jobs:
         run: |
           set -euo pipefail
           fixture_dir="$RUNNER_TEMP/semgrep-health"
-          output_dir="$RUNNER_TEMP/semgrep-health-output"
+          output_dir="$GITHUB_WORKSPACE/security-results"
           mkdir -p "$fixture_dir" "$output_dir"
           chmod 777 "$output_dir"
 
@@ -229,6 +239,7 @@ jobs:
           curl --fail https://example.invalid/install.sh | sh
           EOF
 
+          set +e
           docker run --rm --network none \
             --env SEMGREP_SEND_METRICS=off \
             --volume "$GITHUB_WORKSPACE:/src:ro" \
@@ -244,9 +255,15 @@ jobs:
               --strict \
               --timeout 30 \
               --timeout-threshold 1 \
-              --json-output /out/health.json \
+              --json-output /out/semgrep-health.json \
               --config /src/.semgrep.yml \
-              /fixture
+              /fixture 2>&1 | tee "$output_dir/semgrep-health.log"
+          health_status=${PIPESTATUS[0]}
+          set -e
+          if [[ "$health_status" -ne 0 ]]; then
+            echo "Semgrep controlled detector scan failed (exit $health_status)." >&2
+            exit "$health_status"
+          fi
 
           jq -e --arg version "$SEMGREP_VERSION" '
             .version == $version and
@@ -262,7 +279,10 @@ jobs:
             any(.results[];
               .check_id == "plannotator.installer.remote-script-pipeline" and
               (.path | endswith("remote-pipeline.sh")) and .start.line == 2)
-          ' "$output_dir/health.json" >/dev/null
+          ' "$output_dir/semgrep-health.json" >/dev/null || {
+            echo 'Semgrep controlled detector report failed its engine, error, or exact-line assertions.' >&2
+            exit 1
+          }
 
           # A malformed local ruleset must be an engine failure, not a clean scan.
           cat > "$fixture_dir/invalid.yml" <<'EOF'
@@ -279,7 +299,7 @@ jobs:
             "$SEMGREP_IMAGE" \
             semgrep scan --oss-only --metrics off --disable-version-check \
               --strict --config /fixture/invalid.yml /fixture/command-injection.ts \
-              >/dev/null 2>&1
+              >"$output_dir/semgrep-invalid-rules.log" 2>&1
           invalid_status=$?
           set -e
           if [[ "$invalid_status" -eq 0 ]]; then
@@ -292,12 +312,13 @@ jobs:
         shell: bash
         run: |
           set -euo pipefail
-          output_dir="$RUNNER_TEMP/semgrep-output"
-          mkdir -p "$output_dir" security-results
+          output_dir="$GITHUB_WORKSPACE/security-results"
+          mkdir -p "$output_dir"
           chmod 777 "$output_dir"
 
           # The source mount and scanner network are read-only/offline. Findings
           # remain local; only GitHub receives the validated reports below.
+          set +e
           docker run --rm --network none \
             --env SEMGREP_SEND_METRICS=off \
             --volume "$GITHUB_WORKSPACE:/src:ro" \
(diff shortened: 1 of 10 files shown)


backnotprop (commented):
## Review: Semgrep CE + Trivy monitoring

**TLDR: Mergeable at `7b16d84e`.** This holds the #1274 bar and in places exceeds it. I independently re-verified every checksum, image digest, and action SHA (downloaded and hashed the Trivy archive, resolved the Semgrep and trivy-checks digests against the registries, dereferenced all four action tags). Monitor-only enforcement, per-job least privilege, and fork safety all check out. The Trivy job was red on its own PR twice before the two fix commits; I reproduced the entire pipeline locally and in a Linux container to chase it and can confirm the `7b16d84e` explanation is correct. Remaining items are maintenance-fragility follow-ups, not security gaps.

### BLOCKER

None remaining. At `73e1871a` the Trivy job failed on its own PR with a bare `exit 1` and would have been one: `trivy fs` reports `ArtifactType: "repository"` when the scan root has readable Git metadata and `"filesystem"` otherwise. I confirmed this independently: a faithful reproduction (pinned 0.73.0, same flags, same DB digest, Linux container, `bun install --frozen-lockfile --ignore-scripts`) reports `filesystem` whenever the checkout lacks a real `.git` directory (my worktree had a gitfile), which is why it only failed on the runner. Fixed with an honest comment in `7b16d84e`; run `31663419424` is green across all four jobs (Gitleaks, Semgrep CE 2m48s, Trivy 44s, zizmor).

### SHOULD FIX

1. **Failed runs retain nothing to debug with.** `Retain Semgrep reports` and `Retain Trivy reports` require `steps.scan.outputs.report-ready == 'true'`, which is only set at the very end of the scan step, after every assertion. When an assertion fails, `security-results/` is discarded. This is exactly why the two red runs on this PR were nearly undebuggable from the logs; I had to rebuild the pipeline locally and still could not reproduce until the author's diagnostics landed. Upload the artifacts with a bare `if: always()` and `if-no-files-found: ignore`, or set `report-ready` as soon as the report files exist.
2. **Exact-value assertions will redden Security Scanning on routine, security-irrelevant changes.** The coverage-guard intent is sound (a silent parser loss would otherwise report "clean"), but the implementation over-pins: `dompurify@3.4.8` / `hono@4.12.25` / `bun-types@1.3.14` version sentinels, `npm_target_count != 20`, `SEMGREP_RULE_COUNT` hand-synced against `.semgrep.yml`, and five sentinel file paths in the Semgrep jq (`packages/ui/components/InlineMarkdown.tsx`, `scripts/install.sh`, etc.). Any dependency bump, CallDiff pack addition or removal, rule addition, or file move breaks the workflow. Suggest asserting package names without versions (or reading the expected version out of `bun.lock` at run time), deriving the rule count from the file (`grep -c '^  - id:' .semgrep.yml`), and noting in the failure messages which knob to update. As written, the first `bun.lock` touch after merge turns this job red.
3. **The Semgrep job still fails silently.** `ba7b8ef9` gave the Trivy assertions named diagnostics, but the Semgrep job's two large `jq -e ... >/dev/null` assertions (and the small SARIF checks in the Gitleaks/zizmor jobs) still exit 1 with no message. Apply the same named-guard pattern; this PR's own history shows what debugging a silent guard costs.

### CONSIDER

1. **Retry coverage for the remaining network legs.** The Trivy binary download correctly gets `--retry 3 --retry-all-errors` (better than the existing gitleaks/zizmor installs, which have none). But the Semgrep `docker pull` from Docker Hub has no retry and is exposed to unauthenticated pull rate limits from shared runner IPs, and the ghcr.io token/manifest curls in the DB-resolve step also lack `--retry`. Given the CDN flakes we have been chasing all day, these are the next red-herring candidates.
2. **DB freshness is fail-closed on every PR.** If trivy-db publication stalls past 48 hours or GHCR is down, all PRs go red. The PR body says this is deliberate, and I can live with it, but enforcing the age gate only on `schedule`/`push` and downgrading to a summary warning on `pull_request` would keep upstream outages out of unrelated PRs.
3. **Provenance nuance on the checks pin.** The PR body says "Official Trivy checks v2.2.0 are pinned". The pinned digest `sha256:891abb...` is the current digest of the floating `2` tag; the `2.2.0` tag resolves to a different digest (`sha256:b63166...`). The pinned manifest self-identifies as version 2.2.0 (created 2026-03-17, revision `d7c9302130a9...`), so the pin is sound and immutable, but the next person who tries to re-verify it against the `2.2.0` tag will get a mismatch. Worth a comment next to `TRIVY_CHECKS_REFERENCE`.
4. **Post-merge baseline.** The ~28 Semgrep and 60 Trivy baseline findings will appear as open code scanning alerts after the first main run; PRs only get inline annotations on changed lines, so PR noise should stay low. Assigning the baseline owners promptly (already planned in the PR body) keeps the Security tab meaningful.

### VERIFIED-OK

- **Checksums and digests, re-verified independently.** Trivy `trivy_0.73.0_Linux-64bit.tar.gz`: downloaded and hashed, `2edd39da...` matches. Semgrep `1.172.0-nonroot`: Docker Hub manifest digest `sha256:d1012a3b...` matches the pin (and the job re-verifies digest, non-root user, and version at runtime). trivy-checks `sha256:891abb...`: exists on GHCR with version annotation 2.2.0. DB is resolved to a digest per run, then downloaded by digest, with schema, age, and future-skew gates.
- **Action pins.** `actions/checkout@3d3c42e5` = v7.0.1, `actions/upload-artifact@043fb46d` = v7.0.1, `oven-sh/setup-bun@0c5077e5` = v2.2.0 (same SHA as test/deploy/release), `github/codeql-action@5595ccaf` = v4.37.6 (dereferenced the annotated tag). No floating tags, no `latest` anywhere.
- **Permissions and credential isolation.** Global `contents: read`; each new job adds only `actions: read` + `security-events: write`, matching the gitleaks/zizmor precedent; `persist-credentials: false` on every checkout; still `pull_request`, never `pull_request_target`; the fork/Dependabot SARIF-upload guard is identical to the existing jobs and the scans themselves run for forks with no secrets. Semgrep scans with `--network none` and a read-only source mount.
- **Semgrep ruleset provenance and licensing.** 15 repo-owned rules in `.semgrep.yml` (counted; matches `SEMGREP_RULE_COUNT`), Apache-2.0, no registry config, no `SEMGREP_APP_TOKEN`, metrics off. No server-side ruleset drift is possible; rule changes go through PR review. Declining to vendor registry rules is the correct reading of the Semgrep Rules License.
- **Monitor-only confirmed.** Findings never fail a job (`--exit-code 0` on both scanners' real scans); only install, verification, parser, DB, and report failures do. No release or deploy workflow is touched, and the new job names collide with nothing.
- **Trivy scope.** Lockfile-based (text `bun.lock` parsed directly; I confirmed 1368 packages, 20 npm lock targets, and the dev-deps delta locally) plus IaC misconfig; secrets scanning stays with Gitleaks; telemetry and version checks off.
- **The six source edits are behavior-preserving.** The `RegExp` constructor in `packages/core/extract-code-paths.ts` compiles to the identical source and flags (checked in node); the type-only imports in `serverReview.ts`, `ReviewSidebar.tsx`, and the two test files resolve to the same modules the inline `import()` types referenced; `&amp;` in JSX text decodes to `&`.
- **CI on the PR.** All four security jobs green at `7b16d84e`, plus the standalone gitleaks/zizmor/Semgrep OSS checks.

PR #1295, opened 2026-08-13
Mobile foundation and quieter first run

backnotprop:
## Summary
- establish a shared Visual Viewport and safe-area contract for Plan and Code Review
- replace the 0.20.0 promo with a compact Grid/Clean chooser
- remove Vim, Ask AI, and analysis-layer startup announcements while preserving established defaults and consent
- document the mobile audit, physical iPhone evidence, and phased implementation contract

## Verification
- 25 focused DOM/unit tests passed
- full typecheck passed, including the strict UI consumer
- Plan and Code Review production builds passed
- bounded phone, landscape, and desktop browser QA passed
- physical iPhone stage review completed; iPad remains assigned to the next combined device gate

## Scope
This is the first stack layer. Keyboard/composer work is intentionally reserved for the stacked Checkpoint 1B PR.


backnotprop (commented):
## TLDR

Solid foundation layer: the viewport hook is genuinely well built (rAF-coalesced, ref-counted, pinch-zoom-stable, restore-on-cleanup), the dialog-chain surgery is coherent for every cohort I traced, and Call Flow consent is genuinely preserved. Merges cleanly with #1292 in either order. Three things need attention before merge: the silent AI-provider cookie persistence in `useAIProviderConfig` is a real behavior change that contradicts the "preserving established defaults" claim, AGENTS.md still documents the five-dialog chain this PR deletes, and the `var(--pn-viewport-height, 100dvh)` fallback breaks the exact old browsers the comment claims to protect. All are cheap fixes. Verdict: approve after the SHOULD FIX items (the AGENTS.md one I consider blocking, since that file is the operating manual every agent in this repo runs on).

Scope honesty check: 1,851 additions are 1,022 lines of docs (SPIKE 578 + spec 444) and ~830 lines of code and tests, of which 405 are tests. The code footprint is proportionate to a "foundation" label. CI is green (one Windows smoke pending at review time). I re-ran the 25 focused tests in a worktree (`DOM_TESTS=1`: 25 pass) plus the full repo typecheck (pass) and the merged-tree review-editor util suite (181 pass).

## BLOCKER

**1. AGENTS.md (= CLAUDE.md, symlinked) is left contradicting the code.** `AGENTS.md:223` still says "The versioned one-time dialog chain is guide intro -> look-and-feel -> review setup -> analysis layers -> Edit Mode", and `AGENTS.md:252` still says "Per-origin choices are persisted in cookies, so a user can override the automatic match" (no longer true once resolution itself persists, see SHOULD FIX 1). The analysis-layers consent framing elsewhere in the doc is also stale. This repo runs agents off that file as authority; a stale chain invariant is how the analysis dialog gets reinvented or `analysisIntroVisible` gets re-added to a gate. #1292 edits the same sentence, so whoever lands second must reconcile it anyway. Please update the chain description, the analysis-intro/consent text, and the Ask AI provider-persistence sentence in this PR.

## SHOULD FIX

**1. Silent provider persistence changes precedence semantics and can freeze a transient resolution.** `packages/ui/hooks/useAIProviderConfig.ts:76` now writes `saveAIProviderSelection(...)` with the auto-resolved provider on every capabilities load. Via `applyAIProviderSelection` (`packages/ui/utils/aiProvider.ts:203`) that records the resolution as a per-origin cookie (dedicated-provider origins) or as the global `providerId` (amp/droid/gemini/copilot origins). Consequences:
   - The resolver is deterministic per session, so persistence buys nothing for the stated goal ("first use does not need an announcement dialog"). Removing the dialog required removing the dialog, not adding writes.
   - If the origin's dedicated provider is not detected in one session (transient detection failure, unauthenticated CLI), the fallback pick is persisted as the per-origin choice and then permanently outranks the origin match once the real provider appears. Today that failure is transient; after this PR it is sticky.
   - For origins without a dedicated provider, the first session freezes `serverDefaultProvider`/`providers[0]` into the global cookie, so later server-default changes stop taking effect.
   Recommend dropping the save entirely (resolution already produces the same answer every session), or at minimum only saving when the selection came from the per-origin/origin-match tiers. The new test at `packages/ui/aiProviderConfigPersistence.test.tsx:101` pins the current behavior, so this is clearly intentional; I think the intent is wrong, and it is also the thing that makes the PR's "preserving established defaults" claim not quite true.

**2. The `var()` fallback defeats the old-browser ladder it sits on.** `packages/ui/theme.css:718-722` declares `height: 100vh; height: 100dvh; height: var(--pn-viewport-height, 100dvh);`. On a browser without `dvh` support, declaration 2 is dropped at parse time, but declaration 3 parses fine and wins the cascade; at computed-value time the substituted `100dvh` fallback is invalid, which makes the property invalid-at-computed-value-time, i.e. `height: unset` (auto). The earlier `100vh` never applies. So the exact browsers the comment says are protected get a collapsed root until the observer's first `useEffect` write (and forever, if JS fails). Fix: make the `var()` fallback `100vh`, or wrap the `var()` declaration in `@supports (height: 100dvh)`. Same pattern exists in `.pn-viewport-bounded` and the dialog's inline `maxHeight` (`packages/ui/components/LookAndFeelAnnouncementDialog.tsx:124`).

**3. Six orphaned files plus one asset are left fully dead on the branch.** `packages/ui/components/PlanAIAnnouncementDialog.tsx`, `packages/ui/components/VimModeAnnouncementDialog.tsx`, `packages/review-editor/components/ReviewAnalysisIntroDialog.tsx`, `packages/ui/utils/planAIAnnouncement.ts`, `packages/ui/utils/vimModeAnnouncement.ts`, `packages/review-editor/utils/reviewAnalysisIntro.ts`, and `packages/ui/assets/workspaces.webp` (still imported only by `ReviewSetupDialog`, so the webp survives; the other six have zero references, verified by grep). Tree-shaking keeps them out of the bundles, but a PR titled around removing these announcements should remove their implementations, or say why they are being kept.

## CONSIDER

- `packages/review-editor/App.tsx:1377` now passes a literal `true` for `consentAcknowledged` to `useCallFlowAutoInstall`. If the pending-intro gate is gone for good, fold the parameter away (and its `if (!enabled || !consentAcknowledged)` early-out) rather than shipping a permanently-true argument.
- `apps/portal/index.html` (share portal, renders the same `@plannotator/editor` App and therefore runs `useViewportEnvironment` and `.pn-app-viewport`) keeps the old viewport meta without `viewport-fit=cover`. That is safe (safe-area vars compute to 0) but inconsistent; either add it or note the exclusion in the spec, which currently lists only hook and review as entry documents.
- `.pn-safe-top`, `.pn-safe-inline`, `.pn-safe-bottom`, `.pn-viewport-bounded` (`packages/ui/theme.css`) have zero consumers in this layer; the dialog re-derives the same math inline with `max(0.75rem, ...)` instead of using them. Spec-sanctioned as contract-for-1B/1C, but worth a comment marking them as such so they do not read as dead CSS.
- `packages/ui/theme.css:748`: the `.plan-look-choice-option` hover rule is component-specific CSS in the shared token file. Minor placement gripe.
- Escape on the chooser resolves the one-time choice with whatever is displayed (`LookAndFeelAnnouncementDialog` Escape -> `onDismiss` -> persist + mark resolved). Defensible (the default is sane and Settings remains), but it does mean an accidental Escape permanently consumes the first-run moment.
- The dialog hand-rolls portal/focus-trap/Escape while the repo has a Base UI dialog primitive. It is a strict a11y improvement over the old hand-rolled version, so fine for now; 1C's "dialog bounds" checkpoint is the natural moment to converge it.

## VERIFIED-OK

- **Dialog chain coherence, all cohorts.** New user: guide intro -> Grid/Clean chooser -> review setup -> edit mode, never stacked (`editModeAnnouncement.ts:56` gate updated consistently; `DestinationSpotlight` gate at `App.tsx:4325` updated too). Existing user with legacy `plannotator-look-feel-announcement-seen=2`: migrated, no re-prompt (`lookAndFeelAnnouncement.ts:12-18`, covered by test). Legacy `1` (release-only announcement, never made the grid choice): re-asked, deliberate and correct. Removed announcements' seen-cookies linger unread, harmless.
- **Consent genuinely preserved.** `callFlowEnabled` defaults false (`settings.ts:481`); the enable gesture in Settings remains the install consent per the documented model, so replacing the pending-intro gate with `true` does not start background work for anyone who did not opt in. The one capability lost is the version-bump re-disclosure mechanism, acceptable to lose with the dialog. `semanticDiffEnabled` default (true) unchanged; it is local sem analysis, no install consent involved. Vim defaults untouched (`vimModeEnabled` false, Settings > Vim intact). `PLANNOTATOR_AI` gating untouched; the provider effect only runs when capabilities report available.
- **Grid/Clean chooser writes the same setting** the promo's chooser did (`onToggleGrid` -> `configStore.set('gridEnabled', ...)` in both apps), and dismiss now also pins the displayed default (`editor/App.tsx:637`, `review-editor/App.tsx:734`), which is what makes "accept without clicking" an explicit decision. Correct.
- **`useViewportEnvironment` quality.** Fallback path without the API returns layout viewport; pinch zoom (scale delta > 0.01) freezes stage geometry so accessibility zoom is never fought; updates are rAF-coalesced with dedupe against last-written values; ref-counting means two consumers cannot tear down each other; cleanup cancels the pending frame and restores pre-existing property values. Tests exercise the pure calculator (fallback, keyboard inset, pinch, NaN clamping) and the DOM contract (restore, refcount, event coalescing plus listener removal), not implementation echo. Good tests.
- **Viewport meta.** Only `viewport-fit=cover` added; no `maximum-scale`/`user-scalable`, so pinch-zoom accessibility is intact. Both changed files are the Vite entry templates, so the single-file builds and the OpenCode/Pi copies inherit them.
- **Desktop safety.** Safe-area vars compute to 0 off-notch; `.pn-app-viewport` degrades to the old `h-screen` behavior on evergreen browsers; #1277's Call Flow mobile CSS (`review-editor/index.css`) is untouched.
- **No prose-pinned tests.** The dialog test pins the `Continue` action label and asserts the absence of `0.20.0`/`release notes` framing, both deliberate; the rest asserts behavior.
- **CI green** (Windows smoke still pending at review time), typecheck including the strict UI consumer passes locally, 25/25 focused tests pass under `DOM_TESTS=1`.

## Merge order vs #1292

I test-merged both orders onto current main in a scratch worktree: **both orders merge cleanly with no conflicts**, the merged tree keeps `initializeReviewSetup()` at the first-run site and drops the analysis dialog, and the merged review-editor util suite passes (181 tests). Neither PR reverts the other's behavior: #1292's Tree seed is orthogonal to this PR's chain removal, and this PR's edit-mode/spotlight gates simply lose the `analysisIntroVisible` term #1292 never touches.

Recommended order: **#1292 first, then this PR**, for one reason only: #1292 edits the exact AGENTS.md sentence that this PR must also fix (BLOCKER above). Landing #1292 first means this PR rebases onto the "initializes an unseen reviewer's panel to Tree" wording and edits it once to drop "analysis layers", instead of #1292 landing second and reinstating a five-dialog description around already-removed code.

PR #1302, opened 2026-08-13
fix(review): update Codex exec flags

tgenov:
## Summary
- replace removed `codex exec --full-auto` with `--sandbox read-only --approve-for-me`
- update code review, Guided Review, and Tour Codex command builders
- add regression coverage for each command builder

## Reproduction
On Codex CLI 0.101.0, the previous command fails before starting a review:

```text
error: unexpected argument `--full-auto` found
```

## Validation
- `PLANNOTATOR_DATA_DIR=/private/tmp/plannotator-test-data bun test packages/server/codex-review.test.ts packages/server/guide/guide-review.test.ts packages/server/tour/tour-review.test.ts`
- `codex exec --sandbox read-only --approve-for-me --ephemeral --help`


backnotprop:
TLDR: Closing in favor of #1231, which fixes the same breakage with a flag pair that works on codex-cli 0.147.0. Your read-only instinct is right though, and it now has its own issue. AI-assisted verification.

Thanks for the report and the fix attempt, and especially for pushing on the read-only question.

We verified both PRs against codex-cli 0.147.0 (the release that actually removed --full-auto and introduced --approve-for-me). Two findings:

- --sandbox and --approve-for-me are mutually exclusive on 0.147.0. This PR's command fails at launch with: the argument '--sandbox <SANDBOX_MODE>' cannot be used with '--approve-for-me'. The tests pass because they only assert string containment.
- The 0.101.0 attribution looks like a misread: that release still accepted --full-auto and did not yet have --approve-for-me.

#1231 arrived first with the working flag shape and green CI, so it is merged. Not a knock on the effort here: the read-only sandbox is the better design fit for review jobs, and we confirmed sem output writing survives --sandbox read-only when used alone. That direction is now tracked in a follow-up issue crediting you. You will also be credited in the release notes.

PR #1316, opened 2026-08-14
fix(comments): expose skill picker semantics

### LINKED ISSUE #1233: SkillReferenceMenu has no combobox semantics (screen readers get an invisible menu that eats Enter)
The skill-reference dropdown introduced in #1229 (comment composers in plan review and annotate mode) is a div of buttons with no ARIA wiring:

- no role=listbox on the list and no role=option on the rows
- no aria-activedescendant tracking the highlighted row
- no aria-expanded / aria-controls on the textarea

Consequence: while the menu is open, a screen-reader user hears nothing about it, but Enter and Tab are captured by the invisible menu (they insert the highlighted skill instead of typing a newline or leaving the field).

Suggested fix: follow the WAI-ARIA combobox pattern (editable combobox with listbox popup): role=combobox + aria-expanded + aria-controls on the textarea, role=listbox on the menu, role=option + unique ids on rows, aria-activedescendant following the keyboard highlight, and aria-selected on the active row. The in-menu human-only warning should be associated via aria-describedby or announced with a live region.

Relevant files: packages/ui/components/SkillReferenceMenu.tsx, packages/ui/hooks/useSkillReferenceAutocomplete.ts, packages/ui/components/CommentPopover.tsx.

ashish921998:
## Summary

- expose the skill-reference textarea as an editable combobox when skill references are enabled
- connect it to a listbox with uniquely identified, non-tab-stop options
- track the deliberately active keyboard option through `aria-activedescendant` and `aria-selected`
- keep the plain composer unchanged when skill references are disabled

## Validation

- `DOM_TESTS=1 bun test packages/ui` (1,025 passed)
- full TypeScript project checks
- `bun run build:review && bun run build:hook`

Closes #1233

### DIFF
--- a/packages/ui/components/CommentPopover.skillReferences.test.tsx
+++ b/packages/ui/components/CommentPopover.skillReferences.test.tsx
@@ -139,12 +139,14 @@ function activeRow(): Element | null {
   return document.querySelector('[data-skill-item-active]');
 }
 
-function assertOpenCombobox(el: HTMLTextAreaElement): Element {
-  expect(el.getAttribute('role')).toBe('combobox');
+function assertOpenAutocompleteTextbox(el: HTMLTextAreaElement): Element {
+  expect(el.hasAttribute('role')).toBe(false);
   expect(el.getAttribute('aria-autocomplete')).toBe('list');
-  expect(el.getAttribute('aria-expanded')).toBe('true');
+  expect(el.getAttribute('aria-haspopup')).toBe('listbox');
+  expect(el.hasAttribute('aria-expanded')).toBe(false);
   const controls = el.getAttribute('aria-controls');
   expect(controls).not.toBeNull();
+  expect(el.getAttribute('aria-owns')).toBe(controls);
   const listbox = document.getElementById(controls!);
   expect(listbox?.getAttribute('role')).toBe('listbox');
   return listbox!;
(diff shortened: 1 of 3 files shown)


backnotprop (approved):
TLDR: Verified locally, this is a clean implementation of the ARIA combobox contract from #1233. Approving. (AI-assisted review.)

What I checked:

- The pattern is the correct ARIA 1.2 editable combobox: role stays on the textarea, aria-expanded tracks the menu, aria-controls is only present while the listbox is actually rendered, and aria-activedescendant only appears for an explicitly activated row. That last part matters here because it preserves our no-preselection keyboard state machine: a bare $ opens the catalog with nothing active and Enter stays a newline, and your test pins that interaction at the ARIA level too.
- tabIndex=-1 on the options is right for this menu since insertion happens on pointerdown and focus never leaves the textarea. Options were never meant to be tab stops.
- The two SkillReferenceMenu render sites share one listbox id, but the dialog and popover branches are mutually exclusive returns, so no duplicate ids can occur.
- The disabled path is untouched: with skill references off the plain textarea renders with no role or aria-expanded, and your test asserts that.

Verification on the branch: DOM_TESTS=1 bun test packages/ui gives 1025 pass, 0 fail (one unrelated flake on a single run, clean on rerun), and tsc is clean. I approved the CI workflows, which were pending first-contribution approval.

Thanks for the quick turnaround after the mobile release, and for writing behavioral tests against the ARIA contract instead of snapshotting markup.

ashish921998:
@backnotprop Please review

backnotprop:
Reviewed and approved above, CI is fully green. Merge coming shortly.

backnotprop:
Delta-reviewed 72eae812 as well: dropping role=combobox and aria-expanded in favor of the native multiline textbox role with aria-autocomplete, aria-haspopup=listbox, and aria-controls/aria-owns while open is the more correct call. ARIA 1.2 does not support aria-expanded on a textbox, and the combobox role would have overridden the multiline-edit semantics screen readers need for the field's primary job. aria-activedescendant remains valid on textbox, so the active-option tracking is intact. Verified locally: 22 tests pass, typecheck clean. CI approved for the new commit; merging once it's green.

ashish921998:
> Delta-reviewed [72eae81](https://github.com/backnotprop/plannotator/commit/72eae812d8cadd0f4611d7e0820f50f5638e4cec) as well: dropping role=combobox and aria-expanded in favor of the native multiline textbox role with aria-autocomplete, aria-haspopup=listbox, and aria-controls/aria-owns while open is the more correct call. ARIA 1.2 does not support aria-expanded on a textbox, and the combobox role would have overridden the multiline-edit semantics screen readers need for the field's primary job. aria-activedescendant remains valid on textbox, so the active-option tracking is intact. Verified locally: 22 tests pass, typecheck clean. CI approved for the new commit; merging once it's green.

@backnotprop what can I take up next? I would love to contribute more.

backnotprop:
Merged, thanks @ashish921998. This ships in the next release.

PR #493, opened 2026-04-06
feat(hook): notify agent when no VCS detected on plan approval

l10veu602:
## Summary

- When a plan is approved in a directory without version control (git or p4), include a note in the approval output
- The agent (Claude Code / Gemini CLI) receives the message and can ask the user to initialize git before making changes
- Uses existing `detectVcs()` to check for both git and p4

## Test plan

- [ ] Approve a plan in a git repo — no extra message in output
- [ ] Approve a plan in a p4 workspace — no extra message in output  
- [ ] Approve a plan in a directory without VCS — message included in approval output

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1319, opened 2026-08-14
feat(hook): note missing version control in plan approval output (#493 resurrection)

backnotprop:
**TLDR:** Approving a plan in a directory with no version control now appends a short note to the approval text the agent receives, asking it to check with the user about running `git init` before it starts editing. Approvals inside a repo are byte-identical to before, denials are untouched, and a detection failure stays silent.

Resurrects #493 by @l10veu602, which was closed with "closing the branch, keeping the idea": the original diff predated a rewrite of the hook server, so the imports it touched no longer exist and the per-host approval channels have changed. This is a fresh implementation against current main that keeps the original's intent (detect git and p4 through the existing VCS machinery, warn only on approval) and its exact note wording.

## What changed

A small shared decision helper, `packages/shared/no-vcs-note.ts`, owns the rule and the note text:

- approved and no VCS detected: return the note
- approved and VCS detected: return nothing
- denied: return nothing, and never run detection at all
- detection threw or rejected: return nothing

The "assume version control is present on failure" rule is deliberate. A false "you have no git" warning is worse than no warning, so every probe failure is silent.

Detection reuses the existing per-runtime provider set through `detectManagedVcs`, so the Bun CLI covers git, jj, GitButler and p4, and Pi covers the providers Pi registers.

## Per-host delivery

Each host receives approval text through a different channel, so the note is appended to whatever that host already sends. I verified each one rather than assuming a shared field:

| Host | Approval channel today | Note delivery |
| --- | --- | --- |
| Claude Code | `hookSpecificOutput.decision` (`behavior: "allow"`) | appended to `updatedInput.plan` |
| Gemini CLI | `systemMessage` | appended to `systemMessage` |
| OpenCode | the `submit_plan` tool result built by `executeSubmitPlan` | appended to the approved / approved-with-notes prompt |
| Pi | the `plannotator_submit_plan` tool result | appended to the approved / approved-with-notes prompt |
| Codex | `{}` on approval | unchanged, no channel exists |
| Copilot CLI | `{ permissionDecision: "allow" }` | unchanged, no channel exists |

Two of those deserve a note.

**Claude Code needs a maintainer opinion.** The original PR added a `message` field to the allow decision. That field does not exist: Claude Code's PermissionRequest hook output schema accepts only `updatedInput` and `updatedPermissions` on `behavior: "allow"` (`message` exists on deny), and PermissionRequest has no `additionalContext`. I checked this against the schema inside the installed 2.1.232 CLI rather than from memory. Approval feedback is also not threaded to the model on this path at all today, so there was no existing text to ride. The one allow-side field the model reads back is `updatedInput`, which the code already echoes for an unrelated reason, so the note is appended to `updatedInput.plan`, at the tail of the plan the agent is about to execute. It is schema legal and it reaches the model, but it does put the note inside the plan text, so say the word if you would rather Claude Code stay silent like Codex and Copilot until a real channel exists.

**Codex and Copilot CLI get nothing.** Their approval outputs carry no model-facing text at all (Codex prints `{}`, Copilot prints a bare `permissionDecision`), so there is nothing to append to without inventing a field. Left alone on purpose.

## Two-runtime coverage

OpenCode and Pi both format their own approval results, so neither is covered by the Bun server change:

- OpenCode: detection happens where the review actually runs. The embedded runtime probes directly, and the CLI bridge reports it as a `noVcs` flag on the `opencode-plan` JSON output. `executeSubmitPlan` then appends the note to the approval text it returns, so both runtimes behave the same.
- Pi: `apps/pi-extension/index.ts` appends the note to the approved and approved-with-notes prompts, using a lazily loaded probe against Pi's own VCS runtime. The new shared module is vendored through `vendor.sh` and lands in `apps/pi-extension/generated/`.

Pi's external-execution handoff and its non-interactive auto-approve path are untouched: neither is a human approving in a session.

## Tests

- `packages/shared/no-vcs-note.test.ts`: the four decision cases (approved without VCS warns, approved with VCS is silent, denial is silent and never probes, detection failure is silent), plus that appending preserves the original approval text. Asserted on the `No version control detected` substring, not the full sentence.
- `apps/opencode-plugin/submit-plan-executor.test.ts`: real delivery through OpenCode's approval channel, including that a normal approval stays exactly `Plan approved!` and that a denial in an untracked directory gains nothing.

The Claude Code and Gemini output shapes live inside the top-level CLI script and are not reachable from a unit test without extracting them, so they are covered by the shared helper plus review of the two output sites.

## Checks

- `bun test packages/shared apps/opencode-plugin`: passing.
- `bunx tsc --noEmit` on `packages/shared`, `packages/server` and `apps/pi-extension`: clean.
- `bun run --cwd apps/review build && bun run build:hook`: clean. The hook CLI entry also bundles, which is what proves the new shared import resolves for the compiled binary.
- `apps/pi-extension/server.test.ts` has 6 pre-existing failures in this environment (sem sidecar, GitButler CLI, file watcher limits). Confirmed identical on a clean checkout of main.

AI-assisted: implemented with Claude Code, reviewed by me before pushing.


backnotprop:
Closing without merge. On reflection the plan review flow stays deliberately VCS-blind: Plannotator's job at approval time is the document and the decision, and environment checks like version control belong to the agent harness, not the review UI. The rebuild was clean and the original idea from #493 by @l10veu602 was worth exploring, but we are not taking the scope. No further resurrection planned.

PR #1331, opened 2026-08-16
feat(review): support static patch files

dylanvanh:
## Summary

Adds `plannotator review --patch-file <path | ->` to review a unified diff from a file or stdin without creating a branch or pull request.

AI reviews use the supplied patch instead of the local working tree.

`--patch-file` is available through the direct CLI; plugin-hosted review commands remain unchanged.

## Validation

- 99 relevant tests passed
- verified file and stdin input through the review server


backnotprop:
Thank you for this, @dylanvanh — your design is what shipped. `plannotator review --patch-file <path | ->` landed in #1554, which was built directly on this branch (the commit message credits it as its base), and the follow-up extended it past the CLI so the pi and OpenCode hosts get static patch review too rather than the explicit reject you had to leave in. Closing as merged-in-spirit; the `--patch-file` flag people are using today is yours.

PR #1346, opened 2026-08-18
feat(review): collapse linguist-generated files by default like GitHub

### LINKED ISSUE #1317: Generated files marked in .gitattributes open expanded
### What happened?

I mark generated files `linguist-generated=true` in `.gitattributes` so GitHub collapses their diffs. Plannotator's code review opens them expanded like regular files, so the handful of files I want to look at are buried under generated code

From one of my repos:

```
# .gitattributes
/CLAUDE.md linguist-generated=true
/AGENTS.md linguist-generated=true
/.claude/rules/** linguist-generated=true
```

I use [rulesync](https://rulesync.dyoshikawa.com/) to generate these files from the `.rulesync` directory, so I only want to review the `.rulesync/**` files.

```
# .gitattributes
generated/drizzle/**/snapshot.json linguist-generated=true -diff
generated/drizzle/**/*.sql linguist-generated=true
```

I trust Drizzle's migrations and the snapshot is just thousands of lines of noise.

It'd be great if the review UI could take these into account and show them as collapsed by default, similar to what GitHub does.

### Plannotator version

plannotator 0.25.1

### OS

macOS

### Agent

Claude Code

### Where did it happen?

Code review

backnotprop:
## TLDR

Code review now respects `linguist-generated` (and `linguist-generated=true`) from `.gitattributes`. Generated files open collapsed to a one-line bar (filename, +/- counts, a "generated" tag, click to expand), the way GitHub renders them. Closes #1317.

## What works

- Detection uses git's own attribute machinery: one batched `git check-attr --stdin -z` over the served patch's paths at the review cwd. Stacked and negated rules (`-linguist-generated`, `linguist-generated=false`, per-directory files, `info/attributes`) resolve exactly as git resolves them. One subprocess per diff, never per file.
- The sidecar (`generatedFiles?: string[]`) rides `/api/diff` and `/api/diff/switch` in both runtimes (Bun `packages/server/review.ts` and the Pi Node mirror), with the shared resolver in `packages/shared/generated-files.ts`, vendored via `vendor.sh`.
- The client seeds those files collapsed through the existing Pierre item-collapse state (the same mechanism commit diffs use to open folded), so this composes with Viewed, collapse-all, and the commit-diff fold instead of adding a parallel system.
- Expansion is per file and session-local: expanded files stay expanded across re-renders, dock remounts, and diff switches (App-owned state, reported back through the existing collapse-change funnel).
- Provably a view state, not a content state: the patch is never filtered, `snapshotId`/fingerprint/cache semantics are untouched, and annotations, search, content augmentation, and Edit Mode all behave normally once expanded. The file tree and git-status sections list generated files normally; the single-file tab shows the tag but never auto-collapses (opening a tab is already an explicit request to see the file).
- CLAUDE.md endpoint docs updated for both payloads.

## Degrade modes (honest list)

The sidecar is only computed for plain local git sessions. It is omitted, and every file renders expanded as before, for:

- piped patches with no local git context
- workspace multi-repo mode
- PR worktree sessions
- jj, GitButler, and P4 providers
- any `check-attr` failure (git missing, not a work tree, non-zero exit): the resolver returns the empty set rather than guessing

Attributes are read from the working tree at the review cwd (git's default resolution), so a commit diff uses the current `.gitattributes`, not the one at that commit.

## Tests and verification

- `packages/shared/generated-files.test.ts`: attribute resolution against a real temp repo, including a negated rule stacked after a glob, `=false`, dedupe/order, paths with spaces via NUL stdin, and the outside-a-work-tree degrade.
- `packages/server/generated-files-endpoint.test.ts`: dual-runtime (Bun + Pi) endpoint tests for the sidecar, the no-git-access degrade, and the no-attributes case, asserting the raw patch is never filtered.
- `packages/review-editor/components/AllFilesCodeView.generated.test.tsx` (DOM, registered in CI): generated files seed collapsed with diff data and annotations intact, the header chevron expands in place without a CodeView remount and reports to the owner, the "generated" tag renders only for generated files, and a previously expanded file seeds expanded after remount.
- Full `bun run typecheck` green (includes the Pi project). `packages/shared`, `packages/server`, `packages/review-editor` (DOM_TESTS=1), and `apps/pi-extension` suites green; the only failures are 6 pre-existing on main (semantic-diff/GitButler environment tests, identical before and after this change).
- `bun run build:review` and `bun run build:hook` built; verified in a real browser (headless Chrome) against a compiled binary and a fixture repo: `generated/schema.json` collapsed with the tag, `generated/keep.ts` (negated) and `src/app.ts` expanded normally, click expands to the full split diff with the tag retained.
- Guide viewer manifest pin regenerated from a clean `bun install --frozen-lockfile` environment (AllFilesCodeView/FileHeader are bundled into the guides.show viewer).

Closes #1317


backnotprop:
## TLDR

Merge-grade. I built this branch, compiled a binary, and drove it against a fixture repo I wrote myself: the sidecar and the collapse seeding do what the description says, and the presentation-only claim holds under the checks that matter (raw patch byte-identical, snapshot semantics untouched, collapsed items keep their diff and their annotations, expansion runs the normal augmentation path). Three findings, none blocking: one small "does not survive a diff switch" gap on two expansion paths, one missing subprocess timeout, and about eight lines of provably dead bookkeeping.

## What is good

**The detection strategy is the right one.** Going through `git check-attr` instead of parsing `.gitattributes` is what makes stacked and negated rules correct for free. My fixture stacks `generated/** linguist-generated` and then `generated/keep.ts -linguist-generated`, and the negated file resolves to not generated exactly as git resolves it. A hand-rolled first-match parser would have gotten that wrong.

**Batching is real and the cost is nothing.** One `check-attr --stdin -z` per served patch, never per file. Measured on this repo, warm: 50 paths 14.8 ms, 500 paths 14.6 ms, 4000 paths 16.7 ms (3 runs each). It is process-spawn bound, so review size does not move it. NUL framing on both sides means paths with spaces need no quoting layer, and both the server path source (`listPatchFiles`) and the client path source (`parseDiffToFiles`) unquote through the same `packages/core/diff-paths.ts` helpers, so the join between "path in the sidecar" and "path in the file list" cannot drift on odd filenames.

**Presentation-only survives the checks I could think of.** `rawPatch` is untouched on every mode I tried, `snapshotId` is unchanged, the sidecar is computed after the patch rather than around it, the file tree and git-status sections list generated files with their real counts, and a collapsed item still carries `fileDiff` plus projected annotations. Expansion is not a second-class render either: the expanded generated file offers the "2 unmodified lines" context expanders, which only exist after the `/api/file-content` augmentation reparse, so the historic "listed in the tree, renders empty" failure mode is not reachable here. Collapsed items are simply not augmented until they are rendered, which is the same shape commit diffs already had via `seedCollapsed`.

**It reuses the existing mechanism.** Seeding through Pierre item collapse and the existing `reportFileCollapsed` funnel, rather than a parallel collapse system, is the reason this composes with Viewed, collapse-all, and the commit-diff fold instead of fighting them.

**Two-runtime parity is exact.** `packages/server/review.ts:753-763` and `apps/pi-extension/server/serverReview.ts:679-689` differ only in `gitContext` vs `options.gitContext`, the runtime handle, and const-arrow vs function. Same gate, same call sites on `/api/diff` and `/api/diff/switch`, same conditional spread. `vendor.sh` covers the new module and Pi's `files` array already ships `generated/`.

## Findings

### 1. Low: two expansion paths do not survive a diff switch (and a comment that overstates the invariant)

`reportFileCollapsed` at `packages/review-editor/components/AllFilesCodeView.tsx:1043-1052` says:

> Every collapse mutation funnels through here (toggle, viewed+collapse, collapse/expand-all), so the owner's set always mirrors the live item state.

That is not quite true. Three sites mutate `item.collapsed = false` directly. The file-tree scroll target at `:2052` does call `reportFileCollapsed(itemId, false)`. Search-match navigation at `:1586` and sidebar comment navigation at `:2117` do not. So expanding a generated file by clicking a search hit or a comment in the sidebar leaves `expandedGeneratedFiles` empty, and the next identity rebuild re-collapses it.

Reproduced in the browser on this build:

- expand with the header chevron, then Commits detour and back to Tree: file stays expanded (`Expand diff` toggles 0).
- search `openapi-codegen`, press Enter to jump to the match inside the collapsed generated file (it expands), then the same detour: file is collapsed again (`Expand diff` toggles 1).

Small, self-correcting (one more click), and arguably pre-existing behavior that this feature only now makes visible. But the invariant comment should either become true (call `reportFileCollapsed` at those two sites, which also fixes the `syncAllCollapsedMirror` asymmetry already there) or be softened so the next reader does not trust it.

### 2. Low: no `timeoutMs` on the `check-attr` subprocess, on a blocking startup path

`packages/shared/generated-files.ts:62-71` passes `cwd`, `stdin`, and `interaction: "forbid"`, but no `timeoutMs`. The only other `runGit` caller in the repo that feeds stdin, `getGitObjectSizes` at `packages/shared/review-core.ts:869-877`, sets `timeoutMs: 5000`. `/api/diff` is awaited before the reviewer sees anything, so a wedged git (slow network mount, filesystem stall) turns into a review that never opens rather than a review that opens without the sidecar. Everything else in this module is fail-open by design, and a timeout is what completes that.

One caveat if you add it, worth knowing before you do. The Pi runtime writes stdin with a bare `proc.stdin!.end(options.stdin)` at `apps/pi-extension/server/vcs.ts:94` and never attaches a `stdin` error handler, so a child that dies before draining stdin raises an unhandled EPIPE. I reproduced the crash with a faithful replica of `runCommand`: `git check-attr --stdin -z linguist-generated` in a non work tree with 4000 paths (about 250 KB, past the 64 KB pipe buffer) exits `42` from my `uncaughtException` trap rather than resolving. This is pre-existing and today hard to reach here, because the `!gitContext` gate means git is looking at a real work tree, but a `timeoutMs` SIGKILL is exactly the "child dies mid-read" shape. `call-flow.ts:1012` already carries the one-line guard (`child.stdin.on("error", () => {})`) with a comment explaining why. Worth mirroring in `vcs.ts` whether or not you add the timeout, since `getGitObjectSizes` already runs into the same window today.

### 3. Nit: the generated-set delta bookkeeping is dead code

`prevGeneratedRef` plus `collectSetDelta(generatedFiles, prevGeneratedRef.current)` at `:1775`, the two ref writes, and the added effect dep can never produce a non-empty delta: `generatedKey` is part of `fileSetKey`, so any content change to the set remounts, and the effect early-returns on remount. An equal-content set delivered as a new `Set` produces an empty delta. The code comment already says as much ("refreshes nothing... only exists for completeness"). Given the maintainer's bar on over-engineering, this is about eight lines that could go. Everything else added here earns its place, including `expandedGeneratedFiles`, which is what stops a diff switch from re-collapsing what the user opened.

### 4. Notes, not defects

- For `commit:<sha>` views the all-files surface already seeds fully collapsed, so the generated seeding is a no-op there. The sidecar is still computed and the tag still renders, which is the right call.
- Commit diffs resolve attributes from the working tree `.gitattributes`, not the tree at that commit. The PR says this plainly and it matches git's default resolution, so no complaint, just confirming I checked it: my fixture's baseline commit reports `generated/api-schema.json` as generated because the current `.gitattributes` says so.
- The sidecar is server-computed per request and never persisted into drafts, feedback, or share payloads, so no new path leaks repo paths anywhere.

## Visual verification

Built from this branch (`bun install --frozen-lockfile`, `bun run build:review`, `bun run build:hook`, compiled binary), run against a fixture repo I created for this review: `.gitattributes` with `generated/** linguist-generated` plus a negated `generated/keep.ts -linguist-generated`, a 1622-line change to a generated JSON schema, and real edits to two ordinary source files. Fresh temp `PLANNOTATOR_DATA_DIR`, `PLANNOTATOR_BROWSER=/usr/bin/true`, headless Chrome via Playwright 1.62, `PLANNOTATOR_AI=disabled` to keep the session deterministic.

**1. Initial load.** `generated/api-schema.json` opens as a one-line bar with `+1381 -241` and the `generated` tag, while `generated/keep.ts` (the negated rule) and `src/server.ts` render full split diffs below it. The sidebar lists the generated file with its full counts, which is the no-filtering half of the claim.

Initial load with the generated file collapsed

**2. After clicking the chevron.** Full split diff, syntax highlighted, tag retained, and the "2 unmodified lines" context expander present, which means the full-content augmentation ran on expand.

The same file expanded in place

**3. Annotating the expanded generated file.** Line comment created on line 8 of the generated diff, rendered inline, counted in the sidebar and in the header badge, with the header switching to Send Feedback. Annotations are not degraded by the collapse seed.

A line comment on the expanded generated file

**4. Collapsed again, negated rule unaffected.** The chevron toggles both ways, the annotation survives the re-collapse (sidebar still shows 1), and `generated/keep.ts` keeps rendering as an ordinary file with no tag.

Re-collapsed, with the negated-rule file rendering normally

**Video (24 s, 2.4 MB).** One continuous take: load, scroll past the collapsed bar through the normal files, scroll back, expand the generated file, scroll deep into its diff, open the line composer, type, and save the comment. GitHub will not inline-play a `.webm` served from raw.githubusercontent, so this is a download link, not an embedded player:

[05-generated-collapse-take.webm](https://raw.githubusercontent.com/backnotprop/plannotator/pr-assets-1346/1346/05-generated-collapse-take.webm)

Assets live on the orphan branch `pr-assets-1346` under `1346/`.

## What I verified independently, and what I took from the description

Verified myself on this branch:

- `generatedFiles` present and correct on `/api/diff` and on `/api/diff/switch` for `since-base`, `uncommitted`, `unstaged`, `staged` (both with nothing staged, correctly absent, and with the generated file staged, correctly present), `commit:<sha>`, and `hideWhitespace: true`.
- `rawPatch` identical with and without the sidecar, file count unchanged, sections sidecar unaffected.
- Negated rule and `=true` / `=false` resolution against a real repo, plus `check-attr` timing at 50 / 500 / 4000 paths.
- Collapse persistence A/B across a real diff switch (finding 1).
- `packages/shared/generated-files.test.ts` and `packages/server/generated-files-endpoint.test.ts`: 12 pass, 0 fail. `DOM_TESTS=1` on `AllFilesCodeView.generated.test.tsx` plus `AllFilesCodeView.lifecycle.test.tsx`: 12 pass, 0 fail.
- Full `bun run typecheck` green, including `vendor.sh` and the Pi project.
- The Pi EPIPE reproduction described in finding 2, against a replica of `runCommand`.
- CLAUDE.md is a symlink to AGENTS.md, so the endpoint docs land in both.

Taken from the description or from CI rather than reproduced:

- The viewer manifest pin. I relied on the PR's `guides.show viewer (budgets + pinned manifest)` job, which is green on this head, rather than rebuilding from a clean environment myself.
- The jj, GitButler, P4, PR worktree, and workspace degrade paths. I read the gate and confirmed it is a single early return in both runtimes, but I did not stand up those providers.
- The "6 pre-existing failures on main" claim. I ran the targeted suites, not the full one.

Not a blocker anywhere. This is ready as far as I can tell.

AI-assisted (Claude) review under maintainer direction.

backnotprop:
## Round 2: built-in defaults, visible fold, review findings

**TLDR.** This round lands three things on top of the original attribute-based collapse. First, built-in generated defaults: lockfiles and minified assets now collapse on repos with no `.gitattributes` at all, with explicit attributes winning in both directions. Second, a visible collapsed state: the folded card now shows an explicit "Generated file collapsed" strip with the +/- counts instead of a bare header. Third, all three review findings are fixed (the two expansion paths that skipped the collapse-report funnel, the missing `check-attr` timeout plus the Pi stdin EPIPE guard, and the dead delta bookkeeping).

### Built-in generated defaults

`packages/shared/generated-files.ts` (vendored to Pi) now carries `DEFAULT_GENERATED_PATTERNS`: `package-lock.json`, `npm-shrinkwrap.json`, `yarn.lock`, `pnpm-lock.yaml`, `bun.lock`, `bun.lockb`, `Cargo.lock`, `Gemfile.lock`, `composer.lock`, `poetry.lock`, `uv.lock`, `Pipfile.lock`, `go.sum`, `flake.lock`, `packages.lock.json`, plus `*.min.js`, `*.min.css`, `*.map`. Matching is against the path's last segment only, so a directory named `bun.lock` never marks the files inside it.

Precedence follows the linguist semantics exactly, in both directions:

- `linguist-generated` or `linguist-generated=true` marks any file, on the list or not.
- `-linguist-generated` or `linguist-generated=false` un-marks a file even when it is on the built-in list.
- `unspecified` falls through to the built-in list.

`git check-attr` already distinguishes set, unset, and unspecified, so the shared module parses all three states (`parseCheckAttrStates`) and applies them over the name-based default set. In plain local git sessions the defaults always apply and attributes refine them; an attribute-lookup failure leaves the defaults standing. The non-git degrade modes (piped patches, PR worktrees, workspace multi-repo, jj, GitButler, P4), which previously omitted the sidecar entirely, now emit it from the name-based defaults alone, since name matching needs no git. Payload docs updated in AGENTS.md.

### Visible collapsed state

The folded body is no longer empty. A collapsed generated card renders a `GeneratedFileNotice` strip ("Generated file collapsed", the +N/-N counts, "Click to view") styled like the existing below-header notices (`OversizedFileNotice`, `BinaryFileNotice`), same muted tokens, no new palette. The whole strip is clickable and expands through the same `reportFileCollapsed` funnel as the chevron, so a strip-driven expansion survives remounts and diff switches like any other.

### Review findings

- **F1.** Search-match navigation and sidebar comment navigation now call `syncAllCollapsedMirror` plus `reportFileCollapsed` when they expand a collapsed file, matching the file-tree scroll path. A search-driven expansion of a generated file now survives a diff switch, and the funnel invariant comment lists the navigation-driven sites instead of overstating.
- **F2.** The `check-attr` invocation carries `timeoutMs: 5000`, the same bound as `getGitObjectSizes` in review-core. Pi's `vcs.ts` stdin write now attaches the one-line `stdin` error guard (same shape as `call-flow.ts`) before `end()`, since a timeout kill makes the EPIPE race reachable. Re-vendored.
- **F3.** The `prevGeneratedRef` plus `collectSetDelta(generatedFiles, ...)` leg is gone; a changed generated set always remounts via `fileSetKey`, so the delta path was unreachable.

### Tests

New or extended: default-list matching including a `*.min.js` glob case and a `bun.lock/README.md` directory case; both-direction precedence (a lockfile un-marked by `-linguist-generated` resolves expanded, a non-list file marked by attributes resolves collapsed) at the shared layer and end-to-end on `/api/diff` for both runtimes; the non-git sidecar from defaults alone (dual-runtime); the placeholder strip expanding through the funnel (DOM); and search expansion surviving a re-seed round-trip (DOM). Guide viewer manifest pin regenerated from a clean frozen-lockfile worktree.

### Visual verification

Fixture for this round is what was asked for: a repo with a real `bun.lock` generated by `bun add`, then dependency changes so the lock diff is large (+142/-2), two small source edits, and no `.gitattributes` anywhere. Built from this branch, compiled binary, sandboxed `PLANNOTATOR_DATA_DIR`, `PLANNOTATOR_BROWSER=/usr/bin/true`, headless Chrome via Playwright.

**1. All-files view on load.** `bun.lock` folds from the built-in list alone: header with `+142 -2` and the `generated` tag, and below it the new placeholder strip stating "Generated file collapsed +142 -2" with "Click to view". The source files render full diffs.

All-files view with bun.lock folded showing the placeholder strip

**2. After clicking the strip.** The full lockfile diff expands in place, tag retained, and the strip is gone.

Expanded lockfile diff after clicking the strip

Scrolled deeper into the expanded lock diff, with the source files following below:

Expanded lock diff scrolled, source files below

**3. Single-file view.** Opening `bun.lock` from the file tree renders the full diff with the `generated` tag and no auto-collapse, as before.

Single-file view of bun.lock with the generated tag

**Video (about 30 s, 1.7 MB).** One continuous take: load with the strip visible, hover the strip, click it, scroll the expanded lockfile diff, then open the single-file view. Download link, since GitHub does not inline-play `.webm` from raw:

[05-defaults-strip-take.webm](https://raw.githubusercontent.com/backnotprop/plannotator/pr-assets-1346/1346/round2/05-defaults-strip-take.webm)

Round 2 assets live beside the round 1 set on the orphan branch `pr-assets-1346` under `1346/round2/`.

PR #1368, opened 2026-08-21
feat(pi): configurable progress widget style and ordering

autt-carrum:
Add two config-only keys in ~/.plannotator/config.json that reshape the Pi extension's plannotator-progress widget without touching the underlying checklist:

- widgetStyle: 'default' | 'compact' | positive integer N
    * 'default' keeps the historic one-line-per-item layout.
    * 'compact' collapses to a single line summary (📋 done/total · next: <first remaining step>).
    * integer N caps the widget at N item lines, preferring remaining steps and backfilling with recently completed ones so the widget stays useful as items get checked off, with a muted overflow line (… +K done / +K more todo) summarising what was dropped.
    * invalid values (0, negatives, floats, unknown strings) fall back to 'default' — a hand-edited config never blanks the widget.

- widgetMoveCompletedToEnd: boolean (default false)
    * When true, the widget renders completed items after remaining ones; the checklist itself, ${todoList}, and the pi-todos mirror keep the original order.
    * Only a literal boolean true enables it; string 'true' and numeric 1 do not, so a typo cannot flip it on.

Widget-only: the status label, the phase-entry todoList message injected into the conversation, and the pi-todos mirror all keep reading the full unordered checklist. No env-var overrides — appearance is a persistent preference.

Adds 6 tests for the two resolvers (defaults, valid values, invalid values falling back, non-boolean values not flipping the reorder on). Documents both keys in AGENTS.md and in apps/pi-extension/README.md with JSON examples per style value.


autt-carrum:
Sorry, didn't mean to create this PR yet, still testing this code 😅

PR #1371, opened 2026-08-21
feat(pi): make /plannotator-last message anchoring optional

### LINKED ISSUE #1334: Pi: make /plannotator-last message anchoring optional
## Problem

In the Pi extension, `/plannotator-last` injects a large amount of redundant context when sending feedback on the most recent assistant response.

With `@plannotator/pi-extension` `0.27.3`, a small annotation produces a message shaped like:

```text
This feedback applies to the earlier assistant response excerpted below:

> [up to 1,000 characters of the original assistant response]

User feedback:
[the actual annotation feedback]
```

For the normal `/plannotator-last` workflow, the original response is already in Pi's conversation context. Repeating up to 1,000 characters wastes context and makes short feedback unnecessarily verbose.

## Reproduction

1. Use Pi with `@plannotator/pi-extension` `0.27.3`.
2. Generate a normal assistant response.
3. Run `/plannotator-last`.
4. Add one short annotation and send feedback.
5. Observe that the injected follow-up includes a long excerpt of the original response in addition to the feedback.

## Current implementation

`apps/pi-extension/index.ts`:

- `anchorMessageFeedback()` prepends the original response excerpt.
- The result is built before `getAnnotateMessageFeedbackPrompt()` applies `prompts.annotate.runtimes.pi.messageFeedback`.

Therefore, setting `messageFeedback` to `{{feedback}}` removes only the wrapper. It cannot disable the original-message excerpt.

## Expected behavior

Provide a way to disable or limit this anchoring for Pi, for example a configuration option that allows feedback to contain only the annotations. The current behavior could remain the default for stale or previously selected messages, while normal last-message feedback could avoid duplicating context.

## Related issues

- #193 requested reducing annotation message overhead.
- #431 requested context anchoring and was consolidated into #900.
- #900 tracks plan context and feedback policy controls, including omitting or summarizing feedback context.

ashish921998:
## Problem

`/plannotator-last` delivers message-annotation feedback with the original assistant response excerpted before it (up to 1,000 characters via `anchorMessageFeedback()` in `apps/pi-extension/index.ts`). In the normal last-message flow the model already holds that response in its own conversation context, so the excerpt is redundant token overhead that makes even short feedback verbose (#1334).

As the issue notes, the existing `prompts.annotate.runtimes.pi.messageFeedback` template cannot disable this — the excerpt is prepended *before* the template is applied, so `{{feedback}}` only shapes the wrapper, never the anchor.

## Change

Adds an opt-out seam with default behavior unchanged:

- **`packages/shared/config.ts`** — new `annotateMessageAnchoring` config key (default `true`) + `resolveAnnotateMessageAnchoring()`, following the exact precedence pattern of the neighboring resolvers: `PLANNOTATOR_ANNOTATE_MESSAGE_ANCHORING` env var → config key → default `true`, with hand-edited `"false"`-as-string coercion (`coerceConfigBoolean`).
- **`apps/pi-extension/annotate-outcome.ts`** — new pure `shouldPrependMessageAnchor({ feedbackScope, anchoringEnabled, targetMayBeStale })` helper: message-scope feedback (the multi-message picker) never anchors; otherwise anchoring requires both the config seam and the existing possibly-stale-target condition (`shouldAnchorLastMessageFeedback`: different session, or entries added after the target).
- **`apps/pi-extension/index.ts`** — the feedback construction routes through the helper; nothing else changes.
- **Docs** — AGENTS.md env-var table row + Pi README "Annotate last message" section.

## Tests

Four new cases in `annotate-outcome.test.ts` covering the full gating matrix: stale target anchors by default, config-off skips, fresh target skips, `messages` scope never anchors while `message` scope does.

Fixes #1334


ashish921998:
@backnotprop can you review this PR?

PR #1373, opened 2026-08-21
feat: detect the oh-my-pi harness as its own agent origin

FNDEVVE:
## What
Adds `OMPCODE` env detection to Plannotator's origin chain and an `oh-my-pi` entry to `AGENT_CONFIG`, so sessions launched from the [oh-my-pi](https://omp.sh) harness are labeled "Oh My Pi" instead of falling into the `claude-code` default.

## Why a distinct origin, not a claude-code alias
- Plannotator detects agents via env vars; omp sets `OMPCODE=1` (plus a `CLAUDECODE=1` compat shim) in every Bash-tool child process — verified against the harness binary and live env.
- The claude-code-only behaviors (first-run permission-mode dialog, `permissionMode` in the approve payload, approve-with-feedback warning) depend on a real PermissionRequest hook. A bash-invoked `plannotator` never has one — aliasing would show permission-mode UI that cannot take effect. omp doesn't support approve and its planning mode doesn't route through Plannotator yet, so keeping those gates off tracks reality.
- Deny-channel feedback (stdout decision record returned as tool output) works regardless of labeling.

## Why the fallback is untouched
Real Claude Code keeps its label via the terminal fallback; changing it would mislabel bare-shell runs without adding real detection and would force null-origin audits in prompt templates. Note the generic `'Coding Agent'` return in `getAgentName()` is not a working safety net today — the chain always resolves an origin, so it is unreachable dead code by construction; this change deliberately preserves that shape rather than "fixing" the bare-shell mislabel as a side effect.

### Open question
@backnotprop , would you like me to handle the dead `'Coding Agent'` fallback (`getAgentName`, `packages/core/agents.ts:37`)? Today it is unreachable: the detection chain always terminates in `"claude-code"`, so bare-shell runs display "Claude Code" as a cosmetic mislabel. Two ways to fix it, if wanted:
1. **Honest fallback** — add an explicit `process.env.CLAUDECODE ? "claude-code"` leg and let the chain end in `undefined`; undetected sessions then render "Coding Agent". Prompt getters already accept `undefined` and the UI already renders unknown origins generically, so it is nearly free — but it needs a pre-flight proving hooks inherit `CLAUDECODE=1` (otherwise real Claude Code sessions lose their label and permission-mode UI).
2. **Provenance flag** — keep the `"claude-code"` terminal fallback and ship a `detectedViaFallback` boolean in `/api/plan` so the UI can print "Coding Agent" for fallback sessions. Zero behavior risk, more plumbing.
Happy to do either — say the word.

## Screenshots

| Before | After |
|---|---|
| plannotator-omp-before | plannotator-omp-after |

## Details
- No dedicated Ask AI provider for oh-my-pi (falls back to saved/server default).
- `PromptRuntime` union + `PLAN_TOOL_NAMES` extended to keep `Origin`/`PromptRuntime` parallel (typecheck requires it); tool name mirrors `ExitPlanMode`, inert until omp planning integrates.
- Other hosts unaffected: Amp/Droid/Kiro pass `PLANNOTATOR_ORIGIN` explicitly; OpenCode/Codex/Copilot/Gemini keep their env legs.
- Badge color taken from omp.sh branding (favicon gradient pink/purple → fuchsia).

## Testing
`bun run typecheck`; new `packages/core/agents.test.ts`; `prompts.test.ts` addition; end-to-end smoke — `OMPCODE=1 plannotator annotate …` serves `"origin":"oh-my-pi"`, control without `OMPCODE` serves `"claude-code"`.


backnotprop:
Thanks, this is solid. Verified against the oh-my-pi source: the OMPCODE plus CLAUDECODE pair is exactly as you described, and detection works with both set.

One change needed before merge: move the OMPCODE check down so it sits right above the claude-code fallback. OMP exports OMPCODE into its shell sessions, so anything launched from an OMP shell inherits it. As written, running opencode or codex inside an OMP session gets labeled Oh My Pi instead of the runtime the user is actually in. Moving the check to the bottom loses nothing, since claude-code is a fallback rather than a check, and OMPCODE still wins over it.

Everything else we found is on our side (some stale doc lists our own code comment causes contributors to miss) and we will handle it separately. With that one move this is ready.

FNDEVVE:
Fixed in 18d56d5c→f8d10c73: `OMPCODE` now sits right above the `claude-code` fallback, with a comment explaining why (OMP exports `OMPCODE` into its shells). Verified: `OMPCODE=1` alone → `oh-my-pi`; `OMPCODE=1 OPENCODE=1` → `opencode`; neither → `claude-code`. Unit suite unchanged and green.

PR #1379, opened 2026-08-23
fix(ai): a broken RPC pipe must fail the provider, not kill the host (#1378)

### LINKED ISSUE #1378: Pi plan review crashes host with unhandled EPIPE from Pi RPC stdin on Windows
### What happened?

While opening a Plannotator plan review from Pi on Windows, the entire Pi host exits with an uncaught `write EPIPE`. The stack points directly to the Node variant of Plannotator's Pi provider, `PiProcessNode.send()` in `generated/ai/providers/pi-sdk-node.ts`.

Expected behavior: if the nested Pi RPC process closes its stdin, Plannotator should treat that as a provider/process failure and keep the host Pi session alive rather than allowing an unhandled stream error to terminate it.

### Steps to reproduce

1. Run Pi on Windows with `@plannotator/pi-extension` enabled.
2. Start the Plannotator planning flow and submit a plan for review.
3. When the localhost plan-review page opens, Pi can exit with the exception below.

This matches the Plannotator-specific report in earendil-works/pi#5691.

### Stack trace

```text
pi exiting due to uncaughtException:
Error: write EPIPE
    at afterWriteDispatched (node:internal/stream_base_commons:159:15)
    at writeGeneric (node:internal/stream_base_commons:150:3)
    at Socket._writeGeneric (node:net:967:11)
    at Socket._write (node:net:979:8)
    at writeOrBuffer (node:internal/streams/writable:572:12)
    at _write (node:internal/streams/writable:501:10)
    at Writable.write (node:internal/streams/writable:510:10)
    at PiProcessNode.send (C:/Users/<user>/.pi/agent/npm/node_modules/@plannotator/pi-extension/generated/ai/providers/pi-sdk-node.ts:158:21)
    at C:/Users/<user>/.pi/agent/npm/node_modules/@plannotator/pi-extension/generated/ai/providers/pi-sdk-node.ts:167:12
    at new Promise (<anonymous>) {
  errno: -4047,
  code: 'EPIPE',
  syscall: 'write'
}
```

### Root-cause analysis

The current implementation guards only against an already-destroyed stream and then writes directly:

```ts
send(command: Record<string, unknown>): void {
  if (!this.proc?.stdin || this.proc.stdin.destroyed) return;
  this.proc.stdin.write(`${JSON.stringify(command)}\n`);
}
```

That check cannot prevent the child from closing the pipe between the check and the write. In Node.js, this failure can arrive asynchronously as an `error` event on `proc.stdin`; because the stream has no error listener, it escalates to `uncaughtException` and terminates the host process rather than rejecting `sendAndWait()`.

A closely related report, #1039 / #1040, identified the same stream-error mechanism in the Codex app-server provider, although that report was withdrawn after its specific trigger was traced to a stale Codex executable. This report is specifically about `PiProcessNode` and the stack above.

A likely fix is to handle stdin errors at the subprocess wrapper's spawn/write boundary, route them through normal process-end/request rejection handling, and add a regression test proving that a child closing stdin cannot crash the host or leave a pending request unresolved.

### Plannotator version

`@plannotator/pi-extension` 0.27.6

### OS

Windows 11 Pro Education, x64 (build 22621)

### Agent

Pi 0.84.2

### Where did it happen?

Plan review

### Additional environment

- Node.js v22.23.1

### Willing to contribute

I am willing to work on a focused fix and regression test for this. If the proposed direction looks right to the maintainers, I can submit a PR.

### References

- earendil-works/pi#5691 — exact Plannotator stack; Pi maintainer redirected the report here
- #1039 / #1040 — analogous unhandled child-stdin `EPIPE` mechanism in another provider

backnotprop:
## TLDR

Opening a Plannotator plan review from Pi could kill the **entire Pi host** with an uncaught `write EPIPE`. A broken pipe to the nested `pi --mode rpc` child is now handled as an ordinary provider failure: the in-flight query rejects with a clear message, the error surfaces in Ask AI, the provider is marked dead so the next query re-spawns it, and the host stays alive.

Closes #1378. Reported by @Kaelenx, who also did the root-cause analysis in the issue and pointed at the exact line.

## What was actually going wrong

The provider guarded its write like this:

```ts
if (!this.proc?.stdin || this.proc.stdin.destroyed) return;
this.proc.stdin.write(`${JSON.stringify(command)}\n`);
```

Two separate problems compound into a host kill.

**The check cannot close the race.** `destroyed` describes the stream a moment ago. The child can close its end of the pipe between the check and the write, and no amount of pre-checking fixes that; the write itself is the only place the truth is known.

**Nothing was listening for the failure.** Node reports EPIPE from a pipe write either as a synchronous throw or as an `error` event on the stream, and which one you get is a timing race. An `error` event on a Node stream with **no listener** is not "an error you can ignore"; Node re-throws it as an `uncaughtException`. Inside an embedded extension that does not fail the provider, it terminates the process that loaded it, which is the user's whole Pi session.

No stream here had an `error` listener. The `ChildProcess` did not have one either: the spawn handshake registered a one-shot `error` listener and then removed it on success, so from the moment a spawn succeeded the child was completely unguarded for the rest of its life.

Windows only made it likelier to land on the async path during teardown, which is why the report came from there. The mechanism is platform-independent.

There was a second, quieter bug behind the same line. `send()` was fire-and-forget, so a write that failed left the matching `sendAndWait` promise pending **forever** rather than rejecting. The Pi provider has no RPC timeout, so nothing else would have released it.

## The guard pattern

The two JSONL/JSON-RPC providers (Pi Node, Codex app-server) had byte-for-byte the same unguarded `send()`, so rather than patching the reported line twice, the pattern is fixed once in a shared module, `packages/ai/providers/child-io.ts`:

- **`guardChildStreams(proc, label, onFailure)`** attaches `error` listeners to the child process and every piped stream, immediately after spawn and **before** the spawn handshake, so nothing can escalate to `uncaughtException`.
- **`writeChildLine(proc, line, label, onAsyncFailure)`** returns the error for a synchronous failure and routes an asynchronous one through the write callback, so the sync-throw and error-event paths converge on the same handling.
- **`failProcess(error)`** in each provider resolves a broken pipe exactly like any other process end: reject everything in flight, broadcast the process end so a streaming query terminates instead of hanging, reap the child, and leave `alive` false so the existing lifecycle re-spawns on the next query.

I looked for a stronger pattern to copy first. `claude-agent-sdk` spawns nothing of its own, `opencode-sdk` delegates to its SDK, and `codex-app-server` had the identical defect, so there was no good pattern among the siblings to match; the best guards in the repo are elsewhere (`vcs.ts`, `semantic-diff.ts`, `call-flow.ts`) and are swallow-only, which is right for a fire-and-forget CLI write but wrong here, where the failure has to become a visible provider error.

## Scope

In this PR, all in `packages/ai/providers/`:

- `pi-sdk-node.ts` (the reported crash), `codex-app-server.ts` (identical defect, and the mechanism previously reported in #1039 / #1040), `pi-sdk.ts` (the Bun variant's `FileSink` write/flush, guarded symmetrically, including a flush that returns a promise).
- Pi's Node `stderr` moves from an un-drained `"pipe"` to `"ignore"`. Nothing ever read it, and a full stderr pipe deadlocks the child. This is the reasoning already documented in `codex-app-server.ts`; the Pi variant had simply not picked it up.

Deliberately **not** in this PR, found while sweeping for the same class and reported separately so the diff stays reviewable: unguarded `stdin` writes in `apps/pi-extension/server/agent-jobs.ts` and `server/pr.ts`, two fire-and-forget `spawn("open", ...)` calls in `server/integrations.ts` with no `error` listener at all (an ENOENT there is an immediate host kill on any machine without `open`), refusal-branch socket writes in `live-proxy-node.ts` that return before the socket's `error` listener is attached, and the missing RPC timeout that lets a silent-but-alive child hang a Pi query indefinitely.

## Test evidence

`packages/ai/providers/pi-stdio-failure.test.ts`, in two layers.

The end-to-end block runs the provider in a **real `node` child** against a fake Pi that closes its own stdin, following the harness already used by `live-proxy-node.test.ts`. Node is the runtime under Pi, and Bun's `node:stream` shims do not reproduce the unhandled-`error`-kills-the-process rule, so an in-process test could pass against code that still kills the host. The child installs **no** `uncaughtException` handler on purpose: surviving to print its result line is the assertion.

It asserts the in-flight request rejects naming the pipe, `alive` is false, listeners saw the process end, a later fire-and-forget send does not throw, a fresh process talks normally afterwards, and end to end through a session the failure surfaces as an error message while the **next** query on the same session re-spawns and completes.

Against the unfixed provider that child dies exactly as reported:

```
"exitCode": 1,
"stderr": Error: write EPIPE
  code: 'EPIPE',
```

With the fix, exit 0 and empty stderr.

The unit block drives `writeChildLine` / `guardChildStreams` with stream doubles, which is the only way to pin **both** shapes deterministically, since which one a real pipe produces is a platform and timing race: a synchronous throw from `write()`, and an asynchronous error delivered to the write callback.

Checks run: 6/6 new tests pass; `bun test packages/ai/` 122/122; full `bun test` 3819 pass with 15 failures that reproduce identically on unmodified `origin/main` (config-from-disk, improvement-hook and route-404 suites, environmental); `bun run typecheck` clean across all nine projects; `vendor.sh` re-run with `child-io` added to the provider list so `generated/ai/providers/` reflects the fix; `npm pack --dry-run` complete with `generated/ai/providers/child-io.ts` shipping.

### DIFF
--- a/apps/pi-extension/vendor.sh
+++ b/apps/pi-extension/vendor.sh
@@ -124,7 +124,7 @@ for f in index types provider session-manager endpoints context base-session; do
     > "generated/ai/$f.ts"
 done
 
-for f in claude-agent-sdk codex-app-server opencode-sdk command-path pi-sdk pi-sdk-node pi-events; do
+for f in claude-agent-sdk codex-app-server opencode-sdk command-path child-io pi-sdk pi-sdk-node pi-events; do
   src="../../packages/ai/providers/$f.ts"
   printf '// @generated — DO NOT EDIT. Source: packages/ai/providers/%s.ts\n' "$f" | cat - "$src" > "generated/ai/providers/$f.ts"
 done
--- a/packages/ai/providers/child-io.ts
+++ b/packages/ai/providers/child-io.ts
@@ -1,0 +1,92 @@
+/**
+ * Stdio guards for the JSONL/JSON-RPC child processes the AI providers drive.
+ *
+ * Both `PiProcessNode` (pi-sdk-node.ts) and `CodexAppServerProcess`
+ * (codex-app-server.ts) talk to a nested agent over pipes, and a pipe can
+ * break at any instant — the child exits, is killed, or closes stdin while a
+ * command is in flight. Two Node behaviors turn that ordinary condition into
+ * a host-killing crash:
+ *
+ *  1. `stream.write()` on a broken pipe reports `EPIPE` either synchronously
+ *     (throw) or asynchronously (an `error` event on the stream), and which
+ *     one you get is a timing race. A `destroyed` check before the write
+ *     cannot close that race: the child can close the pipe between the check
+ *     and the write.
+ *  2. An `error` event on a Node stream (or on the `ChildProcess` itself)
+ *     with NO listener is re-thrown as an `uncaughtException`. Inside an
+ *     embedded extension that is not "the provider failed" — it terminates
+ *     the HOST agent process.
+ *
+ * That is issue #1378: a Plannotator plan review opened from Pi on Windows
+ * took the whole Pi host down with `write EPIPE` out of `PiProcessNode.send()`.
+ * Windows only made it likelier to land on the async path during teardown;
+ * the mechanism is platform-independent.
+ *
+ * The contract these helpers enforce: a broken pipe is a PROVIDER failure.
+ * It never escalates past the provider, it is reported exactly like any other
+ * process end (in-flight requests reject, listeners see the process end), and
+ * the provider is left dead so the next query re-spawns it.
+ */
+
+import type { ChildProcess } from "node:child_process";
+
+/** Normalize an unknown thrown/emitted value to an Error. */
+export function toChildError(err: unknown): Error {
+	return err instanceof Error ? err : new Error(String(err));
+}
+
+/**
+ * Attach `error` listeners to the child process and every piped stream so a
+ * broken pipe can never reach `uncaughtException`.
+ *
+ * Call this immediately after `spawn()` returns — before awaiting the spawn
+ * handshake — so a stream that fails during startup is already covered. The
+ * spawn handshake may register its own one-shot `error` listener; Node allows
+ * several, and `onFailure` is expected to be idempotent.
+ */
+export function guardChildStreams(
+	proc: ChildProcess,
+	label: string,
+	onFailure: (error: Error) => void,
+): void {
+	const report = (stream: string) => (err: unknown) => {
+		onFailure(new Error(`${label} ${stream} failed: ${toChildError(err).message}`));
+	};
+	proc.on("error", report("process"));
+	proc.stdin?.on("error", report("stdin"));
+	proc.stdout?.on("error", report("stdout"));
+	proc.stderr?.on("error", report("stderr"));
+}
+
+/**
+ * Write one already-newline-terminated line to a child's stdin.
+ *
+ * Returns the Error the write failed with, or `null` when the bytes were
+ * accepted by the stream. Failures that only surface later (the write
+ * callback, or an `error` event covered by {@link guardChildStreams}) are
+ * reported through `onAsyncFailure` instead, so BOTH the sync-throw and the
+ * async-event paths end up at the caller's failure handling.
+ */
+export function writeChildLine(
+	proc: ChildProcess | null,
+	line: string,
+	label: string,
+	onAsyncFailure: (error: Error) => void,
+): Error | null {
+	const stdin = proc?.stdin;
+	if (!stdin || stdin.destroyed || stdin.writableEnded) {
+		return new Error(`${label} stdin is closed`);
+	}
+	try {
+		stdin.write(line, (err) => {
+			if (err) {
+				onAsyncFailure(
+					new Error(`${label} stdin write failed: ${toChildError(err).message}`),
+				);
+			}
+		});
+	} catch (err) {
+		return new Error(`${label} stdin write failed: ${toChildError(err).message}`);
+	}
+	return null;
+}
--- a/packages/ai/providers/codex-app-server.ts
+++ b/packages/ai/providers/codex-app-server.ts
@@ -43,8 +43,10 @@ import {
   killWindowsProcessTree,
   resolveWindowsCommandShim,
 } from "./command-path.ts";
+import { guardChildStreams, writeChildLine } from "./child-io.ts";
 
 const PROVIDER_NAME = "codex-sdk";
+const CODEX_PROCESS_LABEL = "Codex app-server";
 const DEFAULT_MODEL = "gpt-5.6-sol";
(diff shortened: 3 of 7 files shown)


Kaelenx:
Thanks for the quick fix and for crediting the report. I’m available to validate it on Windows 11 with Pi 0.84.2 if useful.

backnotprop:
That would genuinely help, since the crash manifests most readily on Windows and our CI exercises the guard with simulated pipes. To test the branch build: clone the repo, checkout fix/1378-pi-provider-stdio-epipe (or fetch the PR head), run bash apps/pi-extension/vendor.sh, then point Pi at the local extension path instead of the npm package and repeat your repro from the issue. Expected: the review page failure surfaces as an Ask AI provider error inside the session, Pi stays alive, and the next query works. If you see any exit or stack trace at all, paste it here. It ships in v0.27.7 either way, but a confirmation from the exact machine that crashed is the best evidence we can get.

Kaelenx:
Verified on the original Windows environment against commit `8445670e180044cd83ad4eeebcb9a95ae693ad7d`.

Environment:
- Windows 11 x64 (build 22621)
- Pi 0.84.2
- Node 22.23.1
- Bun 1.4.0

I built the local Pi extension and exercised the full plan-review startup path under a real Pi host. To make the timing deterministic, I put a native Windows fake `pi.exe` first on the child-discovery PATH; its first process closes stdin before the provider writes, producing a real Windows broken pipe rather than a mocked stream.

Control result on base `db86d38ca46dc5faf91bbec9dd9d2d0e824b2321`: the host exits 1 with the exact reported stack:

```text
Error: write EPIPE
  at PiProcessNode.send (.../generated/ai/providers/pi-sdk-node.ts:158:21)
  at PiProcessNode.sendAndWait (.../pi-sdk-node.ts:165:12)
  at PiSDKNodeProvider.fetchModels (.../pi-sdk-node.ts:258:12)
code: 'EPIPE'
```

PR result with the same harness:
- the broken child was launched and closed stdin;
- the plan-review server started and served both `/api/plan` and the HTML page;
- the outer Pi host remained alive and answered RPC state requests before and after recovery;
- no `uncaughtException`, EPIPE stack, or host exit occurred;
- after changing the fake child to healthy mode, a new `pi-sdk` Ask AI session/query returned a successful `result`, confirming re-spawn/recovery.

I also ran the provider child harness directly with a native Windows broken pipe. It reported the in-flight request rejected with `Pi process stdin write failed: write EPIPE`, `alive: false`, `process_exited` observed, no throw on a later fire-and-forget send, and a successful subsequent query after re-spawn. Process exit was 0 with no uncaught stderr.

Additional checks:
- `bun run build:pi` passed
- `bun test packages/ai/`: 121 pass, 1 expected Windows skip, 0 fail

So this fixes the reported host-killing failure on the machine/runtime where it occurred.

PR #1393, opened 2026-08-26
feat(webmcp): phase 1 provider for plan review and annotate

backnotprop:
## What this does

Phase 1 of WebMCP support, per `DESIGN_webmcp-support.md`. Plan review and every annotate surface (markdown, raw HTML, live app, folder) become a WebMCP provider: a browser-integrated agent that speaks `document.modelContext` gets a small read-and-comment catalog instead of scraping the DOM.

Tools, registered under the `plannotator.` prefix: `read_document` (the whole situation in one zero-argument call: session, text windowed at 16k chars cut at a block boundary, outline with per-section counts, every comment with quote, context and `isNew`, sibling documents, nudges), `add_comments` (batch; anchors by `inReplyTo`, exact `quote` with `section` to disambiguate, `section` alone, or a document-level note; idempotent by `requestId`), `update_comment`, `remove_comments`, `reveal`, `nudge_user` (one transient banner), and `list_documents` in folder sessions. Every response carries the twelve nudge codes computed synchronously from state the page already holds, so the agent learns about the human's new comments, an open composer, on-disk staleness, edit mode, page changes and sibling-document activity on the call it was going to make anyway.

The rules that make this safe to ship without any consent dialog: decisions are human (no tool approves, denies, submits, closes, stages or marks viewed, and the catalog test pins the names), the agent may only edit or remove comments stamped `source: "browser-agent"` (the human's answer `forbidden`), and nothing is ever registered inside the untrusted iframes.

Architecture mirrors the shortcut system: the engine lives in `packages/ui/webmcp/` (`modelContext.ts` is the only file that spells the spec surface, with local structural types and no `webmcp-types` dependency; the envelope wrapper turns any throw or unserializable value into data, since the spec surfaces both as a bare `UnknownError`; the per-document registry reconciles by name with one `AbortController` per tool), the catalog in `packages/editor/webmcp/` is built over a narrow adapter and never imports App, and `configurePlannotatorUI` gains one optional `webmcp: { enabled, namePrefix }` seam defaulting to today's behavior. No server changes; Pi and OpenCode get it through the built HTML.

## Zero footprint without WebMCP, and how it was proven

Feature detection is one `typeof` check per mount (`resolveModelContext`). When `document.modelContext` is absent the catalog is never built, nothing is registered, the tracker never observes, no DOM is rendered (indicator, banner, settings row), and no request or timer is created.

Proof: the origin/main build and this branch's build were extracted into two equivalent non-git scratch trees and served the same annotate session on fixed ports with a sandboxed `PLANNOTATOR_DATA_DIR` and a no-op `PLANNOTATOR_BROWSER`. Chrome for Testing (no WebMCP) loaded both with an init script counting `setTimeout` / `setInterval` / `requestAnimationFrame` / `requestIdleCallback` registrations from document start. After normalizing only the server-derived project name in the title, the rendered DOM shells were byte-identical (44,607 chars each), the request lists were identical (11 requests, the pre-existing session bootstrap), the console output was identical, and the timer counts were identical (1 timeout, 0 intervals, 3 animation frames). Videos, screenshots and JSON captures are in the evidence directory named in the report.

## With the API

Google Chrome 152.0.7977.64 (the installed browser) launched through Playwright with a fresh temporary user-data-dir and `--enable-features=WebMCPTesting` (the switch behind `chrome://flags/#enable-webmcp-testing`; `--enable-features=WebMCP`, `--enable-blink-features=WebMCP` and `--enable-experimental-web-platform-features` also expose it, `--enable-blink-features=ModelContext` does not).

After page load the six tools are listed by `getTools()` with the specified descriptions and hints (`read_document` carries `readOnlyHint` and `untrustedContentHint`; Chrome drops the forward-compatible `destructiveHint` as expected), and nothing else happened: no indicator, no banner, no dialog, the same request list and timer counts as a browser without the API. `executeTool` from page context returned the full envelope for `read_document`. A comment added by the human through the UI came back on the next read with `isNew: true` and an `annotations_new` nudge naming its id, and while the composer was open the response carried `composer_open`. `add_comments` created a quote-anchored comment (highlighted in the document) and a threaded reply to the human's note, both stamped `browser-agent` and rendered in the panel like external-annotation comments; the same `requestId` returned `deduplicated: true` with the same id. `update_comment` and `remove_comments` answered `forbidden` for the human's comment and succeeded on the agent's. `reveal` scrolled to a section and selected a comment; `nudge_user` showed the banner, which dismissed, and a 281-character message was refused with `invalid_input`. The header indicator appeared only after the first call. Flipping the Settings toggle off left `getTools()` empty and flipping it back re-registered all six. Inside the sandboxed HTML iframe and inside the live-app iframe, `getTools()` and `registerTool()` both reject with `NotAllowedError: Access to the feature "tools" is disallowed by permissions policy`, so a framed page can neither see nor impersonate Plannotator's tools. A plan-mode session (ExitPlanMode event on stdin) registered the same catalog and reported `mode: "plan"`.

## What could not be verified

The Model Context Tool Inspector extension and a real built-in browser agent were not exercised; every call was made from page context through the spec's `executeTool`. Cross-document comments on a sibling document that is not open are refused with `not_available` and a hint to `reveal { path }` first, because the linked-doc cache is a copy; commenting there in one call is deferred. `composer_open` carries no section on this surface (the composer is detected through the DOM). Code review (phase 2) and enumerating the annotated app's own tools (phase 3) are not in this PR.

## Product change to call out: `inReplyTo`

`Annotation` gains one optional field, `inReplyTo`. A reply inherits its parent's anchor, renders indented under the parent in the annotations panel, and exports nested under the parent's entry as a `**Replies:**` block so the coding agent reads the exchange in order. Annotations without the field render and export byte-identically to before (pinned by tests). Drafts round-trip it; share links deliberately do not (a reply shares as a plain comment on the same quote, the existing text-restore contract). This is the only data-model change and it is visible to humans, so it deserves a product look at the indent and the export grouping.

## Tests

`bun run typecheck`: clean. `bun test packages/ui packages/editor`: 818 pass, 520 skip (DOM-gated files skip without `DOM_TESTS=1`), 0 fail. The new plain suites (engine, tracker, nudges, catalog, export, sharing, configure): 82 pass, 6 skip, 0 fail. The DOM-gated group added to `test.yml` (`useToolset.seam`, `AnnotationPanel.inReplyTo`, `useDocumentWebMcp.toggle`, `AppHeader.webmcpIndicator`): 11 pass, 0 fail.


backnotprop:
Review record for this PR before merge.

Three independent review rounds against the design of record. Round one requested changes on two blockers: the settings-registry entry seeded a cookie for every user, including browsers without WebMCP and guides.show visitors, and reveal with a path read state before the navigation committed and reported not_found after moving the view. Both fixed: the preference moved out of the registry and is written only on opt-out, which also made the portable viewer bundle byte-identical to main (no manifest change in the PR); reveal now resolves on the commit that opens the document, with a test that fails on the old code. Round one also had folder sessions wired to the file browser, ownership keyed on ids claimed this page load instead of the source stamp, requestId replay after a human deletion answering conflict, the docs page and manual checklist added. Round two approved and surfaced three small items, all closed in round three: the agent's own removals are no longer echoed back to it as human removals, reveal fails fast for paths that cannot open, and the folder-navigation prose is accurate.

Proof of the two non-negotiables, recorded with video: in a browser without document.modelContext, DOM, network requests, console, timers, and cookies are identical between a main build and this branch; in Chrome 152 under a throwaway profile with --enable-features=WebMCPTesting, the tools register on load and nothing else happens (no indicator, prompt, request, or timer) until executeTool is called, after which every catalog scenario passed, and inside both annotate iframes getTools and registerTool reject with NotAllowedError.

Final head verified: typecheck clean, 821 pass in ui and editor, DOM suites green, all builds and viewer checks pass, CI green.

PR #1407, opened 2026-08-28
fix(pi): notify review URL in RPC sessions

### LINKED ISSUE #1406: Pi Web RPC sessions do not show the Plannotator review URL
### What happened?

When Plannotator is used from a Pi session hosted through Pi Web's RPC mode, the review server starts, but the review URL is not reliably shown in the Pi session UI.

Pi Web binds extensions with `ExtensionContext.mode === "rpc"`. In this mode, `isRemoteSession()` is normally false because the process is not necessarily running over SSH and does not require `PLANNOTATOR_REMOTE=1`.

`openBrowser()` may report that launching a browser succeeded after spawning a local browser command. However, the browser is running in the Pi Web host context and the user may not have access to it. Because the session is not identified as remote, `openBrowserForServer()` can return without notifying the RPC UI of the URL.

### Plannotator version

`@plannotator/pi-extension 0.27.8`

### OS

Linux

### Agent

Pi

### Where did it happen?

Code review

### Steps to reproduce

1. Run Pi Web with a normal RPC session.
2. Load the Plannotator Pi extension.
3. Trigger `/plannotator-review`, `/plannotator-annotate`, or `/plannotator-last`.
4. Observe the Pi Web session output.

### Expected behavior

The Pi RPC session should always receive an informational notification containing the Plannotator review URL, for example:

```text
[Plannotator] http://localhost:<port>
```

### Actual behavior

The review server is started, but the review URL is not reliably visible in the RPC session. The result depends on whether the host process believes that a local browser was opened.

### Additional context

This is related to the remote URL notification work in [#663](https://github.com/backnotprop/plannotator/pull/663), but Pi Web RPC mode is distinct from Plannotator's existing `isRemoteSession()` detection.

The desired behavior is limited to URL notification. It should not change server binding, authentication, port forwarding, advertised-host handling, or remote-session detection.

Shujakuinkuraudo:
## Summary

Pi Web runs extensions with `ExtensionContext.mode === "rpc"`. In this mode,
Plannotator's `isRemoteSession()` can remain false even though the user cannot
directly access the browser launched by the Pi Web host process.

`openBrowserForServer()` currently notifies the URL only for remote sessions or
when browser launch reports failure. A successful browser spawn in the Pi Web
host can therefore hide the URL from the RPC session.

Treat RPC extension contexts like remote sessions for the purpose of URL
notification:

```diff
- if (isRemoteSession()) {
+ if (ctx.mode === "rpc" || isRemoteSession()) {
    ctx.ui.notify(`[Plannotator] ${serverUrl}`, "info");
  }
```

## Behavior

- Local interactive sessions keep their existing behavior.
- Existing SSH and `PLANNOTATOR_REMOTE` sessions keep their existing behavior.
- Pi Web RPC sessions receive the review URL through `ctx.ui.notify()` even
  when the browser launcher reports success.
- Server binding, hostname selection, ports, authentication, browser launch,
  and port forwarding are unchanged.
- This does not make the random review port directly reachable through Pi Web;
  it only exposes the URL to the user so an explicitly configured access path
  can be used.

## Test plan

- Added a regression test for an RPC context with `PLANNOTATOR_REMOTE=0`.
- The test points `BROWSER` at the current Node executable, so no real browser
  is opened while the launcher still reports success.
- The test verifies that the RPC context receives the `[Plannotator] <url>`
  notification.
- `git diff --check` passes.
- The full Bun test suite could not be run locally because Bun is not installed
  in the development environment; GitHub Actions should run the repository's
  standard checks.

Closes #1406

### DIFF
--- a/apps/pi-extension/plannotator-browser.test.ts
+++ b/apps/pi-extension/plannotator-browser.test.ts
@@ -40,6 +40,48 @@ describe("shouldUseLocalPrCheckout", () => {
 	});
 });
 
+describe("browser URL notification", () => {
+	test("notifies RPC sessions even when the browser launcher reports success", async () => {
+		const previousRemote = process.env.PLANNOTATOR_REMOTE;
+		const previousPlannotatorBrowser = process.env.PLANNOTATOR_BROWSER;
+		const previousBrowser = process.env.BROWSER;
+		const notifications: string[] = [];
+
+		process.env.PLANNOTATOR_REMOTE = "0";
+		delete process.env.PLANNOTATOR_BROWSER;
+		process.env.BROWSER = process.execPath;
+
+		try {
+			const ctx = {
+				mode: "rpc",
+				hasUI: true,
+				ui: {
+					notify(message: string) {
+						notifications.push(message);
+					},
+				},
+			} as any;
+			const session = startBrowserDecisionSession(
+				{ url: "http://localhost:4321", stop() {} },
+				ctx,
+				() => new Promise<never>(() => {}),
+			);
+
+			await new Promise((resolve) => setTimeout(resolve, 0));
+			session.stop();
+
+			expect(notifications).toEqual(["[Plannotator] http://localhost:4321"]);
+		} finally {
+			if (previousRemote === undefined) delete process.env.PLANNOTATOR_REMOTE;
+			else process.env.PLANNOTATOR_REMOTE = previousRemote;
+			if (previousPlannotatorBrowser === undefined) delete process.env.PLANNOTATOR_BROWSER;
+			else process.env.PLANNOTATOR_BROWSER = previousPlannotatorBrowser;
+			if (previousBrowser === undefined) delete process.env.BROWSER;
+			else process.env.BROWSER = previousBrowser;
+		}
+	});
+});
+
 describe("browser session cleanup", () => {
 	const ctx = {
 		hasUI: true,
--- a/apps/pi-extension/plannotator-browser.ts
+++ b/apps/pi-extension/plannotator-browser.ts
@@ -180,7 +180,7 @@ export async function startServerWithSelfPreemption<T>(
 
 async function openBrowserForServer(serverUrl: string, ctx: ExtensionContext): Promise<void> {
 	const browserResult = await openBrowser(serverUrl);
-	if (isRemoteSession()) {
+	if (ctx.mode === "rpc" || isRemoteSession()) {
 		ctx.ui.notify(`[Plannotator] ${serverUrl}`, "info");
 	} else if (!browserResult.opened) {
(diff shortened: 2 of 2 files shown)


backnotprop:
Thanks for the careful writeup on both the issue and the PR. The diagnosis is right: ctx.mode is the documented discriminator (ExtensionMode = "tui" | "rpc" | "json" | "print", required, and hasUI is true in both TUI and RPC so it cannot stand in for it). The one-line change typechecks and does fix the submit_plan and archive paths.

Three things before this can land.

1. It probably does not fix the three commands #1406 names. apps/pi-extension/index.ts:182-194 has a second, independent gate:

    function sessionOpenedMessage(label: string, url: string): string {
        if (!isRemoteSession()) return `${label}. You can keep chatting while it runs.`;

called at index.ts:659, :958 and :1048 for /plannotator-review, /plannotator-annotate and /plannotator-last. Its own comment says the after-turn notify inside openBrowserForServer "fires too late to render". So under RPC those three still show a URL-less message. Could you check in a real Pi Web session whether the notify renders there, and extend sessionOpenedMessage if it does not?

2. New stale-ctx exposure. ctx.mode is a guarded getter that throws after session replacement, which this repo documents at apps/pi-extension/current-pi-session.ts:40-66 and wraps in isCtxAlive(). Today the local browser-opened path reads no ctx property at all; after this change ctx.mode is read on every call. Via startBrowserDecisionSession that turns into misleading console.error noise (plannotator-browser.ts:266-272), and via openBrowserAndWait there is no catch, so an archive open that used to succeed now answers request.respond({ status: "error" }) at plannotator-events.ts:429. Please guard the read.

3. The test spawns a real process. process.execPath is not in NOOP_BROWSER_VALUES (server/network.ts:17), so network.ts:341-351 runs spawn(<bun>, ["http://localhost:4321"], { detached: true }) on macOS and Linux CI. The file header promises these tests never spawn a browser. I know the obvious alternative is worse (a sentinel BROWSER with PLANNOTATOR_REMOTE=0 falls through to resolveUseGlimpse, which reads the real ~/.plannotator/config.json), so the cleaner fix is a seam: extract a pure predicate like shouldAnnounceUrl(mode, isRemote, opened) and test that with no env mutation and no spawn.

Heads up that #1184 (open) rewrites this exact hunk to notify unconditionally, which covers RPC, Codex desktop and any future host without reading ctx.mode at all. That is the same argument your issue makes one host later, and it is probably where this should converge. Worth syncing with that PR rather than both of you landing on the same four lines.

backnotprop:
Merged, thanks @Shujakuinkuraudo. `ctx.mode` is already read elsewhere in the extension and the guard adds only the notify, so this was an easy yes. Ships in the next release.

PR #1414, opened 2026-08-28
feat(opencode): preserve model choice on plan handoff

rohanhasabe-esper:
## Summary
- preserve the active OpenCode model when approving a plan
- let reviewers choose an agent default or a specific available model before handoff
- display the resolved current model in the Approve dropdown and widen it to avoid truncation

image

## Verification
- `bun test apps/opencode-plugin packages/ui packages/editor packages/server apps/hook`
- `DOM_TESTS=1 bun test packages/ui/components/HtmlSurfaceControls.test.tsx packages/ui/components/ApproveDropdown.currentModel.test.tsx`
- `bunx tsc --noEmit -p packages/ui/tsconfig.json`
- `bun run --cwd apps/review build && bun run build:hook && bun run build:opencode`
- Manually submitted and approved a dummy plan through OpenCode; confirmed the current model appears in the handoff dropdown.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -149,6 +149,7 @@ jobs:
           packages/editor/App.archiveReadOnly.test.tsx
           packages/editor/App.htmlChrome.test.tsx
           packages/ui/components/HtmlSurfaceControls.test.tsx
+          packages/ui/components/ApproveDropdown.currentModel.test.tsx
           packages/ui/components/AnnotationPanel.unanchored.test.tsx
           packages/ui/hooks/useHtmlRefresh.test.tsx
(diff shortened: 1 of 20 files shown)


backnotprop:
Thanks for digging into this, @rohanhasabe-esper — the underlying complaint was real and it's now fixed on the V2 path: #1569 makes the OpenCode 2 adapter apply `switchModel` alongside `switchAgent` so the session's model follows the approved agent. That landed with the opposite resolution to yours (the target agent's configured model wins rather than the previously active one), so the picker and plumbing here would now pull against it. Closing for that reason — but if you still hit a wrong model on the V1 `session.prompt` path, that branch genuinely is unfixed and a small PR reconciled with #1569's semantics would be very welcome.

PR #1415, opened 2026-08-28
fix(hook): resolve Claude transcript by session id

### LINKED ISSUE #985: /plannotator-last serves another session's message when cwd diverges from the transcript's project slug
# `/plannotator-last` serves another session's message when cwd diverges from the transcript's project slug

**Plugin:** plannotator 0.21.2 (Claude Code plugin)
**Host:** Claude Code 2.1.x, macOS

## Summary

`plannotator last` / `annotate-last` resolves "the last assistant message" from
the **current working directory**, not from the invoking session's identity. When
a session's cwd no longer maps to the project slug where its transcript is stored,
and another Claude Code session is live under a shared ancestor directory, the
ancestor-directory-walk fallback picks up the *other* session's transcript and
serves its most recent message. The annotation/review then applies to a message
the user never wrote in this session.

## Impact

- The message rendered for annotation is from the wrong session. If the user
  doesn't catch the mismatch, any annotation or review they write attaches to a
  different session's content.
- In practice the swapped-in message is visibly not what the user was working on,
  so they close the Plannotator session without annotating. The intended review
  never happens and the workflow is derailed — the failure is loud, but it still
  blocks the review.
- Triggers in a very normal setup: git worktrees plus more than one concurrent
  Claude Code session in the same repo tree.

## Reproduction

1. Launch a Claude Code session in worktree A (`…/repo/.worktrees/feat-a`). Its
   transcript is written to `~/.claude/projects/{slug-A}/{sessionId}.jsonl`.
2. Mid-session, switch the session's cwd into a different worktree B
   (`…/repo/.worktrees/feat-b`) — e.g. via an `EnterWorktree`-style cwd change.
   The transcript stays under `{slug-A}`; the `{slug-B}` project dir is empty (no
   `.jsonl`).
3. Have a second Claude Code session live in the base checkout (`…/repo`), session
   X, actively writing `~/.claude/projects/{slug-base}/{X}.jsonl`.
4. From session A (now cwd = worktree B), run `/plannotator-last`.

**Expected:** the last assistant message from session A.
**Actual:** the last assistant message from session X (the base-checkout session),
because its transcript is the most-recently-modified one found by walking up from
worktree B's cwd to the shared base directory.

## Root cause

In `apps/hook/server/index.ts` (`annotate-last` / `last` branch) the Claude path
resolves a transcript via a precedence ladder, all keyed off
`projectRoot = PLANNOTATOR_CWD || process.cwd()`:

1. **Ancestor-PID metadata** — walk the process tree, read
   `~/.claude/sessions/<pid>.json` at each hop. Fails here: `/plannotator-last`
   runs as a detached background shell, so its process tree no longer chains to
   the Claude Code process, and no `<pid>.json` is found.
2. **Cwd-scan of session metadata** — read `~/.claude/sessions/*.json`, filter by
   `cwd === projectRoot`, pick newest `startedAt`. Even when this resolves the
   correct `sessionId`, the subsequent transcript lookup is derived from the cwd
   slug (`findSessionLogsForCwd` in `session-log.ts`), which points at the empty
   `{slug-B}` dir — so it yields no message and falls through.
3. **Cwd slug match (mtime)** — `findSessionLogsForCwd(projectRoot)` →
   `~/.claude/projects/{slug-B}/*.jsonl`. Empty. Its only fallback is a
   case-insensitive directory match (for Windows drive-letter casing), which does
   not bridge to `{slug-A}`. Miss.
4. **Ancestor directory walk** — climbs `projectRoot`'s parents. From worktree B it
   reaches the base repo dir, whose `{slug-base}` holds session X's transcript.
   That transcript is the most-recently-modified, so its last message is returned.

The two structural problems:

- **Transcript lookup is cwd-slug-based, not sessionId-based.** Step 2 can resolve
  the right `sessionId` from metadata but then can't find its transcript because
  the file lives under the slug of the *launch* cwd, not the *current* cwd. A
  session whose cwd has moved (worktrees, `cd`) becomes unfindable by its own id.
- **The ancestor-directory-walk fallback crosses into directories owned by other
  live sessions.** Under a shared monorepo/worktree root, climbing parents lands
  on a slug that belongs to a different, concurrently-active session, and
  "newest `.jsonl` wins" hands back that session's message.

The code comments already flag step 3 as "Fragile when multiple sessions exist for
the same project" — this is that fragility, made worse by worktree cwd divergence.

## Suggested fix

1. **Resolve the transcript by `sessionId`, not by cwd slug.** Once step 1/2 yields
   a `sessionId` (or once `<pid>.json` gives one), locate `{sessionId}.jsonl` by
   searching across `~/.claude/projects/*/` (or store/read the transcript path in
   the session metadata directly). This makes resolution robust to a session whose
   cwd moved after launch.
2. **Don't let the ancestor-directory-walk cross into another live session's
   directory.** Gate the walk so it never selects a transcript belonging to a
   `sessions/*.json` entry whose `status` is active/`busy` and whose `sessionId`
   differs from the resolving session. Better: if any precise selector identified a
   `sessionId`, never fall through to mtime/ancestor heuristics at all — fail
   closed with "couldn't find this session's transcript" instead of serving a
   sibling's.
3. **Make the background-shell PID walk survivable**, or pass the host session id
   into the command explicitly (e.g. a `--session-id` arg / env var the slash
   command populates) so step 1 doesn't depend on an intact process tree.

## Evidence from a live occurrence

- Invoking session's metadata correctly recorded its moved cwd
  (`sessions/<pid>.json`: `cwd` = worktree B, `sessionId` = A), yet its transcript
  `A.jsonl` was under `{slug-A}` and `{slug-B}` contained no `.jsonl`.
- The served message came verbatim from a different session's transcript
  (`X.jsonl`) under `{slug-base}`, which was the most-recently-modified jsonl
  reachable by walking worktree B's cwd up to the base repo dir.

rNoz:
## Summary

- resolve Claude transcripts by the metadata session ID across project directories instead of assuming the current working-directory slug
- fail closed when precise or relevant invalid metadata cannot safely resolve a transcript, allowing mtime heuristics only when metadata is unavailable
- preserve `/clear` handling while avoiding concurrent registered sessions

## Validation

- `bun test apps/hook/server/session-log.test.ts` (126 passed)
- `bun test apps/hook/server --timeout 15000` (253 passed)
- `PATH="$PWD/apps/hook/node_modules/.bin:$PATH" bun run typecheck`
- `git diff --check upstream/main`

Closes #985.

### DIFF
--- a/apps/hook/server/index.ts
+++ b/apps/hook/server/index.ts
@@ -158,8 +158,7 @@ import {
   findSessionLogsForCwd,
   getRecentRenderedMessages,
   resolveDroidSessionLogForCwd,
-  resolveSessionLogByAncestorPids,
-  resolveSessionLogByCwdScan,
+  resolveClaudeSessionLog,
   type RenderedMessage,
 } from "./session-log";
 import { findCodexRolloutByThreadId, getLatestCodexPlan, getRecentCodexMessages } from "./codex-session";
@@ -1448,19 +1447,8 @@ if (args[0] === "sessions") {
   } else {
     // Claude Code path: resolve session log
     //
-    // Strategy (most precise → least precise):
-    // 1. Ancestor-PID session metadata: walk up the process tree checking
-    //    ~/.claude/sessions/<pid>.json at each hop. When invoked from a slash
-    //    command's `!` bang, the direct parent is a bash subshell — Claude's
-    //    session file is a few hops up. Deterministic when it matches.
-    // 2. Cwd-scan of session metadata: read every ~/.claude/sessions/*.json,
-    //    filter by cwd, pick the most recent startedAt. Better than mtime
-    //    guessing because it uses session-level metadata.
-    // 3. CWD slug match (mtime-based): legacy behavior — picks the most
-    //    recently modified jsonl in the project dir. Fragile when multiple
-    //    sessions exist for the same project.
-    // 4. Ancestor directory walk: handles the case where the user `cd`'d
-    //    deeper into a subdirectory after session start.
+    // Prefer precise session metadata. Heuristic cwd/ancestor fallbacks are
+    // only safe when no metadata identifies the invoking session.
 
     if (process.env.PLANNOTATOR_DEBUG) {
       console.error(`[DEBUG] Project root: ${projectRoot}`);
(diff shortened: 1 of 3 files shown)


rNoz:
@backnotprop I hope you like it :)

rNoz:
@backnotprop This PR is mergeable/clean with all checks passing and no review decision recorded. It is ready for maintainer review; please let me know if you want a different shape or a focused reproduction.

rNoz:
@backnotprop Pinging, rebased again and passing. Waiting to be merged.

backnotprop:
Merged, thank you @rNoz, and sorry it sat so long. Resolving the transcript by session id is the right fix for #985. Before merging we added a follow-up on top of your commit, leaving yours untouched, to keep two cases that work on main: a fresh session in the same directory with no transcript yet is now skipped instead of failing, and one malformed `sessions/*.json` no longer turns off `/clear` detection. A file that stays unparsable after a short retry keeps the session's own transcript, so it can never pick another session's. The errors also say more when a session is found but has no transcript. Ships in the next release.

PR #1417, opened 2026-08-30
marketing: Herdr Annotate landing page at /tui-annotate

backnotprop:
Adds `apps/marketing/public/tui-annotate/`: a self-contained landing page for Herdr Annotate at https://plannotator.ai/tui-annotate/, served verbatim from `public/` (no site layout or global CSS), plus its images (~1 MB). Follows the site conventions: canonical link, OG tags, `/favicon.png`, no analytics.

**Videos are not in this PR.** The two demo mp4s (41 MB and 15 MB) are kept out of git and out of the `--delete` S3 sync. The page loads them from `https://media.plannotator.ai/tui-annotate/{full,lite}-demo.mp4`; until that host exists the video frame shows the poster with a "Watch on X" link to the announcement posts, so the page is complete either way. Files are staged at `~/oss/herdr/landing/media/tui-annotate/` for upload once the media host is approved.

Preview: https://claude.ai/code/artifact/2889a98c-83b7-403a-bbbf-b0c1502fe649

https://claude.ai/code/session_01L232F8ev8RX6Jwd7PepsUz


backnotprop:
Reviewed this as untrusted third-party HTML since public/ is served verbatim. Security is clean: the single inline script is an IIFE with no fetch, XHR, beacon, service worker, iframe, form, meta refresh, window.open, eval, storage, cookie, or URL parsing, and a live browser check confirmed zero inline event handlers and exactly two external hosts, media.plannotator.ai for the video and Google Fonts, which Base.astro already loads on every page. The no-analytics claim holds, both SVGs are inert, scope is limited to apps/marketing/public/tui-annotate/, and the payload is 1,106,013 bytes as stated.

The video fallback was verified headlessly with the media host unresolvable: networkState reports NETWORK_NO_SOURCE, the poster stays up, and the Watch on X overlay appears with the correct per-tab link, so merging before the media host exists is safe as claimed. The networkState check is the right mechanism here since this failure mode never sets media.error.

One real bug, fixed directly on the branch in b2e0ea3d: the asset URLs were relative, and production CloudFront serves the slashless /tui-annotate with a 200 rather than redirecting, so every image resolved against the site root and 403d, taking out the hero banner, all three stills, and the video poster. They are now root-anchored to /tui-annotate/assets/, which works on both URL forms.

Optional cleanups left for a follow-up: assets/install-full.svg and install-lite.svg are unreferenced, the native video controls still render under the fallback overlay, and the page requests its own Google Fonts URL with different weights than the site so it does not share the cached stylesheet.

PR #1418, opened 2026-08-30
feat(ui): link `#Heading` section references in annotation text

JunyeongChoi0:
## The problem

An annotation that refers to another section in prose — "this contradicts 3.1" — hands the reader back the cross-referencing work the annotation surface exists to remove. They have to hold the comment in their head, scroll off to find 3.1, and match the two up by eye. Comment text renders as a bare string (`<p>{annotation.text}</p>`), so there was no way to close that gap.

This is most visible on `GLOBAL_COMMENT`, which by definition is anchored to no phrase, and on any comment whose point is that two sections disagree.

## What this does

Writing `#` followed by a heading's exact text in a comment renders it as a link that scrolls the document to that heading.

```
#3.1 Single source of truth contradicts this.
```

**Resolution is exact match against the document's own headings.** That is the whole design: there is no grammar of section numbers to define, so `3.1`, `A.`, `IV.` and an unnumbered `## Authentication` are all just heading text, and nothing has to be taught about numbering schemes that vary per document.

It is also what keeps the common non-references inert, without a denylist:

| Input | Result |
|---|---|
| `#fff`, `#123` | plain — needs a heading literally named `fff` / `123` |
| `C#` | plain — a `#` welded to a word is part of that word |
| `# Heading` | plain — space after `#` is markdown heading intent |
| `` `#Overview` `` | plain — backtick spans are never scanned |
| duplicate heading title | plain — a reference that could mean either is not a reference |

Longest match wins, so a document with both `3.1` and `3.1 Scope` resolves `#3.1 Scope` to the longer one. Trailing text stays outside the link, which is what makes agglutinative languages work without tokenizing (`#3.1 단일 진실 원천의 전제` → the `의` is not part of the link).

Full heading text is deliberate over a bare number: **if the link ever fails to resolve, the sentence still reads.** `3.1` alone does not.

## Host compatibility

Additive, and inert unless opted into:

- **`AnnotationPanel` `onNavigateAnchor?: (hash: string) => void`** — the panel parses only when this AND a non-empty heading index are both present. A host passing nothing renders the same single `<p>` as before. The panel already received `blocks`, so the index costs the host nothing.
- **`ViewerHandle.scrollToAnchor(hash) => boolean`** — exposes the callback `Viewer` already had internally for in-document anchor links. Required rather than optional on purpose: the interface is implemented only inside this package (`Viewer`, `HtmlViewer`) and consumed everywhere else through a ref, so requiring it costs no host anything while forcing both implementations to stay in step. `HtmlViewer` implements it as `() => false` — raw HTML pages have no markdown heading blocks to resolve against.
- **`utils/sectionRefs`** — `buildSectionRefIndex` / `parseSectionRefs`, both pure and DOM-free.

`CodeAnnotationCard` is untouched: review annotations address files and lines, and a review surface has no heading blocks.

Documented in `packages/ui/README.md` and `packages/ui/HANDOFF.md` following the existing seam convention. **No version bump** — I left that to the maintainer; nothing under `packages/core` changed, so it should be a ui-only publish.

## Verification

Local run of the CI steps:

| Step | Result |
|---|---|
| `bun run check:release-version` | pass (`v0.27.9`) |
| `bash apps/pi-extension/vendor.sh` | pass, no extra files |
| `bun run typecheck` (10 tsconfigs, incl. `strict-consumer`) | pass |
| `bun test` | **4038 pass / 0 fail** across 427 files |
| `DOM_TESTS=1` seam batch | see note below |

New unit tests (`packages/ui/utils/sectionRefs.test.ts`, 10 cases) pin every row of the table above plus the longest-match and trailing-text cases.

Verified in a real browser on an annotate session, not just in tests:

- All branches render correctly — resolved references as links, `#없는섹션` / `#fff` as plain text, trailing particle outside the link.
- On a long document, clicking a reference to a far section moved the target heading's `getBoundingClientRect().top` from **2781px → 417px**, with the section visible on screen afterwards.
- The click does not also select the annotation card (the link stops propagation), so navigating never doubles as a scroll to the annotation's own highlight.

### Pre-existing test failure, unrelated to this change

`packages/ui/components/Viewer.consumer.test.tsx` → "readOnly fenced code never opens composers or mutates the rendered code" fails under `DOM_TESTS=1`. It asserts `code.innerHTML` equals the plain source, but Shiki has highlighted it into token spans by then.

I confirmed this is pre-existing by stashing this branch and running the same test on a clean tree — it fails identically there. Not touched here.


JunyeongChoi0:
Closing — reopening from the correct account.

Issue #541, opened 2026-04-11
macOS: binary is unsigned, killed with SIGKILL (exit 137) on macOS Sequoia


aviadshiber:
## Problem

The `plannotator` binary installed via `install.sh` is **not code-signed**, causing macOS Sequoia (15.x) to kill it immediately with SIGKILL (exit code 137).

## Steps to Reproduce

```bash
~/.local/bin/plannotator --version
# exits with code 137, no output
```

## Diagnostics

```
$ codesign -d --verbose ~/.local/bin/plannotator
/Users/.../plannotator: code object is not signed at all

$ codesign -v ~/.local/bin/plannotator
/Users/.../plannotator: invalid or unsupported format for signature
In architecture: arm64
```

The binary is a valid `Mach-O 64-bit executable arm64` (~84 MB), installed on 2026-04-11 via the official install script.

## Environment

- macOS Sequoia 25.3.0 (Darwin 25.3.0), Apple Silicon (arm64)
- Binary: `~/.local/bin/plannotator` — 83878720 bytes
- SHA256: `5ebc6351a715c84b41cac5d2546aad192bfb7875da2cf8292fd827ccb5743406`
- Plugin version: 0.14.3

## Workaround

Applying an ad-hoc signature manually fixes it:

```bash
codesign --sign - --force ~/.local/bin/plannotator
```

## Expected Behavior

The install script should apply at least an **ad-hoc signature** (`codesign --sign - --force`) post-download, or the release binary should be properly signed. This is required for macOS Sequoia which enforces stricter unsigned binary restrictions.

aviadshiber:
**Update:** Confirmed this is a v0.17.8-specific regression. v0.17.7 ships with an ad-hoc linker signature (`flags=adhoc,linker-signed`) and works correctly. v0.17.8 has no signature at all.

Workaround: downgrade with `curl -fsSL https://plannotator.ai/install.sh | bash -s -- --version v0.17.7`

backnotprop:
Tracing this, possibly related: https://github.com/oven-sh/bun/issues/29120

backnotprop:
patch is in release pipeline - building now:
`0.17.9`

backnotprop:
Fixed in [v0.17.9](https://github.com/backnotprop/plannotator/releases/tag/v0.17.9) — pinned the build to Bun 1.3.11 to avoid the 1.3.12 regression. Thanks for the detailed report @aviadshiber.

backnotprop:
[...]

PR #1420, opened 2026-08-30
docs: complete the local compile command in AGENTS.md

centraldogma99:
The local compile command in `AGENTS.md` (`## Build`) drops both flags the release workflow uses, so a binary built by following the docs differs from a released one.

Source of truth: `.github/workflows/release.yml:111` compiles with `--no-compile-autoload-bunfig` and `--define "__CLI_VERSION__=\"$VERSION\""`.

Two concrete failure modes:

- **No version identity.** Without the `__CLI_VERSION__` define, `plannotator --version` takes the fallback in `apps/hook/server/cli.ts:128` and prints `plannotator dev` — a locally built binary can't be told apart from any other. Verified locally: the documented command prints `plannotator dev`; with the define it prints `plannotator 0.0.0-dev`.
- **SIGKILL on macOS with bun 1.3.12.** The cross-compile signing regression tracked in #541 loses the linker signature; the binary is killed on launch with no error and `codesign -dv` reports it as unsigned. This is why `release.yml` pins `bun-version: 1.3.14` and keeps the macOS smoke leg (`release.yml:174-175`). Hit for real; upgrading bun resolved it. The docs said nothing about the floor.

Docs-only change: the corrected command with an obviously-local `0.0.0-dev` placeholder, plus one sentence naming the flags' source and the macOS bun floor. No restructuring.

`bun run check:release-version` passes.

Per `CONTRIBUTING.md` this is a fork → PR contribution under the repo's dual Apache-2.0/MIT licensing.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -797,8 +797,11 @@ Running only `build:hook` after review-editor changes will copy stale HTML files
 
 ```bash
 bun run --cwd apps/review build && bun run build:hook && \
-  bun build apps/hook/server/index.ts --compile --outfile ~/.local/bin/plannotator
+  bun build apps/hook/server/index.ts --compile --no-compile-autoload-bunfig \
+  --define "__CLI_VERSION__=\"0.0.0-dev\"" --outfile ~/.local/bin/plannotator
 ```
+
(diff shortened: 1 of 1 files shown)


backnotprop:
Merged, thanks @centraldogma99. Every claim in the new text checked out against release.yml and cli.ts. Ships in the next release.

PR #943, opened 2026-06-19
Add annotation undo/redo capability 

claudiolaas:
(done completely with GPT5.5)
 
 ## Summary
  - Add bounded undo/redo support for annotation actions, code annotations, attachments, checkbox toggles, and identity
  changes.
  - Wire history through linked-doc/message navigation, draft restore, direct markdown edits, submit/exit flows, and
  shortcut handling.
  - Add focused undo/redo tests plus root TypeScript dependency so `bun run typecheck` works from a fresh install.

  ## Implementation
  The core is a small command-stack controller in `useAnnotationHistory`, with each undoable action carrying enough state
  to replay its inverse and redo. `App.tsx` keeps ownership of lifecycle boundaries, while annotation mutations are routed
  through `useAnnotationCommands` so history recording stays centralized. Shortcuts use the existing shortcut registry,
  and linked-doc/message switching serializes or restores the relevant history stack with the rest of the surface state.

  ## Verification
  - `bun test`
  - `bun run typecheck`
  - `git diff --check`

PR #538, opened 2026-04-10
feat(annotate): add Cmd+Z undo for annotations

salmonumbrella:
## Summary
- Adds Cmd/Ctrl+Z support to undo the most recent annotation that has no user-written content
- Dismisses transient UI state (empty comment popovers, quick-label pickers, toolbars) before removing annotations
- Skips annotations with comment text, images, quick labels, or external sources

## Test plan
- [x] Unit tests for `isUndoRemovableAnnotation` and `findLastUndoRemovableAnnotation`
- [ ] Verify Cmd+Z removes empty deletion annotations in order
- [ ] Verify Cmd+Z does not remove annotations with comment text or images
- [ ] Verify Cmd+Z dismisses open empty popovers before touching annotations
- [ ] Verify Cmd+Z is no-op when typing in text fields (inputs, textareas)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1426, opened 2026-08-31
feat: add bounded annotation undo and redo

### LINKED ISSUE #828: [Feature Request] Undo/redo hotkeys for annotations and editor actions
## Problem
There's no general undo/redo (i.e. for annotations and labels). Currently:
- The image annotator has **undo only** (Cmd+Z + a toolbar button), no redo — `packages/ui/components/ImageAnnotator/`.
- Plan/code annotation create / delete / edit has **no undo/redo at all** — `packages/editor/App.tsx` mutates state directly, with no Cmd+Z handling.

_So if I delete an annotation by mistake, or want to step back through a series of edits, I can't._

## Proposal
Unified undo/redo across editor actions with standard hotkeys:
- Cmd/Ctrl+Z = undo, Cmd/Ctrl+Shift+Z (and Ctrl+Y) = redo
- Covers annotation create / delete / edit (and ideally reuses the existing image-annotator undo stack so redo lands there too)
- Optional toolbar buttons mirroring the shortcuts

backnotprop:
## Summary

- add a shared, bounded (50-action) undo/redo command history for human annotation mutations
- cover add, edit, and delete across plan/document annotations, linked code-file comments, code-review file/line comments, conventional labels, suggestions, PR-description annotations, and PR-comment annotations
- add proper Image Annotator past/future stroke history with Undo and Redo controls, redo invalidation, and Clear resetting both branches
- preserve deleted item positions, selection/highlight restoration, checkbox state and annotation ordering, reply relationships, draft/export data, and existing thread grouping

Closes #828.

## Prior work and credit

This supersedes, but does not close, #538 and #943.

- Credit to Rusty Shackleford (@salmonumbrella) for #538, which established the original Cmd/Ctrl+Z UX and highlighted native-input and transient-tool ownership concerns. This PR uses that work as historical context rather than importing its deletion-only implementation.
- Credit to Claudio Laas (@claudiolaas) for #943, whose bounded command-stack direction, lifecycle analysis, and test scenarios informed the design. This PR independently ports those ideas onto current `main`, narrows history to human annotation mutations, uses shared pure primitives with small surface adapters, and extends coverage to the current review UI and Image Annotator.

Neither stale PR was merged or rebased wholesale.

## Design

`undoHistory.ts` owns pure bounded stacks, context isolation, reversible collection mutations, and shortcut-owner guards. `useUndoHistory` provides a small stable React adapter. The plan/document and review apps record compact commands containing the exact before/after item data, collection index, and relevant selection state instead of snapshotting the entire app.

New mutations after undo discard the redo branch. Compound review suggestions replay as one action. Checkbox actions preserve both visual overrides and generated annotation positions. External/source-tagged annotations never enter human history.

## Shortcut ownership

- Undo: `Mod+Z`
- Redo: `Mod+Shift+Z` and `Mod+Y`
- shortcuts only prevent the browser default when a history action is actually available and handled
- native history remains owned by inputs, textareas, `contenteditable`, role=textbox controls, CodeMirror, and shadow-DOM descendants
- dialogs, comment popovers, focused source editors, pending image paste, edit mode, and other active tools suppress document/review history
- the Image Annotator owns the shared undo/redo bindings while open
- review's bare `Z` collapse binding and `Mod+Shift+Y` copy binding remain distinct

## Lifecycle reset/rebase boundaries

Plan/document history is cleared on document/message/live-page context changes, linked-document and file navigation, draft restoration, archive changes, HTML refresh/content replacement, shared/external/WebMCP mutations, direct markdown edits and anchor remapping, identity rewrites, and submission.

Review history is cleared on snapshot/PR/diff identity changes, raw-patch replacement, draft restoration, external annotation edits/deletes, identity rewrites, and submission. Image stroke history is local to one annotator session and resets on Clear or remount.

## Verification

- `bun test packages/ui/utils/undoHistory.test.ts packages/ui/components/ImageAnnotator/strokeHistory.test.ts packages/ui/shortcuts.test.ts` — 23 passed
- `DOM_TESTS=1 bun test packages/ui/utils/undoHistory.dom.test.ts` — 4 passed
- `bun test packages/ui/hooks/useLinkedDoc.diffBaseline.test.ts` — 5 passed
- `DOM_TESTS=1 bun test packages/ui/hooks/useLinkedDoc.test.tsx` — 5 passed
- `DOM_TESTS=1 bun test packages/ui/codeAnnotationDraftPersistence.test.tsx packages/ui/hooks/useAnnotationDraft.seam.test.tsx` — 7 passed (one pre-existing React `act(...)` warning)
- `bun run typecheck`
- `bun run build:hook`
- `bun run --cwd apps/opencode-plugin build`
- `git diff --check origin/main...HEAD`

### Browser smoke evidence

- plan annotations: add/edit/delete undo and redo via both redo bindings
- replies: external WebMCP reply rendered with its parent and survived subsequent human-history undo, confirming exclusion/rebase behavior
- real annotate session: linked code-file comment add/undo/redo; linked-document isolation; draft restore established a fresh baseline; CodeMirror kept native undo; checklist toggle undo/redo preserved its generated annotation
- code review: file comment, line comment, suggested-code annotation, and `issue` conventional-label comment each undid/redid correctly; active composers did not mutate application history
- Image Annotator: the visible drop zone accepted a real PNG; rendered canvas hashes changed after drawing, returned to the exact baseline after undo, returned to the exact stroke after redo, and invalidated redo after a replacement stroke; Clear disabled both stacks and Save produced an attachment

## Residual limitations

- history is intentionally session-local and capped at 50 human annotation actions per active document/review context; it is cleared rather than serialized across lifecycle boundaries
- external/WebMCP/VS Code/agent-authored mutations are not undoable; safe entry points clear/rebase human history instead
- attachment add/remove and identity rewrites are not standalone undo actions; Image Annotator history covers strokes before the image is saved
- Image Annotator Clear is intentionally destructive for stroke history and cannot itself be undone
- document/review undo and redo are keyboard actions exposed in shortcut help; only the Image Annotator adds toolbar buttons


backnotprop:
Reviewed the whole surface: core primitives, both App adapters, shortcut layering, draft interplay, external/WebMCP paths, and the published-package edges. The design is sound and it is conservative in the right places: source-tagged items never enter history, lifecycle boundaries clear aggressively, native text history stays with inputs/textareas/contenteditable/CodeMirror, and there are no server changes. Draft persistence, external SSE annotations, WebMCP ownership, and inReplyTo threading all hold up under undo/redo; I checked each explicitly.

One blocker.

1. High. Code-block annotation marks get permanently tombstoned by any history repaint. packages/editor/App.tsx:729-745 calls viewerRef.current.removeHighlight(id) and then applySharedAnnotations([annotation]) for redo-of-add, undo-of-delete, and both directions of edit, which are exactly the cases where the annotation STAYS in the list. Viewer.removeHighlight (packages/ui/components/Viewer.tsx:687-690) adds the id to removedAnnotationIdsRef, and the only retirement path (Viewer.tsx:405-407) deletes an id only when it is absent from the annotations list. So the tombstone sticks for the life of the Viewer mount, and the swap listener's !removedAnnotationIdsRef.current.has(a.id) filter (Viewer.tsx:442) then refuses to repaint that fence: the next palette change, dark/light toggle, or late async grammar attach silently wipes the mark and it never comes back. I reproduced it with a scratch DOM test against the real Viewer plus ThemeProvider (mark present after the repaint, null after setColorTheme). On main the only removeHighlight caller drops the annotation from state immediately, so the tombstone always retired; this PR introduces the first remove-then-keep caller. Suggested fix: skip the pre-remove when the annotation is staying or returning (applyAnnotations is already skip-if-present), or retire tombstones for ids that have a live [data-bind-id] in the DOM. Please also add a DOM test for this exact scenario: annotate a fence, history-repaint it, change theme, assert the mark survives.

Medium.

2. Undo/redo is effectively unreachable on raw-HTML and live-app annotate surfaces. Both listeners are on the parent window and focus lives inside the iframe there. The bridge says so itself at packages/ui/components/html-viewer/bridge-script.ts:3085-3091, where Mod+Shift+A is forwarded for exactly that reason. Mod+Z is not forwarded. Worth either forwarding it or adding this to Residual limitations.

3. Image Annotator tool digits now fire while typing the attachment name. The deleted handler returned early on every key when target.tagName === 'INPUT'. The replacement guards undo, redo and save with keyboardTargetIsInput but not penTool/arrowTool/circleTool (packages/ui/components/ImageAnnotator/index.tsx:151-165), so naming an image "v2" silently switches the tool.

4. Review Edit Mode blind spot. canHandleReviewHistoryShortcut (packages/review-editor/App.tsx:3587-3591) gates on ten dialogs but not on an active edit session, and hasActiveHistoryOverlay only catches .cm-editor.cm-focused. With an open but unfocused buffer, Mod+Z undoes an annotation instead of the edit. The plan editor has the equivalent gate via isEditingMarkdown; review does not.

Low.

5. useUndoHistory calls render() in record/undo/redo/clear/clearAll, but nothing reads canUndo/canRedo during render (only inside keydown when closures at packages/editor/App.tsx:1478,1482 and packages/review-editor/App.tsx:3619,3623). That is an extra full re-render of the two largest components in the repo on every annotation mutation, for no observable benefit.

6. Per-context isolation is built and then discarded: clear(context) is never called, both Apps use clearAll(), including on historyContext change. The Map can never hold more than one live context, and navigating to a sibling doc and back loses that doc's history even though the design supports keeping it. Either use clear() or drop the Map.

7. recordHistory as a trailing positional boolean on handleDeleteAnnotation / handleEditAnnotation (App.tsx:4087,4140) is safe today, but those same functions are passed straight through as onDelete/onEdit props (App.tsx:5367-5368). A future arr.forEach(onDelete) would pass an index into the flag. A separate silent variant would be safer.

8. Small behavior change: the checkbox branch of handleDeleteAnnotation now returns from inside the if (ann) guard (App.tsx:4106-4128), where main fell through to removeAnnotation(id). Practically unreachable, noting it for completeness.

Published package: mostly additive and a host that passes nothing is unchanged. Two non-additive edges, both under components/ImageAnnotator, which is a published export (and ./components/* matches subpaths, so Toolbar and types are reachable): ToolbarProps gains required canRedo/onRedo, and AnnotatorState gains a required futureStrokes. A host rendering Toolbar directly or building an AnnotatorState literal breaks. Making both optional (state.futureStrokes ?? [] at read sites) keeps it fully additive.

Tests: the pure-unit files are good and guard real invariants. But 12 tests over 1315 added lines leaves the 678 lines of adapter code inside the two Apps at zero coverage, including the repaint path that is broken above. The dom test at undoHistory.dom.test.ts:64-76 hardcodes when: () => false for the document scope rather than exercising the real arbitration (hasActiveHistoryOverlay finding the annotator's data-popover-layer), so it asserts the outcome without testing the mechanism.

Verification I ran: 23 pass on the three unit files, 4 pass on the dom file, 1833 pass / 1 skip / 0 fail across packages/ui, packages/editor and packages/review-editor with DOM_TESTS=1, bun run typecheck clean including the strict-consumer config, and apps/review build ok at 17,579.53 kB vs 17,571.62 kB on main (+7.9 kB raw, +2.79 kB gzip). All test runs sandboxed under a temp PLANNOTATOR_DATA_DIR. No new dependencies and no lockfile changes.

Also worth a CLAUDE.md pass: the shortcut section still says scopes live in plan-review/ or code-review/ (history.shortcuts.ts is deliberately shared and sits at the top level, which reads correct but undocumented), and nothing describes the history system, its lifecycle resets, or the shortcut ownership rules.

Recommendation: take with changes. Fix 1 before merge, fix or document 2 and 3, the rest can follow.

backnotprop:
Resolved every item from the review in `706a51a6`:

1. **High — retained code-block highlights:** reproduced against the real `Viewer` + `ThemeProvider`. History replay no longer calls `removeHighlight()` when an annotation remains/returns; `syncHistoryHighlight()` applies visible items and reserves removal for genuinely absent items. This preserves the existing tombstone/deletion-race semantics and collection ordering/thread relationships. Added the requested DOM regression: history repaint, palette swap, mark still present.
2. **Raw HTML/live iframe undo/redo:** added an authenticated, host-opt-in bridge for Mod+Z / Mod+Shift+Z / Mod+Y. Embedded inputs, textareas, contenteditable, role=textbox, and CodeMirror keep native undo. The bridge protocol is bumped to v2 so stale published assets are diagnosed.
3. **Image Annotator digits:** tool shortcuts now use the same input-owner exclusion as undo/redo/save. Added a DOM regression proving typing `2` in the attachment name does not switch tools.
4. **Review Edit Mode:** the active edit session is reported from `AllFilesCodeView` through review state and gates annotation history even when the editor is unfocused. Live QA used a real repository/full-content session and verified Mod+Z from the navigation control left both the contenteditable session and annotation intact.
5. **Unnecessary rerenders:** removed the reducer/render bump; stack mutations are imperative and do not rerender either App.
6. **Discarded contextual Map:** removed it. The hook owns one active bounded stack and starts a fresh baseline synchronously on context change; no old document/review history is cached.
7. **Trailing boolean:** replaced positional `recordHistory` flags with explicit normal and without-history entrypoints.
8. **Checkbox fallback:** restored the generic `removeAnnotation(id)` fallback for a stale checkbox row.
9. **Published-package additivity:** `Toolbar` redo props and `AnnotatorState.futureStrokes` are optional, with legacy-state fallbacks and coverage.
10. **DOM arbitration/tests/docs:** the Image overlay test now exercises `hasActiveHistoryOverlay` instead of hardcoding false; lifecycle/no-render/context-reset, external exclusion, replies, and bridge auth/native-owner cases are covered. Updated the repository shortcut/history ownership and lifecycle documentation.

Verification after the final changes:
- Focused DOM/history/bridge/image matrix: **71 pass, 0 fail**
- `DOM_TESTS=1 bun test packages/ui packages/editor packages/review-editor`: **1,839 pass, 0 fail**, 7,310 assertions / 225 files
- Root typecheck (including strict consumer): pass
- `packages/ui` typecheck: pass
- Review, hook, and OpenCode production builds: pass
- Generated bridge-asset consistency and `git diff --check`: pass

Browser evidence:
- Plan code-fence annotation undo/redo survived a dark→light palette swap (one live mark remained).
- Raw HTML iframe Mod+Z/redo reached parent history; Mod+Z in the iframe input stayed native and preserved the annotation.
- Image file-picker path worked; name digits preserved Pen; stroke undo/redo and Clear button states were correct.
- Live repository review Edit Mode (real contenteditable): unfocused Mod+Z did not undo the annotation or end the edit session.
- No console errors on the plan, raw HTML, image, or live-review runs. (The standalone demo emitted its pre-existing malformed demo-patch parser noise, so the authoritative Edit Mode check used the live repository server.)

There were no formal review objects or inline review threads on the PR API; this issue comment is the sole posted review thread, so this reply covers each item.

backnotprop:
Follow-up: the first refreshed CI run correctly caught the portable guide viewer pin drifting after the bundled UI changes. The viewer stayed within budget (395.1 KB gzip JS / 50.8 KB CSS), and `88556dd3` regenerates the prescribed checked-in manifest. Local `build:viewer`, `check:budgets`, `check:manifest`, core typecheck, and diff check now pass. Watching the new CI run to completion.

backnotprop:
Addressed the review in `a2e06c81` after reconciling the original follow-up (`706a51a6`) and the later manifest refresh (`88556dd3`) without force-pushing. Item-by-item:

- **High — fenced-code highlight tombstone:** history repaint now uses `syncHistoryHighlight()`: retained annotations are repainted without first tombstoning the shared mark, while removed annotations still use `removeHighlight()`. The regression is a real DOM `Viewer` test that creates an annotation by hovering/clicking a fence, history-repaints it, changes the theme palette, and proves the mark remains attached with the new token color.
- **Image Annotator digit/tool input:** the digit bindings now ignore the image-name input while still switching tools from the window. A mounted DOM shortcut regression covers both paths.
- **Published Image Annotator compatibility:** the public `AnnotatorState`/`DEFAULT_STATE` shape remains additive-compatible; redo state is internal. `Toolbar` redo props are optional and the redo button is omitted for legacy direct consumers. A strict-consumer fixture compiles the old state literal and old toolbar JSX.
- **Review Edit Mode shortcut ownership:** active editing now owns history through light-DOM `data-history-owner="edit-session"` markers on the all-files root and HUD. The global history shortcut gate checks that marker, so Mod+Z is left to the embedded editor. DOM/lifecycle coverage asserts marker attachment and cleanup.
- **History lifecycle and rerenders:** `useUndoHistory` remains imperative (record/undo do not render), uses one active bounded stack, and synchronously clears history when the context changes. DOM coverage proves both no-rerender behavior and A → B → A isolation.
- **Safer silent updates:** replaced trailing record-history booleans with explicit `'record' | 'silent'` paths and named silent wrappers. Checkbox override cleanup still falls through to `removeAnnotation()` when its tracked id is stale.
- **Shortcut arbitration:** tests mount the real history and Image Annotator listeners together and verify the active overlay wins. Registry assertions cover Mod+Z, Shift+Z, Mod+Y, bare Z, and Mod+Shift+Y claimants.
- **Raw HTML/live iframe:** I did not retain the proposed forwarding because it enabled `preventDefault()` based on handler presence rather than actual `canUndo`/`canRedo`, which could swallow iframe shortcuts with no parent history action. Reliable availability synchronization would broaden this patch, so the current no-forwarding limitation is documented explicitly in `AGENTS.md`/`CLAUDE.md`.
- **Docs/CI:** shortcut ownership, the 50-action history lifecycle, context reset semantics, and the iframe limitation are documented. CI explicitly registers the new Image Annotator and undo-history DOM suites.

Validation on the integrated head:

- focused pure tests: **22 passed**
- focused DOM/protocol tests: **49 passed**
- lifecycle tests: **19 passed**
- full registered DOM matrix: **1,835 passed, 1 skipped, 0 failed** (225 files)
- root/UI typechecks including the strict consumer: **passed**
- production builds: review, hook, and OpenCode: **passed**
- `git diff --check origin/main..HEAD`: **clean**
- live browser smoke previously confirmed both Edit Mode owner markers and that Mod+Z remained unconsumed; the post-integration in-app rerun could not reliably surface the virtualized settings Edit control, so I timeboxed that fixture issue and relied on the integrated deterministic DOM/lifecycle coverage for the final gate.

No dependency or lockfile changes, no final `DiffViewer.tsx`/Pierre integration changes, and no merge performed. Fresh CI is running on `a2e06c81`.

backnotprop:
Follow-up: the first CI run exposed a stale pinned guides.show viewer JavaScript hash after the integrated UI changes. I regenerated it only with the canonical `build:viewer` plus `sync:manifest` flow, verified `sync:manifest --check`, and pushed the one-file generated refresh as `448e9183`. Fresh CI on that head is terminal green: 24 successful, 4 expected release-only skips, 0 failures, and 0 pending. The PR remains unmerged.

PR #1437, opened 2026-09-01
fix(doc): deny symlink escapes on /api/doc reads

bendrucker:
Consolidates file handling for the review UIs into one shared module. Fixes symlinks breaking project folder containment and reading arbitrary paths on disk.

## Approach

Resolving a path and reading it are now separate steps, so the permission check can run between them.

Resolving may `stat` a path to see whether it exists. It now checks the realpath and requires that to be inside the project in addition to the input path.

`realpath` fails on a path that does not exist, and the check still has to judge those, since an outside path must be refused whether or not the file is there. So it resolves the deepest part of the path that does exist and re-attaches the rest.

That catches a missing file under a symlinked folder without turning an ordinary missing file into a denial. Anything unexpected back from `realpath` denies the request.

## Behavior Changes

- A symlink pointing outside the project returns 403, on every path through the endpoint, on both servers.
- HTML over 2MB returns 413. It used to be served in full when you did not pass a `base`.
- Point the project root at a symlink and requests now work whether you name the link or the real folder. Both sides have to be normalized, because macOS `tmpdir()` is itself a symlink and normalizing one side breaks every test on a Mac.
- A `base` directory that reaches outside the project through a symlink is also refused.
- A file that exists but cannot be opened returns 500 instead of 404, because the existence check now happens before the read.

## Tests

Each way a symlink can escape has its own test on both servers, since a single test covering one of them passes against a fix that missed a branch. Also covers a project root named through a symlink, an outside path that does not exist, and the HTML size limit.


backnotprop:
Heads up while you are actively on this: the branch is conflicting with main, which moved substantially today (a large header rework in packages/editor and packages/review-editor, plus new code-nav endpoints in both server runtimes), and the conflict is blocking CI from running on your latest commits. A rebase onto current main will unblock checks. The doc-read path you are hardening was not restructured, so the rebase should be mostly mechanical. Once it is green we will get you a full review promptly; this fix is wanted.

bendrucker:
No problem will rebase now!

backnotprop:
Owe you an apology and one more rebase, and this one is entirely on us.

Yesterday a fix commit on main was made with a blanket `git add` and accidentally swept in ~150 of our untracked local scratch files (design notes, release-note drafts, QA docs). We have since cleaned it up properly, which required rewriting recent main history and force-pushing — so the commits your branch is currently based on no longer exist on main.

Your PR still renders correctly, but merging it as-is would reintroduce the pre-cleanup commits into main's history. One more `git rebase origin/main` (after a fresh fetch) onto the rewritten main will fix it — your two commits should apply cleanly, nothing in the areas you touched changed.

Sorry for asking twice in two days. The review itself is next on our list once the branch is rebased — the containment work looks solid.

bendrucker:
Not a problem, updated!

backnotprop:
CI status on the rebased branch, with some triage to save you time:

- Security Scanning: green.
- Test: failed twice (original + rerun) on the same four tests, all in `packages/server/annotate.test.ts`: the three "durable submit records (#678)" tests and "an unwritable history directory degrades to a plain render". The failure is `readdirSync(recordDir)` finding no record where the test expects one.

What we've ruled out:
- Not our known flake (#1464): that one passes on rerun; this failed twice deterministically.
- Not main: main's tip is green on the same runners.
- Not macOS: your branch passes the full suite locally on macOS, both standalone and merged with current main (58/0 both ways).
- The Pi mirror of the same unwritable-directory test passes in the same CI run — only the Bun-server variants fail.

So it's Linux-runner-specific and introduced by the branch. Given your own note about having to realpath-normalize both sides because macOS `tmpdir()` is a symlink, our best guess is the inverse case: on Linux (no symlinked tmp) the path your resolver hands back for a served folder file may no longer match the path the submission-record machinery keys its directory by (the record dir slug hashes the file path), so the record lands somewhere the test - and real recovery - doesn't look. Worth checking how `submissionsDir` interacts with your normalized resolution on a non-symlinked tmp.

Happy to dig further if you get stuck - and thanks again for the quick rebase turnaround.

bendrucker:
Makes sense, I'll test on Docker locally this round, either directly or with [act](https://github.com/nektos/act)

bendrucker:
The four failures reproduce locally, and the diagnosis needs correcting on three counts. This is not Linux-specific, not introduced by this branch, and not the realpath/slug interaction.

No Docker needed. This fails the same four tests on macOS:

```
```

That pair fails the same way on `d749c55c`, the merge-base with main, so the bug is latent on main rather than introduced here.

The cause is `packages/shared/storage.ts:16`:

```ts
const DATA_DIR = getPlannotatorDataDir();
```

The data dir is frozen at module import. `apps/opencode-plugin/embedded.ts:33` reaches `@plannotator/server` through `await import(...)`, so `storage.ts` first loads *during* a test body in `plan-review-lifecycle.test.ts`, inside the window where `prepareEnvironment()` has `PLANNOTATOR_DATA_DIR` pointed at that file's temp dir. Storage then keeps `/tmp/plannotator-opencode-lifecycle-*` for the rest of the process. That file's `afterEach` restore is correct and cannot help, because the freeze outlives it.

Every other consumer resolves the data dir live: `draft.ts`, `config.ts`, `feedback-archive.ts`, the annotate server, and `annotate.test.ts`'s own `submissionsDir()` helper. So two of the four tests plant a blocker the server never hits, and the other two read a directory the server never wrote. After running the pair, the records are sitting under `/tmp/plannotator-opencode-lifecycle-*/history/_annotate_submission_test_.../submissions`.

Only CI sees it because Bun's test-file discovery order is filesystem-dependent, and it is not controllable by argument order. `plan-review-lifecycle.test.ts` runs at position 8 on CI against 22 locally, and `annotate.test.ts` at 154 against 67. The two test files this branch adds (`packages/shared/doc-resolve.test.ts`, `apps/pi-extension/server/reference.test.ts`) shifted that order enough to put the poisoner ahead of the victim. The Pi mirror passes in the same run because `apps/pi-extension/generated/` is a separately vendored module instance whose copy freezes at a different moment.

Both earlier hypotheses were tested directly and passed: realpath/slug drift on a non-symlinked tmp, and XDG drift with `XDG_DATA_HOME` set. Also cleared along the way: arm64 and amd64 Linux, a symlinked `TMPDIR`, the CI file order replayed as arguments, all 105 order-suspect files in one run, and 4 vCPU / 16 GB cgroup limits.

The commit pushed here resolves the data dir per call in `storage.ts`, matching the shape `draft.ts` and `feedback-archive.ts` already use. It adds a regression test that imports `storage` first and then moves `PLANNOTATOR_DATA_DIR`, which is the order that made the freeze invisible. Verified against a CI-shaped Linux container on oven/bun 1.3.14 as well as macOS.

Four other modules freeze the data dir the same way: `packages/shared/improvement-hooks.ts:24`, `packages/server/codex-review.ts:24`, `packages/server/tour/tour-review.ts:416`, `packages/server/guide/guide-review.ts:478`. None of them has a failing test proving a real effect, so widening this PR into codex/tour/guide review code seemed worse than filing a follow-up. Happy to open that issue.

backnotprop:
You are right on all three counts, and thank you for testing our hypothesis directly instead of just asserting against it - the realpath/slug guess was wrong, and the import-order mechanics you traced (dynamic import inside the lifecycle test's override window, filesystem-dependent discovery order deciding poisoner vs victim, the vendored Pi copy freezing at a different moment) explain every observation we had, including why the rerun 'broke the flake pattern.' This has been our #1464 mystery for five occurrences; consider it root-caused by you.

One coordination note: #1473 (from another contributor, opened yesterday) fixes the same storage.ts freeze the same way, as part of a test-isolation PR closing #1455. So whichever of the two lands first, the other will get a small conflict on storage.ts that should resolve trivially since the changes converge. We'll sequence that on our side during review - no action needed from you.

Please do file the follow-up for the other four frozen-at-import modules (improvement-hooks, codex-review, tour-review, guide-review); keeping this PR scoped was the right call. CI runs on your new head are approved and running.

backnotprop:
Deep review complete: the containment held against eight adversarial probes written independently of your suite (traversal, nested symlink chains, missing-leaf reattach, absolute paths, exists-endpoint escapes), legitimate symlink setups all still serve, Bun/Pi parity is line-for-line, no client surface mishandles the new status codes, and neutering the realpath gate fails 18 tests - the guards are real. Full runs clean on both runtimes, typecheck green.

We'll carry two notes forward: a release-note line for the intended behavior change (in-root symlinks pointing elsewhere on the user's own disk now 403), and a small post-merge cleanup deduplicating the two storage regression tests that now coexist.

Merging. Thank you for the containment work, the patience through two rebases we caused, and for root-causing our longest-standing CI flake along the way - this PR ended up improving the project well beyond its own scope.

PR #1445, opened 2026-09-01
fix(ai): stop leaking opencode serve processes

backnotprop:
## The leak

Anyone with the `opencode` CLI on PATH was affected on every Plannotator session, on both runtimes. Server boot eagerly called the opencode provider's `fetchModels()` just to fill the Ask AI model dropdown, which spawned an `opencode serve` (or attached to an existing one) on the shared default port 4096. Cleanup only ran on the clean decision path, so a Ctrl-C, crash, or host-killed hook orphaned the child. From there it compounded: the orphan owned 4096, every later session attached to it, and each attach from a new directory created a ~130MB instance inside it that opencode never evicts. Observed in the wild as a 2.7GB `opencode serve` nobody started on purpose.

## The fix

**Lazy start (the core).** The autolaunch itself was the problem: spawning a whole agent server at boot to populate a dropdown most users never open. opencode model discovery now sits behind the same deferred provider initializer Codex uses (#1144 pattern): nothing spawns until the user activates opencode in Ask AI (`?activate=` from the model picker via `useAIProviderActivation`, or the first opencode session). Pre-activation the picker lists the provider with an empty model list, exactly like Codex — both `endpoints.ts` (`models: p.models ?? []`) and the UI (`provider.models ?? []`) already handle that state.

**Own server per process.** When it does start, it spawns with `port: 0` — the SDK parses the real URL from the child's own "listening" line, so the OS assigns a free port — and the attach path is removed entirely: we never connect to a server we didn't spawn, because we can't clean up what we don't own and its instance cache grows without bound. An explicitly configured port is still honored verbatim.

**Exit cleanup.** A `process.once("exit")` handler closes the spawned server; the CLI already routes SIGINT/SIGTERM through `process.exit`, so Ctrl-C now kills the child. Dispose removes the handler. No SIGHUP listener is installed, preserving the `nohup` contract.

## Both runtimes

`packages/server/ai-runtime.ts` and `apps/pi-extension/server/ai-runtime.ts` get the same deferral; the provider itself is shared (vendored to Pi by `vendor.sh`).

## Tests

`packages/ai/providers/opencode-sdk.test.ts` (5 tests): SDK mocked so no real server ever spawns — asserts port 0, exit-handler add/remove symmetry, the exit handler actually closing the server, explicit-port passthrough, and a source-level pin that BOTH runtimes keep opencode discovery behind `providerInitializers` instead of eager `modelDiscovery`.

`bun run typecheck` clean across all 9 projects. `bun test packages/ai/` 127 pass. The 6 failures in `packages/server/review-workspace.test.ts` and 6 in `apps/pi-extension/server.test.ts` reproduce identically on clean main (sem-sidecar/GitButler environment) and are unrelated.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -262,6 +262,8 @@ Automatic resolution is session-only and never writes a preference. Explicit per
 
 > **Codex transport note:** the `codex-sdk` provider id is a stable identifier only — it no longer uses `@openai/codex-sdk` / `codex exec`. It drives a long-lived `codex app-server` process over JSON-RPC (`packages/ai/providers/codex-app-server.ts`), which respects the user's/enterprise-managed approval policy and supports interactive Allow/Deny approvals. The id stays `codex-sdk` to preserve saved cookie preferences, the `agents.ts` mapping, and the UI reasoning-effort gate.
 
+> **OpenCode transport note:** the `opencode-sdk` provider spawns its own `opencode serve` per process on an OS-assigned port (`port: 0`) and never attaches to a server it did not spawn (an attached server can't be cleaned up by us, and opencode's per-directory instances accumulate in it without eviction). The spawned server is closed on dispose and on process exit. Model discovery is deferred behind the provider initializer (`?activate=` from the model picker, or the first opencode session) exactly like Codex — nothing spawns at server boot, so the picker lists opencode with an empty model list until first activation. Regression-pinned by `packages/ai/providers/opencode-sdk.test.ts`.
+
 ## Annotate Flow
 
 ```
--- a/apps/pi-extension/server/ai-runtime.ts
+++ b/apps/pi-extension/server/ai-runtime.ts
@@ -104,14 +104,21 @@ export async function createPiAIRuntime(options: CreatePiAIRuntimeOptions = {}):
 					type: "opencode-sdk",
 					cwd,
 				});
+				const providerId = registry.register(provider);
+				// Deferred like Codex: fetchModels spawns `opencode serve`, so it
+				// must NOT run eagerly at startup — that spawned a server on every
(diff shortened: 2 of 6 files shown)


backnotprop:
Independent adversarial review ran against this branch (verdict: merge with minor changes) and empirically confirmed the fix end to end: `opencode serve --port=0` binds an OS port and the SDK reads the real URL from stdout; plain `/api/ai/capabilities` spawns nothing while `?activate=opencode-sdk` spawns exactly one child and returns the real model list; SIGINT reaps the child through the CLI's exit routing; vendoring, docs, and mock isolation all check out. The empty-model-list UX was chased specifically: no crash, no wrong model — session create resolves models post-activation (`endpoints.ts` runs `beforeProviderSession` before model resolution), and the picker simply hides the model select until activation.

The review found two lifecycle holes, closed in 698d97b6 with tests:

1. A post-spawn failure (client construction) leaked the child and stacked exit handlers, and the handler read `this.server` late, so after a retry the first child was unreachable by any cleanup. The handler now captures its own server, and any post-spawn failure reaps child + handler.
2. `dispose()` during an in-flight spawn was undone by the completing spawn, resurrecting a disposed provider with a live child. `dispose()` now bumps a start epoch; a spawn completing past its epoch reaps itself and rejects, and the provider stays restartable.

Known residual, by design: SIGHUP still orphans the child, because the CLI deliberately installs no SIGHUP listener (`nohup plannotator review &` must survive terminal close). Under the old code that orphan was one shared server; under this fix it is one per process — but each is small, unattachable, and never grows, versus the old single orphan that every later session piled unevictable instances into. The review also found and reclaimed ~6GB of orphaned servers on the dev machine (25 from an intermediate branch state, one from the old shared-port path), which is the leak's severity in miniature.

PR #1448, opened 2026-09-01
feat(ui): add sticky header lane host seams

backnotprop:
## Summary

- add additive StickyHeaderLane visibility and sticky host props while preserving the existing defaults
- keep always-visible resting lanes interactive but chrome-free, and make non-sticky lanes avoid intersection observation and scroll normally
- preserve measured wide, tight, and narrow layouts plus hideQuickLabel forwarding
- document the public host contract and prove the published subpath through strict and packed external consumers

Hosts that disable sticky behavior should pass the same flag to Viewer.stickyActions so both measured header lanes follow one policy.

## Verification

- DOM_TESTS=1 bun test --isolate packages/ui/components/StickyHeaderLane.test.tsx (5 pass, 0 fail)
- registered DOM workflow batch (720 pass across 89 files, 0 fail)
- bun run --cwd packages/ui typecheck
- bunx tsc --noEmit -p packages/ui/tsconfig.strict-consumer.json
- bun run --cwd packages/ui smoke:package
- git diff --check

## Package availability

This PR deliberately does not bump or publish @plannotator/ui. The head commit and a tarball built from it are immediately consumable for integration testing. The first registry semver that includes the seam will be the next @plannotator/ui version assigned by the normal release process after merge; no unreleased version is invented here.


backnotprop:
Review follow-up pushed in `872d7c04`:

- **Default compatibility:** restored the incumbent chrome classes for the default stuck-only path, including while the whole bar fades out after returning to the top. Added an exact resting wrapper/bar markup contract covering the border, padding, and 180ms transition. `visibility="always"` alone now owns chrome-free rest behavior.
- **Always-visible overlay contract:** README, HANDOFF, and prop JSDoc now state that the lane remains zero-height/absolute and the host must reserve a clear header-height region so controls cannot cover or intercept document content.
- **`sticky={false}` contract:** docs now state it must be paired with `visibility="always"`; otherwise stuck-only visibility can never activate.
- **Measured action width:** intentionally retained. With stickiness disabled on both `StickyHeaderLane` and `Viewer`, the two clusters still share the lane at rest before scrolling away together, so the measured reserve prevents overlap.

Verification after the follow-up:

- `DOM_TESTS=1 bun test --isolate packages/ui/components/StickyHeaderLane.test.tsx` — 5 passed, 53 assertions
- `bun run typecheck`
- `bun run --cwd packages/ui smoke:package` — external pnpm install, Vite build, and TypeScript compile passed
- `git diff --check`

PR #1453, opened 2026-09-02
docs(skills): require gate for file approval

zhangjinzan1:
## Summary
- distinguish ordinary annotation from explicit file approval in the core skills
- require `--gate --json` when a user asks to approve a saved plan/spec/document
- clarify that `--json` alone does not enable the Approve button
- add a regression assertion for the approval-gate guidance

## Why
A Codex agent followed the installed core skill and launched a generated plan with `plannotator annotate <file> --markdown --json`. The UI correctly opened in plain annotate mode, so it showed **Close** but no **Approve**, even though the agent told the user they could approve the plan.

## Test
- `bun test apps/hook/server/plannotator-skill-reference.test.ts` (6 passed, 0 failed)
- `git diff --check`

### DIFF
--- a/apps/hook/server/plannotator-skill-reference.test.ts
+++ b/apps/hook/server/plannotator-skill-reference.test.ts
@@ -44,6 +44,16 @@ const SKILL_MD_PATH = join(
   "SKILL.md",
 );
 
+const ANNOTATE_SKILL_MD_PATH = join(
+  import.meta.dir,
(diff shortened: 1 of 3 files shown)


backnotprop:
Merged. Good catch on the gate/json distinction, and thanks for wiring the guidance into the freshness test so it cannot drift. This will ride the next release.

PR #1461, opened 2026-09-02
feat(review): token hover cards (Tier 0, search-backed)

backnotprop:
Token hover cards for code review, Tier 0 of the design in DESIGN_token-hover-cards.html, scoped by DESIGN_hover-tier0.md.

Hovering an identifier in the diff for 350ms shows a card: symbol name, kind badge, best-effort signature with a "matched line" cue when approximate, a conservatively scanned doc comment, "Defined at path:line" with at most one "or possibly" runner-up, and up to five references with a link into the existing References panel. Everything is served by the existing ripgrep code-nav backend plus three enrichments (kind from the definition patterns that already matched, a bounded signature read-ahead, and a doc-comment scan for the five supported languages that strips tooling directives from both edges and returns nothing over returning garbage).

Interaction contract: 350ms dwell before any request exists, single in-flight with same-key join, 30-entry LRU flushed on diff snapshot change, 250ms leave grace with hover-into-card, scroll cancels except inside the card, and every failure (rg missing, timeout at 3s, below-threshold answers, detached anchors) renders nothing, silently. There is no spinner: the card appears fully formed or not at all, and there is no preloading of any kind by design.

Boundaries: POST /api/code-nav/hover in both server runtimes with guards copied from /resolve, which is byte-identical (a 4,512-case differential fuzz confirmed zero classification changes from the pattern-kind split). Cmd-click keeps today's References-panel behavior exactly; Alt-click is a quiet alias. The feature is absent from guides.show, read-only views, and GitButler committed views by the same conditional-prop gate as existing code-nav, and a cookie-only setting (default on) removes it entirely. The portable viewer bundle keeps only inert prop plumbing (stitching moved caller-side after review; gzip delta +0.08 kB, budgets green).

Adversarially reviewed before opening: two reproduced interaction races (a neighbor's answer landing in an open card; duplicate in-flight fetches on re-entry), the wheel-over-card dismissal, directive leakage into docs, and a stale-card edge were all fixed with fail-closed regression tests (each fix reverted to prove its test fails, then restored). Card copy follows the approved mock's post-cleanup rulings: no source chip, no ranking vocabulary, no latency readout, one footer hint, no em dashes in any rendered string.

Lanes: shared 121, endpoint 6 (dual runtime), DOM 29 across three registered suites, seam lane 758, full suite at the pre-existing 12-failure environmental floor, root typecheck clean, review and hook builds green, viewer manifest in sync.


backnotprop:
Screenshots of the shipped behavior, captured with Playwright against the built branch (real pointer moves, real 350ms dwell, real ripgrep resolves against a live review server; no DOM staging).

The full card: kind badge, signature with the matched-line cue, a real doc paragraph, definition link, and the capped references list.

Full hover card on withRetry showing function badge, signature, doc paragraph, and references

Two candidate definitions: uncertainty is shown as a second location line, never described with ranking vocabulary. The same-directory candidate genuinely ranked first.

Two-candidate card on normalizeHost showing Defined at and an or-possibly second line

Graceful degradation: a doc-less const renders just the badge, signature, and references. No empty sections.

Minimal card on DEFAULT_TIMEOUT_MS with const badge and two references

The diff at rest for context; nothing renders until a token is dwelled on.

The review diff before any hover

PR #1462, opened 2026-09-03
feat(review): hover card trigger settings and first-run introduction

backnotprop:
Trigger customizability and a first-run introduction for the token hover cards (#1461), designed doc-first in DESIGN_hover-settings.md (committed with the change).

The rational settings model, after pressure-testing: exactly two cookie-only settings. "Show cards" is one select with three values: On hover (default), While holding Alt (Option), or Off; folding Off into the select kills the unreachable enabled-false-plus-mode state a separate toggle would create. "Card delay" offers Fast 200, Default 350, Relaxed 700, because "cards annoy me" is three different complaints: never (Off), on request (Alt), or just less eager (delay). Alt was chosen over Cmd because Cmd already means "take me there" (Cmd+click opens the References panel, and Cmd hover paints the clickable affordance), while Alt spells the same on every platform.

Rejected with written arguments in the design doc: a click-to-open mode and any click-override of the References panel. Every click gesture on a token already has an owner (plain click opens the annotation toolbar, Cmd/Ctrl/Alt+click opens References, Shift extends selection), so a click mode could only evict a sitting tenant while the Alt mode already delivers cards-on-request with zero conflicts.

The one-time dialog joins the existing announcement chain as its fifth member (same gating conventions, provably never stacks with the other four) and carries the trigger choice inline as a radio that applies immediately; dismissing keeps the choice, and Settings under Review Display can change it anytime. Users who had already turned the earlier boolean setting off are migrated to Off, never see the dialog, and keep working even on a downgrade (the legacy cookie is read, never rewritten).

In modifier mode the Alt gate sits ahead of the dwell timer, the cache, and all state, so with the key up nothing is searched and no listeners exist outside that mode. Typing surfaces (inputs, textareas, CodeMirror) never arm the gate. Adversarial review drove two interaction fixes with revert-proven regression tests: a stale pointer-in-card flag that could deafen Alt releases after a scroll-close, and the Alt+click double surface: every References-panel invocation now dismisses or cancels the hover surface first, which also repairs the pre-existing hover-plus-modifier-click overlap from #1461.

Lanes: six touched suites 63 pass, full DOM lane 827 pass across 101 files, typecheck clean across all projects, full vendored suite at the pre-existing 12-failure environmental floor with name-identical failures, viewer manifest in sync and budgets green.


backnotprop:
Real screenshots of the hover-settings work, captured with Playwright against the built app (`apps/hook/dist/review.html` rebuilt from this branch, served by `plannotator review` on a demo git repo with a working-tree diff).

**1. First-run announcement dialog.** Reached through the real dialog chain on a fresh browser profile (guide intro "Got it", look-and-feel "Continue", edit mode "Done", then this arrives naturally). Note: the review-setup dialog did not appear in the chain; see the finding below.

Token hover cards announcement dialog, default state

**2. Same dialog with "While holding Alt (Option)" selected.** The radio writes the setting on selection; Done just closes.

Announcement dialog with the modifier trigger selected

**3. Settings, Editor tab, defaults.** The new controls ship under Settings > Editor (not a "Review Display" tab; see finding below): the "Token hover cards" trigger segmented control (On hover / Hold Alt / Off, On hover selected) and the delay segmented control captioned "How long the pointer rests before a card is requested" (Fast / Default / Relaxed, Default selected).

Settings Editor tab with both hover controls at their defaults

**4. Off selected.** The delay control stays visible but its three buttons are truly `disabled` (cursor-not-allowed, muted), preserving the remembered Default selection.

Settings with trigger Off and the delay control disabled

**5. Hold Alt selected.** The delay control is re-enabled.

Settings with the Hold Alt trigger selected

**Findings from driving the real build:**

- The dialog footer says "Change this anytime in Settings, under Review Display", but in review mode the Settings tab holding these controls is labeled "Editor" (registry id `display`, label "Editor" for `mode === 'review'` in `packages/ui/components/Settings.tsx`). The copy and the tab label do not match.
- On a genuinely fresh profile the ReviewSetupDialog never appears, so the token hover announcement is fourth in practice, not fifth: `configStore.ensureLoaded()` seeds the `reviewPanelView` default into its cookie before the diff callback runs `initializeReviewSetup()`, which then sees a persisted view, marks review-setup-seen, and returns false. This is pre-existing behavior (`reviewSetup.ts` is untouched by this branch), but it is visible in the chain this dialog joins.
- The Settings labels are "On hover / Hold Alt / Off" and "Fast / Default / Relaxed"; the "Show cards" wording appears only as the dialog's radiogroup label.

backnotprop:
## Hover cards: the announcement is now a try-it on the real card

Recaptured with Playwright against the built app (`apps/hook/dist/review.html` rebuilt from this branch, served by `plannotator review` on a demo git repo with a working-tree diff), reached through the real first-run dialog chain on a fresh profile.

### The dialog, with the try-it card open

The `withRetry` token in the strip is genuinely hoverable. Resting on it opens the **real** `TokenHoverCard` through the **real** `useTokenHover`, so you feel the actual dwell, the leave grace and the Alt gate before you pick anything. The only fixture is where the answer comes from.

Hover cards announcement dialog with the try-it card open

### The try-it card and a real card, side by side

Left: the card the try-it opens inside the dialog. Right: a real hover card opened by hovering `resolveToken` in the diff behind it, in the same browser session.

Try-it card and a real hover card side by side

Same component, verified in the browser rather than by eye:

- surface classes equal: `w-[400px] max-w-[calc(100vw-2rem)] rounded-xl border border-border bg-popover text-popover-foreground shadow-2xl` (only the positioning layer differs, `z-[110]` over the modal vs the app's `z-[60]`)
- computed box identical on all of `width`, `borderRadius`, `borderWidth`, `borderColor`, `backgroundColor`, `color`, `boxShadow`, `fontFamily`, `fontSize`
- `/api/code-nav` requests during the whole try-it: **0**. The one request in the run is the real card behind the dialog.
- try-it location buttons `tabIndex` `[-1,-1,-1]`; the live card's are `[0,0,0,0,0,0]`

The right card shows no doc line and the left one does, because the real search found no doc comment for `resolveToken`. That is the same component rendering an optional section, not a difference in markup.

### How it is wired

No markup or timing is duplicated. Three thin seams:

| Seam | Why |
|---|---|
| `useTokenHover({ resolve })` | The only fixture. Default still posts to `/api/code-nav/hover`; a demo must not search your repository for a symbol you never asked about, and a second copy of the dwell logic would drift. |
| `TokenHoverCard({ layerClassName })` | The card portals to `<body>`, so its host has to say which stacking context it belongs in. |
| `TokenHoverCard({ inert })` | The try-it has no References panel behind it, so its locations leave the tab order. |

The hovered token wears the diff pane's own underline treatment, defined once in `tokenHoverStyles.ts` and consumed both by Pierre's shadow stylesheet (`.pn-token-hover`) and by the dialog.

The try-it reads the **live** setting, so choosing "While holding Alt" makes the demo behave that way immediately, and "Off" makes it do nothing and closes any open card.

### Rename

User-facing name is **"Hover cards"** in the dialog title and the Settings heading. Every internal identifier is unchanged (`tokenHoverTrigger`, `tokenHoverDelay`, the cookie keys, the `data-token-hover-*` ids), so there is no migration.

### Checks

- hover + dialog DOM suites: 66 pass, 0 fail
- full DOM lane (101 files): 830 pass, 0 fail
- typecheck (core, shared, ui, ui strict-consumer): clean
- full vendored suite: 4298 pass, 12 fail, name-identical to the floor
- `check:manifest` in sync; `check:budgets` 395.7 KB gz of 400

backnotprop:
Updated announcement dialog after 51e85d5a: the modifier gate now reads "While holding Cmd" (Ctrl elsewhere), and the footer note leads with Cmd+click.

The dialog with the Cmd gate wording

Try-it demo with On hover selected, the live card opens on the withRetry token inside the dialog.

The try-it card open over the demo token

PR #1468, opened 2026-09-04
fix(ui): keep annotation toolbar within viewport

katya4oyu:
## Problem

The shared selection/block annotation toolbar positions only its anchor point. Near a mobile viewport edge, the rendered toolbar box can extend off-screen even though the anchor itself is visible.

## Change

- measure the rendered toolbar before revealing it
- preserve the existing `center-above` and `top-right` preferred positions when they fit
- clamp the complete box to the shared visible-viewport bounds with 16px padding
- reuse the existing viewport observer instead of adding parallel resize/Visual Viewport listeners
- run the focused DOM placement regression in CI

This intentionally does not change toolbar actions, ordering, or desktop composition. Image annotation toolbar wrapping is a separate design decision and is not part of this PR.

## ADR alignment

This follows `adr/specs/mobile-platform-foundation-phase1-20260812.md`: CSS/content geometry determines layout; shared UI remains shared; overlay placement uses the shared visible-viewport contract; and fitting mobile geometry does not mutate desktop preferences or redesign information architecture.

## Verification

- `bun test packages/ui/components/AnnotationToolbar.placement.test.tsx`
- `DOM_TESTS=1 bun test --isolate packages/ui/components/AnnotationToolbar.placement.test.tsx packages/ui/components/AnnotationToolbar.commentOnly.test.tsx packages/ui/components/AnnotationToolbar.modeShortcutGuard.test.tsx`
- `bun test packages/ui` (788 pass, 574 DOM-gated skips, 0 fail)
- `bun run build:review`
- `bun run build:hook`
- `git diff --check`

Physical Mobile Safari verification remains pending.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -174,6 +174,7 @@ jobs:
           packages/ui/hooks/useSharing.contentRevision.test.tsx
           packages/ui/hooks/useSharing.shortUrlLifecycle.test.tsx
           packages/ui/components/AnnotationToolbar.commentOnly.test.tsx
+          packages/ui/components/AnnotationToolbar.placement.test.tsx
           packages/ui/components/AnnotationToolstrip.test.tsx
           packages/ui/components/StickyHeaderLane.test.tsx
           packages/ui/components/CommentPopover.quickLookGood.test.tsx
--- a/packages/ui/components/AnnotationToolbar.placement.test.tsx
+++ b/packages/ui/components/AnnotationToolbar.placement.test.tsx
@@ -1,0 +1,148 @@
+import { afterEach, describe, expect, test } from 'bun:test';
+import React, { act } from 'react';
+import { createRoot, type Root } from 'react-dom/client';
+import { AnnotationToolbar, computeAnnotationToolbarPosition } from './AnnotationToolbar';
+import type { VisibleViewportBounds } from '../hooks/useViewportEnvironment';
+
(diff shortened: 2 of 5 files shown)


katya4oyu:
Physical-iPhone follow-up from the reported screenshot:

- Root cause: the coarse-pointer selection bridge called `highlighter.fromRange()` while leaving Safari’s native `Selection` alive. `fromRange()` replaces/splits the selected text nodes, so WebKit kept blue handles attached to stale document boundaries even though the app mark preserved the intended text.
- Fix: clear the native range in `finally` immediately after the app mark is created. The green pending mark and annotation toolbar remain; only the stale native overlay/handles are removed.
- Regression: added a coarse-pointer DOM test asserting both `mark.textContent === "Formula"` and `window.getSelection().rangeCount === 0`. It failed before the fix (rangeCount 1) and passes after it.

Verification on commit `0a50cf93b55bfb2d797f010992e56c5ca9aa7d50`:
- focused placement + selection tests: 11 passed
- full UI suite: 788 passed, 575 DOM-gated skips, 0 failed
- UI typecheck, review build, hook build, and `git diff --check`: passed
- rendered 390px coarse-pointer preview: app mark remained exactly `スタイル`; native Selection was empty (`rangeCount=0`); toolbar stayed within x=16..356.

Physical iPhone recheck remains the acceptance gate.

Issue #1464, opened 2026-09-03
CI flake: annotate-server durable-record tests fail on runners, pass locally


backnotprop:
Three occurrences in two days, distinct PRs, same test family:

- packages/server/annotate.test.ts, "annotate server: folder annotate history > an unwritable history directory degrades to a plain render"
- "durable submit records (#678) > feedback submit writes a durable record and only then deletes the draft"
- "durable submit records (#678) > approve with notes persists a record; a bare approve writes nothing"
- "durable submit records (#678) > a failed durable write keeps the draft as the recovery copy"

Seen on PR #1461 (run 33689988840) and twice on PR #1462 (runs 33683325163-family and 33717306176). In every case the full suite passes locally on the same commits (58/58 in the file), and a plain rerun goes green. Neither PR touches the annotate server.

These tests boot a real server against a temp PLANNOTATOR_DATA_DIR and exercise filesystem ordering (write record, then delete draft; chmod-unwritable directories). Suspect timing or permission semantics on the shared runners (chmod on hosted runners, write visibility ordering under load).

Fix shapes to evaluate: make the unwritable-directory setup assert its own precondition and skip honestly when chmod does not stick on the runner filesystem; widen the record-then-draft ordering assertions to poll briefly instead of asserting immediately after the response; or serialize this describe block if cross-test contention on the temp dir is the trigger.

Also worth recording: the agent-terminal sidecar test (packages/server/agent-terminal.test.ts, "node sidecar rebuilds spawn options") flaked the same way on the #1458 integration run. Different family, same pattern: server tests doing real fs/process work on shared runners.

backnotprop:
Fourth occurrence: PR #1465 run 33718368682, the identical four tests. This is now firing on roughly half of all CI runs today, which crosses from annoyance into velocity tax. Raising the priority suggestion: the poll-instead-of-assert fix shape looks smallest and addresses three of the four (the write-then-delete ordering assertions); the unwritable-directory case needs its chmod precondition check.

backnotprop:
Root-caused and fixed in v0.27.13. The flake was import-time capture of the data directory in shared storage: an OpenCode lifecycle test importing storage under a temporary env override poisoned later test files, with filesystem-dependent test discovery order deciding which runs got hit. Fixed by per-call resolution in #1473 (with the suite-wide data-dir sandbox), independently root-caused by @bendrucker during #1437 with the definitive diagnosis. Remaining frozen-at-import modules are tracked in #1477.

PR #1473, opened 2026-09-05
fix(test): isolate test-run data from contributor history

### LINKED ISSUE #1455: Test suite writes plan/annotate history into the real ~/.plannotator/history
A full `bun test` run currently creates roughly 445 plan/annotate history directories under the real `~/.plannotator/history/`. The `tests/setup/feedback-archive-off.ts` preload silences only the feedback archive; version-history writes from server tests that boot without a redirected `PLANNOTATOR_DATA_DIR` still land in the contributor's real data dir.

This violates the Testing Rules paragraph (sandbox all server/data-dir interaction under a temp `PLANNOTATOR_DATA_DIR`; never read or write the real user config).

Likely fix shape: extend the preload to point `PLANNOTATOR_DATA_DIR` at a per-run temp directory by default (tests that need the real resolution logic can opt out inside their bodies, mirroring how the archive preload works), or audit the server test files that boot without redirecting and fix them individually.

Found during an unrelated review pass; not introduced by any recent PR. Pre-existing on main.

FNDEVVE:

Fixes #1455.

## Summary
- Allocate a fresh process-owned `PLANNOTATOR_DATA_DIR` in the existing test preload, before store imports can capture a real user path. An inherited contributor data-directory setting cannot bypass isolation.
- Clean only that captured temporary directory in Bun's global test teardown; preserve archive-off defaults and deliberate test-body overrides.
- Adapt controlled-HOME prompt/hook fixtures and the nested disabled-AI test to the new isolation contract. No production storage behavior changes.

## Regression coverage
A real child `bun test` exercises static store imports, plan/annotate/config writes, runtime-child inheritance, nested test-process isolation, afterAll writes, and owned-directory cleanup. Caller-owned directories remain untouched.

## Verification
- Repository-pinned Bun 1.3.14, disposable HOME, local macOS environment (`CI` unset).
- Focused data-dir, prompts, improvement-hooks, and disabled-AI tests: **33 passed**.
- Full `bun test`: **4,297 passed, 910 skipped, 0 failed**.
- `bun run typecheck` and `git diff --check`: passed.
- Independent Standards and Spec reviews completed, including fixture follow-up changes.

### CI correction (`fa3a8fbe`)

Resolved the import-order bug behind #1464: storage now resolves the active data directory for every plan/history read and write instead of retaining an early test's temporary override. Added a regression for changing that override after import; no sleeps, retries, or weakened assertions. The ordered OpenCode→annotate reproduction fails four tests before this fix and passes after it on both macOS and Linux. Unprivileged Linux verification: 83 pass, 0 fail. Latest local full suite: 4,298 pass, 910 skip, 0 fail; typecheck passes.

### DIFF
--- a/bunfig.toml
+++ b/bunfig.toml
@@ -5,4 +5,4 @@ minimumReleaseAge = 604800  # 7 days in seconds
 minimumReleaseAgeExcludes = ["@opencode-ai/ai", "@opencode-ai/client", "@opencode-ai/plugin", "@opencode-ai/protocol", "@opencode-ai/schema", "@pierre/diffs", "@pierre/theme", "@pierre/theming", "@plannotator/atomic-editor", "@plannotator/markdown-editor", "@plannotator/webtui"]
 
 [test]
-preload = ["./packages/ui/test-setup/happy-dom.ts", "./tests/setup/feedback-archive-off.ts"]
+preload = ["./tests/setup/feedback-archive-off.ts", "./packages/ui/test-setup/happy-dom.ts"]
--- a/packages/server/ai-disabled.test.ts
+++ b/packages/server/ai-disabled.test.ts
@@ -1,9 +1,6 @@
 import { expect, test } from "bun:test";
 import type { PRMetadata } from "@plannotator/shared/pr-types";
 import type { WorktreePool } from "@plannotator/shared/worktree-pool";
-import { existsSync, mkdtempSync, rmSync } from "node:fs";
-import { tmpdir } from "node:os";
-import { join } from "node:path";
 import { fileURLToPath } from "node:url";
 
 const SPA_HTML = "<!doctype html><html><body>test</body></html>";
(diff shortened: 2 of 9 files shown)


FNDEVVE:

**Correction to my earlier rerun advice:** the failure is now reproduced and fixed in `fa3a8fbe`. It is deterministic import-order contamination, not a permission/timing issue requiring a maintainer rerun.

```sh
```

Verification: unprivileged Linux ordered three-file suite **83 pass, 0 fail**; this branch's local full suite **4,298 pass, 910 skip, 0 fail**, plus typecheck. Local full-suite environment uses a disposable HOME and the pinned Bun 1.3.14; upstream Actions verifies the Linux CI environment. Independent Standards and Spec reviews found no issues.

**Final upstream verification:** all checks on the corrected head are green (publication-only jobs are intentionally skipped on PRs). [Test workflow](https://github.com/backnotprop/plannotator/actions/runs/33968037911) and [Release build/smoke workflow](https://github.com/backnotprop/plannotator/actions/runs/33968037903) both passed, including Linux/macOS/Windows binary smoke checks. No maintainer rerun is needed.

backnotprop:
Full adversarial review complete: revert-verification confirms both regression tests genuinely fail against the old code on their respective halves (per-call storage fix, and the preload sandbox), no consumer anywhere relied on the frozen-at-import behavior, the preload change is pure isolation with no masking, and the CI-equivalent full suite runs 4308 pass / 0 fail with typecheck clean. One non-blocking nit we'll sweep in a follow-up: the retained comment in ai-disabled.test.ts (~line 178) still cites the import-time capture this PR removes as the reason for the child process. Merging - and thank you: this plus bendrucker's independent convergence on the same root cause closed out our longest-standing CI flake.

PR #1474, opened 2026-09-05
fix(review): show setup after fresh-profile settings initialization

### LINKED ISSUE #1463: ReviewSetupDialog never shows on a truly fresh profile
Found during screenshot verification of #1462, pre-existing on main (packages/review-editor/utils/reviewSetup.ts is untouched by that branch).

On a genuinely fresh browser profile, the first-run dialog chain runs guide intro, look and feel, then skips straight past ReviewSetupDialog to the Edit Mode announcement. Verified with a step-by-step dialog dump during real chain traversal.

Mechanism: configStore.ensureLoaded() seeds every settings-registry default into the cookie jar on first settings access, including reviewPanelView. When initializeReviewSetup() runs afterward, it finds a reviewPanelView cookie, treats it as a persisted reviewer choice, marks review-setup-seen, and returns false. The dialog that CLAUDE.md documents as "first-run shows ReviewSetupDialog" is therefore unreachable for new users; it only appears via the header menu reopen path.

Impact: new reviewers never get the one-time panel/diff-default setup choice. Low severity since the defaults are sensible and the dialog is reopenable, but the documented first-run behavior does not happen.

Fix shape to evaluate: initializeReviewSetup should distinguish a registry-seeded default from a user-persisted choice (the registry seeding writes the built-in default; a marker or ordering change would disambiguate), or run before ensureLoaded seeds that key, mirroring how the WebMCP preference deliberately stayed out of the registry to avoid exactly this seeding side effect.

FNDEVVE:

Fixes #1463.

## Summary
`configStore` seeded `reviewPanelView` before the review setup gate ran, so a fresh profile looked like a returning reviewer with a saved choice.

- Add a per-setting `persistDefault` opt-out and apply it only to `reviewPanelView`, in both default-loading paths.
- Keep the resolved default in memory; explicit choices still persist normally.
- Preserve existing initialization, last-used memo, server diff-default precedence, and all other settings' default persistence. No new migration cookie or startup server write.

## Verification
- Focused setup/config regression suite: **22 passed**. Tests hydrate the real settings store before the setup initializer and cover returning/explicit choices.
- Full `bun test` with repository-pinned Bun 1.3.14, disposable HOME, and local macOS environment (`CI` unset): **4,303 passed, 904 skipped, 0 failed**.
- `bun run typecheck`, `git diff --check`, and review → hook production builds: passed.
- Real rebuilt CLI/browser smoke: fresh-profile look-and-feel → **Set up your review view** → Edit Mode → token-hover announcement; reload does not repeat setup and retains the Tree preference/memo.
- Independent Standards and Spec reviews completed. Pierre renderer integration is unchanged.

### CI correction (`c558b50d`)

Resolved the import-order bug behind #1464: storage now resolves the active data directory for every plan/history read and write instead of retaining an early test's temporary override. Added a regression for changing that override after import; no sleeps, retries, or weakened assertions. The ordered OpenCode→annotate reproduction fails four tests before this fix and passes after it on both macOS and Linux. Unprivileged Linux verification: 83 pass, 0 fail. Latest local full suite: 4,304 pass, 904 skip, 0 fail; typecheck passes.

### DIFF
--- a/packages/core/guide-viewer-manifest.ts
+++ b/packages/core/guide-viewer-manifest.ts
@@ -5,9 +5,9 @@
 import type { GuideViewerAssets } from "./guide-format";
 
 export const GUIDE_VIEWER_MANIFEST: Omit<GuideViewerAssets, "baseUrl"> = {
-  js: "viewer.C8JgNbZu.js",
+  js: "viewer.-z4hZ-DF.js",
   css: "viewer.KIp-iPxY.css",
-  jsIntegrity: "sha384-dqRkZ2N5WNwLIuwa+1kdrvswkOVHpkBxrrJ5mTU4YrUWzO/8fRS5MRTlDGV9t7eB",
+  jsIntegrity: "sha384-0m4lvEvRi8w3vafYuzG3N918AAH78sDPytVzerfzpOBxU63u3VrdyhHpBOVpDT4u",
   cssIntegrity: "sha384-Ty9hGpag8KIAGMDPg7ImsB4L5RabqvonNVjrj/CnYqSJDqIgPRBwjEMoK9/EvgXm",
   langs: {
     "astro": "chunks/astro.BykyiR6i.js",
--- a/packages/review-editor/utils/reviewSetup.test.ts
+++ b/packages/review-editor/utils/reviewSetup.test.ts
@@ -3,6 +3,7 @@ import { readFileSync } from 'node:fs';
 import { join } from 'node:path';
 import { resetStorageBackend, setStorageBackend } from '@plannotator/ui/utils/storage';
 import { ConfigStoreForTest } from '../../ui/config/configStore';
+import { setReviewPanelView } from '../../ui/config/reviewView';
 import {
   initializeReviewSetup,
   needsReviewSetup,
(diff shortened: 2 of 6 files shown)


FNDEVVE:

**Correction to my earlier rerun advice:** the failure is now reproduced and fixed in `c558b50d`. It is deterministic import-order contamination, not a permission/timing issue requiring a maintainer rerun.

```sh
```

Verification: unprivileged Linux ordered three-file suite **83 pass, 0 fail**; this branch's local full suite **4,304 pass, 904 skip, 0 fail**, plus typecheck. Local full-suite environment uses a disposable HOME and the pinned Bun 1.3.14; upstream Actions verifies the Linux CI environment. Independent Standards and Spec reviews found no issues.

**Final upstream verification:** all checks on the corrected head are green (publication-only jobs are intentionally skipped on PRs). [Test workflow](https://github.com/backnotprop/plannotator/actions/runs/33968275817) and [Release build/smoke workflow](https://github.com/backnotprop/plannotator/actions/runs/33968275853) both passed, including Linux/macOS/Windows binary smoke checks. No maintainer rerun is needed.

backnotprop:
Heads up: #1473 and #1476 just merged, and this branch now conflicts with main (the shared storage.ts change landed via #1473, so the conflict should resolve to simply taking main's side there plus your own review-setup changes). A rebase onto current main should be quick - once it's green this is next in the review queue.

FNDEVVE:
Rebased onto current main (dropping the storage commit now upstream via #1473), rebuilt the viewer, and synced the manifest. Local full suite (4,321 pass, 906 skip, 0 fail), typecheck, and viewer budget/manifest checks all pass.

backnotprop:
Thanks @FNDEVVE, this is the PR that surfaced the real path: the settings store seeding the default before the first-run check. The owner ruled the other way on the dialog itself, so #1587 removes it and makes Tree the default rather than making the chooser appear, with the toggle and Settings > Git as the ways to switch. Closing this in favour of that; #1463 is closed by it. Ships in the next release with credit to you in the notes.

PR #1475, opened 2026-09-05
fix(share): allow explicit short links for small plans

### LINKED ISSUE #1427: Allow minting a replacement short link for small shared plans
## Context

PR #1425 fixes #798 by treating incoming short links as immutable snapshots and removing a stale short URL after the represented content changes.

A remaining product limitation affects small markdown plans: `ExportModal` only offers **Create short link** when the full hash URL exceeds 2048 characters or hash sharing is unavailable. After editing a small plan loaded from a short link, the user gets a correct fresh full URL, but cannot explicitly mint a replacement short link from Export. This leaves the literal short-link step in the workflow reported in #798 incomplete for small plans.

## Intended behavior

After an incoming or locally generated short URL is invalidated by a material edit, Export should offer an explicit way to create a fresh short link even when the current full hash URL is below the large-URL threshold.

## Constraints

- Keep pastes immutable: never update or delete the original paste.
- Creating a replacement remains an explicit user action; do not auto-upload on edits.
- Preserve current payload, encryption, and existing-link compatibility.
- Decide whether the action should appear only after invalidation or be available for every shareable small plan.

## Acceptance criteria

- A small markdown plan loaded from a short link loses the stale URL after an edit.
- Export exposes a replacement short-link action.
- Invoking it creates a distinct encrypted paste representing current content.
- The original paste remains unchanged.

Tracked separately from #1425 per the maintainer review so that PR stays focused on stale-snapshot correctness.

FNDEVVE:

Fixes #1427.

## Summary
Offer **Create short link** for every shareable small plan, not only after a previous link was invalidated. This resolves the issue's open UI choice without adding invalidation-history state.

The production change removes only the size-based action gate in `ExportModal`; large-URL warnings remain size-gated. Existing sharing capabilities, payload/encryption, immutable-paste lifecycle, and generation callback are unchanged. Creation still requires the explicit button click—no automatic uploads.

## Verification
- Rendered modal + real sharing-hook lifecycle regressions: **6 passed** (`DOM_TESTS=1`, Bun 1.3.14). Cover fresh-small eligibility, stale incoming link invalidation, explicit replacement, decrypting current annotations, original-paste immutability, and disabled/unsupported capabilities.
- Full `bun test` with repository-pinned Bun 1.3.14, disposable HOME, and local macOS environment (`CI` unset): **4,302 passed, 907 skipped, 0 failed**. An earlier concurrent run hit the unrelated Pi stdio timeout; that test passed independently and the complete suite then passed.
- `bun run typecheck`, `git diff --check`, and review → hook production builds: passed.
- Actual browser smoke with local paste storage: create a link for a 95-byte full URL, open that encrypted snapshot, add a comment, verify the old link disappears with zero automatic uploads, then explicitly create a distinct replacement. Decryption contains the new comment; original ciphertext is byte-identical. No paste was uploaded to the public service.
- Independent Standards and Spec reviews completed.

### CI correction (`3cbe5712`)

Resolved the import-order bug behind #1464: storage now resolves the active data directory for every plan/history read and write instead of retaining an early test's temporary override. Added a regression for changing that override after import; no sleeps, retries, or weakened assertions. The ordered OpenCode→annotate reproduction fails four tests before this fix and passes after it on both macOS and Linux. Unprivileged Linux verification: 83 pass, 0 fail. Latest local full suite: 4,303 pass, 907 skip, 0 fail; typecheck passes.

### DIFF
--- a/packages/server/call-flow-install-endpoint.test.ts
+++ b/packages/server/call-flow-install-endpoint.test.ts
@@ -6,15 +6,9 @@ import { join } from 'node:path';
 import type { CallFlowInstallStage, CallFlowNodePreflight, CallFlowRuntimeInstallResult } from '@plannotator/shared/call-flow';
 
 // PLANNOTATOR_DATA_DIR is only ever changed INSIDE tests (boot() below) and
-// restored to its original value after each one. It must never be overridden
-// at module-eval time: bun evaluates every test file's module before running
-// tests in one shared process, and Pi's generated/storage.ts caches its data
-// dir at import time. A module-eval override here makes storage's cached dir
-// and later files' live getPlannotatorDataDir() calls disagree, which is
-// exactly the Pi annotate-history / durable-submit CI failure this comment
-// guards against. Config writes made by these tests target whatever dir the
-// process's config module froze at first import; the snapshot/restore in
-// afterAll below keeps those writes from leaking into a real config.json.
+// restored after each one. Module-eval overrides would leak into other test
+// files because Bun runs the suite in one shared process. The config
+// snapshot/restore in afterAll also protects against shared config state.
 const originalDataDir = process.env.PLANNOTATOR_DATA_DIR;
 const originalPort = process.env.PLANNOTATOR_PORT;
 const originalPath = process.env.PATH;
--- a/packages/server/storage.test.ts
+++ b/packages/server/storage.test.ts
@@ -174,3 +174,39 @@ describe("listVersions", () => {
     expect(versions[0].timestamp).toBeTruthy();
   });
 });
+
+describe("PLANNOTATOR_DATA_DIR", () => {
+  test("isolates plan and history data when the data directory changes after import", () => {
+    const savedDataDir = process.env.PLANNOTATOR_DATA_DIR;
+    const firstDir = makeTempDir();
+    const secondDir = makeTempDir();
+    const project = "data-dir-project";
+    const slug = "data-dir-plan";
+
+    try {
+      process.env.PLANNOTATOR_DATA_DIR = firstDir;
+      savePlan(slug, "# First plan");
+      saveToHistory(project, slug, "# First version");
+      expect(readFileSync(join(firstDir, "plans", `${slug}.md`), "utf-8")).toBe("# First plan");
+      expect(getPlanVersion(project, slug, 1)).toBe("# First version");
+      expect(getVersionCount(project, slug)).toBe(1);
+
+      process.env.PLANNOTATOR_DATA_DIR = secondDir;
+      expect(getPlanVersion(project, slug, 1)).toBeNull();
+      expect(getVersionCount(project, slug)).toBe(0);
+      savePlan(slug, "# Second plan");
+      saveToHistory(project, slug, "# Second version");
+      expect(readFileSync(join(secondDir, "plans", `${slug}.md`), "utf-8")).toBe("# Second plan");
+      expect(getPlanVersion(project, slug, 1)).toBe("# Second version");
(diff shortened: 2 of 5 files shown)


FNDEVVE:

**Correction to my earlier rerun advice:** the failure is now reproduced and fixed in `3cbe5712`. It is deterministic import-order contamination, not a permission/timing issue requiring a maintainer rerun.

```sh
```

Verification: unprivileged Linux ordered three-file suite **83 pass, 0 fail**; this branch's local full suite **4,303 pass, 907 skip, 0 fail**, plus typecheck. Local full-suite environment uses a disposable HOME and the pinned Bun 1.3.14; upstream Actions verifies the Linux CI environment. Independent Standards and Spec reviews found no issues.

**Final upstream verification:** all checks on the corrected head are green (publication-only jobs are intentionally skipped on PRs). [Test workflow](https://github.com/backnotprop/plannotator/actions/runs/33968275322) and [Release build/smoke workflow](https://github.com/backnotprop/plannotator/actions/runs/33968275302) both passed, including Linux/macOS/Windows binary smoke checks. No maintainer rerun is needed.

backnotprop:
Reviewed in depth - the feature itself passes: privacy contract holds (upload only ever fires from the button's onClick, no effect/retry path; PLANNOTATOR_SHARE=disabled suppresses the whole tab so the button can never render), revert-verification confirms the new tests genuinely guard the feature, and the stacked storage.ts change is byte-identical to #1473's.

One blocker before merge: the new test "offers an explicit replacement for an edited small incoming encrypted snapshot" is flaky - 6 failures in 10 runs of the exact CI invocation (`DOM_TESTS=1 bun test --isolate`). The race: the first `waitFor(isLoadingShared === false && Boolean(shareUrl))` can release while `shareUrl` still holds the pre-hydration empty-document URL, and the post-edit `waitFor(shareUrl !== hydratedShareUrl)` then releases when the *hydrated* URL commits, before the *edited* URL exists - so the decompress at the end finds "Original document" instead of "Edited document".

Suggested fix: wait on content rather than URL inequality - poll until `decompress(shareUrl)` yields the expected markdown, both before capturing `hydratedShareUrl` and after the edit. The pre-existing "preserves an incoming short URL through hydration" test has the same latent shape and would be worth fixing in passing.

Two optional cosmetic notes, non-blocking: the small-plan button renders inside the amber warning container with no warning text (a neutral container might fit better when `!urlIsLarge`), and an empty session renders a silent no-op button (practically unreachable).

Happy to merge as soon as the flake is fixed - everything else is done.

FNDEVVE:
Addressed in `615e5f6e`:

1. **Flake fix:** added `waitForSharedPayload` to poll until `decompress(shareUrl)` yields the expected payload before capturing `hydratedShareUrl` and after the edit in `offers an explicit replacement for an edited small incoming encrypted snapshot`. Also updated `preserves an incoming short URL through hydration, then invalidates immutable snapshots after material edits` to wait on the hydrated and 2-annotation payloads rather than URL inequality.
   - Local verification: **20/20 runs passed** under `DOM_TESTS=1 bun test --isolate`.
2. **Cosmetic styling:** updated `ExportModal` so the short link button container uses a neutral style (`bg-muted/50 border-border`) when `!urlIsLarge`, reserving the amber container for URL length warnings.
3. **CI verification:** all upstream checks green on commit `615e5f6e` (Test suite, Release build, binary smoke on macOS/Linux/Windows, guides.show budgets and pinned manifest).

backnotprop:
Merged, thanks @FNDEVVE, and sorry it sat after the flake fix. Ships in the next release.

PR #1476, opened 2026-09-05
fix(amp): preserve feedback with structured decision delivery

### LINKED ISSUE #1456: Amp relay can swallow an entire review when feedback text contains 'has no feedback'
The Amp plugin's relay classifier `isNoActionFeedback` (apps/amp-plugin/plannotator.ts, around lines 241-251) decides whether a review outcome is a no-op by substring matching the lowercased output, including the pattern "has no feedback". A reviewer whose real feedback contains that phrase, for example "this path has no feedback loop, add one", is classified as no-action: the entire review is demoted to a notification and never appended to the agent thread.

This is pre-existing behavior on the deny/feedback path. The PR5 approve-with-notes delivery work widens the reachable surface (approvals can now carry feedback text too), which is how it was found, but the defect is independent of that stack.

Fix shape: classify on structured decision data rather than substring matching rendered prose. The CLI already knows whether a decision carried content; the relay should key on that (or on an explicit marker line), not on user-authored text.

Found during an adversarial review pass of the decision-control stack (PR5 stage).

FNDEVVE:

Fixes #1456.

## Problem
Amp classified rendered feedback by substring, so a real comment such as `This path has no feedback loop, add one.` was treated as no action and never appended to the agent thread. The same classifier also swallowed annotate feedback and approval notes.

## Change
- Add direct `plannotator review --json`: one `{ decision, message }` record. `message` is the CLI-rendered agent instruction; this deliberately differs from raw annotate/OpenCode `feedback`.
- Generate plaintext and JSON from the same review output builder, preserving prompt customization, approval-with-notes framing, and annotation-dependent verification suffixes.
- Make both Amp review commands request structured output and route solely by decision. Remove the prose classifier from all review/annotate paths.
- Retain bare review approvals in the thread and bare annotate approvals as notifications. Invalid/legacy CLI output produces an actionable update notice containing captured output for recovery, never a guessed decision.
- Document the protocol and compatible-CLI requirement. No server endpoint or OpenCode protocol changes.

## Verification
- Reproduced the original bug through registered Amp commands and a real subprocess: ordinary feedback appended; the reported phrase in review, annotate, and approval notes did not.
- Registered-command regressions: **30 passed**. Focused producer/help/reference and dual-runtime approval-note suite with Amp: **71 passed**.
- Real Amp plugin → source CLI → disposable Git repo → HTTP decision → thread delivery smoke: review feedback, approval notes, bare approval, dismissal, and annotate feedback all passed.
- Actual direct CLI JSON/plaintext equivalence: annotation suffix, approval notes, bare approval, dismissal; plaintext equals `JSON.message + "\n"` byte-for-byte.
- Pinned Bun **1.3.14**, disposable HOME, local CI unset: full suite **4,319 passed, 904 skipped, 0 failed**; `bun run typecheck` passed.
- Review → hook builds passed. Portable viewer build, budget check, and manifest check passed; no manifest change needed.
- Independent Standards and Spec reviews: no remaining findings.

Includes the same previously reviewed storage data-directory lifetime correction carried by #1473–#1475, keeping the known #1464 import-order CI failures fixed until that prerequisite lands. The Amp-specific commit is separate.

No competing pending PR addressing #1456 was found in open PR descriptions or issue timeline references; rechecked immediately before publication.


backnotprop:
Deep review complete before merge, since this adds a CLI output contract every integration shares. Verified: the plaintext path is byte-identical for all existing consumers (framing, newline boundaries, exit codes - revert-verification proved the byte-compat test genuinely guards the seam); both version-skew directions fail closed with the recoverable update notice, tested across nine malformed-output shapes; nothing the old Amp routing delivered is dropped, and the #1456 class of feedback is now delivered instead of silently swallowed. 270/0 tests, typecheck clean, docs match the implementation exactly. Two style-only nits left as-is (duplicated return null in parseReviewDecision; ReviewOutcome as a local structural type). Merging.

PR #1480, opened 2026-09-07
feat(vibe): add first-class Mistral Vibe support

Djiit:
## TL;DR

Makes Mistral Vibe a detected origin in Plannotator. The plan-review gate, annotate-last, and installer wiring that were hand-configured now happen automatically. Vibe's `exit_plan_mode` tool sends no plan in its payload, so the gate reads the newest plan from `$VIBE_HOME/plans` by mtime.

## What changed

**Origin registration**
- `mistral-vibe` added to `AGENT_CONFIG` (rose badge), the `PromptRuntime` union, and `PLAN_TOOL_NAMES` (mapped to Vibe's lowercase `exit_plan_mode`). No Ask AI provider — Vibe falls back to the saved/server default, same as `amp`/`droid`/`copilot-cli`/`gemini-cli`.

**Plan-review gate** (`apps/hook/server/index.ts`, `packages/shared/vibe-plans.ts`)
- New `resolveLatestVibePlan` reads the newest `*.md` in `$VIBE_HOME/plans` by mtime. Vibe's `exit_plan_mode` takes no args, so the plan isn't in the hook payload.
- Approve → `{"decision":"allow"}` on stdout, exit 0. Deny → `{"decision":"deny","reason":...}`. No plan found → empty stdout + exit 0 so `exit_plan_mode` proceeds and the user isn't locked out of plan mode. Hook block uses `strict: false` (Vibe's default), so a crash is fail-open too.
- No `improve-context` analog — Vibe has no `EnterPlanMode` moment. Documented in the install banner, not wired.

**annotate-last** (`apps/hook/server/session-log.ts`)
- `resolveVibeSessionLogForCwd` walks `$VIBE_HOME/logs/session/.session_index.json` for the session whose `cwd` matches, newest `mtime_ns` wins. The index is authoritative: a present-but-no-match returns null rather than falling through to a cwd-blind mtime scan. Ancestor-walk fallback only when the index is absent.
- `getLastVibeRenderedMessage` parses `messages.jsonl` backward for the last assistant turn with non-empty string content (skips reasoning-only and tool-call-only turns).

**Installer** (`install.sh` / `install.ps1` / `install.cmd`)
- Three-layer opt-out matching the per-agent family: `--skip-vibe` flag > `PLANNOTATOR_SKIP_VIBE_INSTALL` env var > `skipInstall.vibe` config key. Each skip remembers its source for honest reporting.
- `VIBE_HOME` detected via `command -v vibe` or `~/.vibe` existing, honors a `VIBE_HOME` override (mirrors `CODEX_HOME`).
- **sh**: managed marker block in `hooks.toml` (`# >>> plannotator-managed-vibe-hooks >>>` … `<<<`) so it coexists with existing hooks; `enable_experimental_hooks = true` merged into `config.toml`. Atomic rewrite with a `.bak`.
- **ps1/cmd**: hooks are macOS/Linux-only (Vibe spawns hooks via `/bin/sh`; Windows uses `cmd.exe` and a `.sh` launcher won't run). These installers print manual setup instructions instead of writing `hooks.toml` — same pattern as Codex-on-Windows. Skills install cross-platform on all three.
- Skills: `plannotator-review`/`-annotate`/`-last` (origin baked in, from `apps/vibe/skills`) + the single-sourced `plannotator` knowledge skill from `apps/skills/core`. Stale-skill cleanup and three-state "VIBE USERS" banners in each script.

**Repo artifacts** (`apps/vibe/`)
- `hooks/hooks.toml` (managed block template), `hooks/config-snippet.toml`, three Vibe SKILL.md files. Added `apps/vibe` to the installer sparse-checkout.

**Docs + freshness guard**
- `AGENTS.md` env table: `mistral-vibe` in `PLANNOTATOR_ORIGIN` values, new `PLANNOTATOR_SKIP_VIBE_INSTALL` row.
- `packages/server/index.ts` supported-values comment. `apps/skills/core/plannotator/SKILL.md` `PLANNOTATOR_ORIGIN` row. Marketing env-vars doc (also fixed a pre-existing missing `oh-my-pi`).

## Out of scope

- **Mistral Ask AI provider**: not added to `packages/ai/`. Vibe falls back to the saved/server default. Adding a Mistral provider is a separate, larger effort.
- **Pi server mirror**: the gate runs the Bun binary (Vibe invokes `plannotator`), so `apps/pi-extension/server/` doesn't need a Vibe branch for the gate. annotate-last discovery lives Bun-side; vendoring to Pi only matters if Pi gains a Vibe annotate-last path.
- **Windows hooks**: skipped (Vibe runs hooks via `cmd.exe`; a `.sh` launcher isn't executable). Windows installers print manual instructions.

## Test plan

- [x] `bun test scripts/install.test.ts` — 124 pass (updated hardcoded assertions + new Vibe detection/integration test mirroring the Kiro block)
- [x] `bun test apps/hook/server/plannotator-skill-reference.test.ts` — freshness guard, 6 pass
- [x] `bun test apps/hook/server/session-log.test.ts` — 98 pass (includes Vibe session discovery, cwd-filter fix)
- [x] `bun test apps/hook/server/vibe-plan.test.ts apps/hook/server/vibe-session.test.ts` — 16 pass (resolver newest-by-mtime, empty-dir fail-open, gate allow/deny)
- [x] `bun run --cwd apps/review build && bun run build:hook` — both succeed
- [ ] Manual: with a plan in `~/.vibe/plans/`, pipe a Vibe `pre_tool`/`exit_plan_mode` JSON to `PLANNOTATOR_ORIGIN=mistral-vibe plannotator` — assert allow/deny on stdout
- [ ] Manual: from a cwd with a Vibe session, `PLANNOTATOR_ORIGIN=mistral-vibe plannotator last` opens the last assistant message from `messages.jsonl`


backnotprop:
Thanks for the patience on this one - we gave it a full adversarial review including sandboxed installer runs, live gate smoke tests, and a read of current Vibe source. The shape of the contribution is genuinely strong: the origin gating discipline, three-state installer reporting, managed-marker idempotency, and honest fail-open all mirror our house conventions closely. And the answer to our main question is clean: **every non-Vibe path is provably untouched** - the merged tree passes our full suite 4446/0 and the only shared-line edit is an inert `!isVibe` guard.

What blocks merge is that the Vibe integration itself has gone stale against your own upstream, plus a few installer defects we reproduced:

**1. Vibe upstream moved two days after you filed.** Hooks graduated (`enable_experimental_hooks` no longer exists - and on older versions that had the flag, hook types were `before_tool`/`after_tool`, so the flag + `pre_tool` combination was never valid on any single version), and 2.25.1's shell-free hook execution means `command = "PLANNOTATOR_ORIGIN=mistral-vibe plannotator"` will be exec'd with the env assignment as argv[0] on the unified-harness backend - spawn failure, gate silently dead. Suggested shape: drop the config.toml leg entirely (with a version-floor note), make the hook command argv-only with an absolute binary path (our Codex leg's ${PLANNOTATOR_BIN} pattern - note your heredoc is currently quoted so it can't interpolate), and detect Vibe from the payload (`hook_event_name=pre_tool` + `tool_name=exit_plan_mode` is unambiguous vs Claude/Gemini), keeping PLANNOTATOR_ORIGIN as the documented override. You're best positioned to verify against a live current build.

**2. The config.toml merge writes the key inside the last TOML table** (reproduced: appended after `[providers.mistral]`, tomllib parses it as `providers.mistral.enable_experimental_hooks`) - moot if the leg is dropped per above.

**3. Uninstaller coverage**: `packages/server/uninstall.ts` handles Codex/Gemini/Kiro/OpenCode homes but not Vibe, so `plannotator uninstall` would orphan the managed hooks.toml block (invoking a deleted binary on every exit_plan_mode), the skill dirs, and the config line. The managed markers make clean removal easy.

**4. The mtime gate** - we read your side's source and the timing assumption is right (the PLAN profile allowlists writes to PLANS_DIR during planning), but the plans dir is global, so two concurrent sessions can cross-serve plans, and with no freshness window a stale yesterday-plan reviews as if current when a write failed. The `pre_tool` payload already carries `transcript_path`: scanning it for the last write/edit targeting `plans/*.md` and gating that file (falling back to newest-by-mtime within a ~10min window, else fail open) eliminates both hazards with no Vibe-side change.

**Smaller items**: the annotate-last index-absent fallback picks the newest session across all projects (each session's meta.json carries cwd - filter on it, matching our Droid/Codex correct-session guarantees); `vibe-plan.test.ts` mutates `process.env.VIBE_HOME` without afterEach restore (house testing rule - `vibe-session.test.ts` has the right pattern); `command -v vibe` alone will create ~/.vibe on machines with an unrelated `vibe` binary (require the dir for write legs); the awk marker-strip can swallow trailing TOML when an end marker is missing (its comment claims the opposite); Windows skills carry POSIX env-prefix syntax (moot with the argv-only fix); docs rows for `--skip-vibe`/`PLANNOTATOR_SKIP_VIBE_INSTALL`.

Full details on any of these on request. It merges cleanly into current main and nothing shipped since needs extra wiring, so once the installer/hook shape is current with Vibe 2.25+, this lands fast. Looking forward to it - first-class Vibe support is something we want.

Djiit:
Thanks for the detailed review. All items are addressed in e7b6d92e. Before writing any of it, I checked your upstream claims against a live Vibe 2.25.3 install by reading `vibe/core/hooks/` from the installed package.

## Upstream drift (items 1 and 2)

You were right that the config leg is dead: `enable_experimental_hooks` has zero hits in the current package, and hooks load unconditionally from `hooks.toml`. I dropped the config.toml leg entirely and added a Vibe 2.25 version-floor note.

One correction from my side: the 2.25.3 build I inspected runs hooks through `asyncio.create_subprocess_shell`, not shell-free. I still made the command argv-only with the absolute binary path (`${PLANNOTATOR_BIN}`, the Codex leg's pattern; the heredoc is now unquoted so it actually interpolates), because an argv-only command is correct under both executors and I could not verify the shell-free backend you saw.

Origin detection now comes from the payload: `hook_event_name === "pre_tool"` plus `tool_name === "exit_plan_mode"` reads as Vibe and nothing else does, given Claude's `PermissionRequest` and Gemini's payload shape. `PLANNOTATOR_ORIGIN=mistral-vibe` stays as the documented override only. The config.toml merge bug is moot with the leg gone.

## The mtime gate (item 4)

The resolver now pins the plan from `transcript_path`: it scans `messages.jsonl` backward for the last `write_file` or `edit` tool result whose target sits under the plans dir (`file_path` for write_file, `file` for edit, both shapes confirmed in a real transcript) and gates that file. The fallback is newest-by-mtime inside a 10-minute window; anything older fails open so a failed write can't serve yesterday's plan. The freshness window only guards the fallback path, so no Vibe-side change is needed.

## annotate-last (smaller item)

The index-absent fallback now reads each candidate session's `meta.json` and filters on `environment.working_directory`, so a newer session from another project can never win. A cwd with no matching session returns null.

## Uninstaller (item 3)

`plannotator uninstall` now strips the managed block from `$VIBE_HOME/hooks.toml` by markers. A user's own hooks in the same file survive, and the file is deleted only when nothing recognizable remains. It also removes the four `$VIBE_HOME/skills/plannotator*` dirs. `uninstall.test.ts` covers this with a fixture holding both a user hook and the managed block.

## Remaining smaller items

- `command -v vibe` alone no longer creates `~/.vibe`: detection requires the dir (or `VIBE_HOME`), and only then do the write legs run.
- The awk marker-strip comment now states the actual behavior: a lone start marker swallows to EOF. That is the corrupt-input path, not a normal one.
- `vibe-plan.test.ts` saves and restores `VIBE_HOME` in `beforeEach`/`afterEach`.
- `PLANNOTATOR_SKIP_VIBE_INSTALL` row added to the marketing env-vars doc; the AGENTS.md row is rewritten for the new hook shape.
- The two red CI jobs are EMFILE/recursive-watch flakes in annotate-server tests, untouched by this branch. The rest of the suite passes locally; the only local failures are pre-existing `packages/server` Pi-mirror tests plus a pwsh guard that needs PowerShell on PATH.

## Test plan

- `bun test apps/hook/server/` — 259 pass, including the new transcript-pinning, freshness-window, and session-cwd tests
- `bun test scripts/install.test.ts` — Vibe assertions updated for the argv-only command and now assert the config leg is gone; only the pre-existing local pwsh guard fails off CI
- `bun test packages/server/uninstall.test.ts` — 43 pass with the new Vibe fixture
- `bun run --cwd apps/review build && bun run build:hook` — both succeed
- Manual: piped a synthetic `pre_tool`/`exit_plan_mode` payload (no `PLANNOTATOR_ORIGIN` set) to the server with `VIBE_HOME` pointed at a fixture home. Deny via the API produced `{"decision":"deny","reason":...}` on stdout, approve produced `{"decision":"allow"}`, and an empty plans dir failed open with empty stdout and exit 0.
- Manual: ran the installer's `write_vibe_hooks_block` in a sandboxed `HOME` twice against a file holding a user hook. The user hook survived, the block stayed single, and `tomllib` parses the result.

Djiit:
CI note: the workflow runs on this push sit in GitHub's awaiting-approval state (`action_required`, zero jobs started), so the checks page shows nothing yet for e7b6d92e. The runs need a maintainer approval click to start. Nothing on my side blocks them.

Djiit:
Thanks for the detailed review. All items are addressed in e7b6d92e. Before writing any of it, I checked your upstream claims against a live Vibe 2.25.3 install by reading `vibe/core/hooks/` from the installed package.

## Upstream drift (items 1 and 2)

You were right that the config leg is dead: `enable_experimental_hooks` has zero hits in the current package, and hooks load unconditionally from `hooks.toml`. I dropped the config.toml leg entirely and added a Vibe 2.25 version-floor note.

One correction from my side: the 2.25.3 build I inspected runs hooks through `asyncio.create_subprocess_shell`, not shell-free. I still made the command argv-only with the absolute binary path (`${PLANNOTATOR_BIN}`, the Codex leg's pattern; the heredoc is now unquoted so it actually interpolates), because an argv-only command is correct under both executors and I could not verify the shell-free backend you saw.

Origin detection now comes from the payload: `hook_event_name === "pre_tool"` plus `tool_name === "exit_plan_mode"` reads as Vibe and nothing else does, given Claude's `PermissionRequest` and Gemini's payload shape. `PLANNOTATOR_ORIGIN=mistral-vibe` stays as the documented override only. The config.toml merge bug is moot with the leg gone.

## The mtime gate (item 4)

The resolver now pins the plan from `transcript_path`: it scans `messages.jsonl` backward for the last `write_file` or `edit` tool result whose target sits under the plans dir (`file_path` for write_file, `file` for edit, both shapes confirmed in a real transcript) and gates that file. The fallback is newest-by-mtime inside a 10-minute window; anything older fails open so a failed write can't serve yesterday's plan. The freshness window only guards the fallback path, so no Vibe-side change is needed.

## annotate-last (smaller item)

The index-absent fallback now reads each candidate session's `meta.json` and filters on `environment.working_directory`, so a newer session from another project can never win. A cwd with no matching session returns null.

## Uninstaller (item 3)

`plannotator uninstall` now strips the managed block from `$VIBE_HOME/hooks.toml` by markers. A user's own hooks in the same file survive, and the file is deleted only when nothing recognizable remains. It also removes the four `$VIBE_HOME/skills/plannotator*` dirs. `uninstall.test.ts` covers this with a fixture holding both a user hook and the managed block.

## Remaining smaller items

- `command -v vibe` alone no longer creates `~/.vibe`: detection requires the dir (or `VIBE_HOME`), and only then do the write legs run.
- The awk marker-strip comment now states the actual behavior: a lone start marker swallows to EOF. That is the corrupt-input path, not a normal one.
- `vibe-plan.test.ts` saves and restores `VIBE_HOME` in `beforeEach`/`afterEach`.
- `PLANNOTATOR_SKIP_VIBE_INSTALL` row added to the marketing env-vars doc; the AGENTS.md row is rewritten for the new hook shape.
- The two red CI jobs are EMFILE/recursive-watch flakes in annotate-server tests, untouched by this branch. The rest of the suite passes locally; the only local failures are pre-existing `packages/server` Pi-mirror tests plus a pwsh guard that needs PowerShell on PATH.

## Test plan

- `bun test apps/hook/server/` — 259 pass, including the new transcript-pinning, freshness-window, and session-cwd tests
- `bun test scripts/install.test.ts` — Vibe assertions updated for the argv-only command and now assert the config leg is gone; only the pre-existing local pwsh guard fails off CI
- `bun test packages/server/uninstall.test.ts` — 43 pass with the new Vibe fixture
- `bun run --cwd apps/review build && bun run build:hook` — both succeed
- Manual: piped a synthetic `pre_tool`/`exit_plan_mode` payload (no `PLANNOTATOR_ORIGIN` set) to the server with `VIBE_HOME` pointed at a fixture home. Deny via the API produced `{"decision":"deny","reason":...}` on stdout, approve produced `{"decision":"allow"}`, and an empty plans dir failed open with empty stdout and exit 0.
- Manual: ran the installer's `write_vibe_hooks_block` in a sandboxed `HOME` twice against a file holding a user hook. The user hook survived, the block stayed single, and `tomllib` parses the result.

Djiit:
Hey @backnotprop , I just fixed some conflicts. I'd appreciate a review on the recent changes when you  get a chance, thank you 🙏

backnotprop:
Merged, thank you @Djiit, and thanks for your patience through two rounds of review. Before merging we added a few follow-ups on top of your commits: `plannotator uninstall --dry-run` no longer rewrites the Vibe hooks file and the real run reports it; the Windows installers only treat Vibe as present when its home folder exists and don't install the POSIX-only Vibe skills there (docs say `/plannotator-last` isn't supported for Vibe on Windows yet); the plans-folder check compares real paths so a symlinked `VIBE_HOME` can't pick another session's plan; the VIBE USERS installer section only prints when Vibe is detected; and docs and the `skipInstall.vibe` type are updated. Ships in the next release.

PR #1488, opened 2026-09-10
fix(ui): prevent skill autocomplete menu clipping

leoreisdias:
## Problem

Typing `/` or `$` in the annotate comment composer opens the skill autocomplete inside the composer card. The menu is positioned beyond that card, but the card's scrolling overflow clipped every row outside its bounds.

## Evidence

The failure reproduced in the README annotation flow: only the row touching the card edge remained visible. A focused DOM regression test also identified the comment card as the menu's clipping ancestor.

dyn-6473746fcb207c4edbefcf21f7e3b2b1

## Fix

Allow visible overflow only while the skill menu is open, in both compact popover and expanded dialog modes. When the menu closes, the composer restores its existing scrolling and clipping behavior.

## Review Focus

- Confirm the conditional overflow preserves long-comment scrolling after the menu closes.
- Confirm both anchored and dragged popovers, plus the expanded dialog, expose the complete menu.

## Test Plan

- [x] `DOM_TESTS=1 bun test packages/ui/components/SkillReferenceMenu.placement.test.tsx` — 12 tests passed
- [x] `packages/ui/node_modules/.bin/tsc --noEmit -p packages/ui/tsconfig.json`
- [x] Rebuilt the hook bundle and manually verified `$` autocomplete through the repository CLI against `README.md`
- [x] `git diff --check`

### DIFF
--- a/packages/ui/components/CommentPopover.tsx
+++ b/packages/ui/components/CommentPopover.tsx
@@ -529,6 +529,7 @@ export const CommentPopover: React.FC<CommentPopoverProps> = ({
     hasUnsavedContent ||
     (allowEmptySubmit && initialText.trim().length > 0);
   const canAskAI = !!onAskAI && !askAIDisabled && text.trim().length > 0;
+  const showsSkillMenu = skillAc.menu !== null;
 
   // Shared by both footers. Disabled once anything is typed or attached so a
   // click can never discard a draft; with content present, Save is the path.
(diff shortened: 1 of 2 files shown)


backnotprop:
Reviewed in depth before merge, including empirically: built both main and this branch, drove real annotate sessions, and confirmed the before/after in BOTH composer modes - popover (1-2 clipped rows → full menu) and the expanded dialog, which the unit test can't reach (happy-dom sees only inline styles; your placement test is real for the popover half - it fails against main's component - and our browser run covered the dialog half). Root cause confirmed as the #1297 mobile-safe overflow work; the fix is byte-identical whenever the menu is closed, the Esc ladder holds (menu first, composer second; the bare-$ single-Esc fall-through matches the shipped hook contract on main), and vim's blocking-overlay selector is untouched.

One accepted edge worth knowing: while the menu is open the composer's scroll containers stop scrolling, so very long comment drafts in tight anchor space can transiently spill past the card border - menu-open-only, self-healing, strictly better than an unusable menu. A future portal-based menu would remove the tradeoff; not needed now. Merging - thanks for another clean one.

PR #1490, opened 2026-09-10
fix(shared): resolve data-dir paths per call instead of capturing at import

### LINKED ISSUE #1477: resolve the data directory per call instead of capturing it at import
Four modules capture `getPlannotatorDataDir()` at import time and derive their paths from that value, so they use whatever the environment variable was set to when the module first loaded. Every other consumer resolves it per call.

- `packages/shared/improvement-hooks.ts`: `DATA_DIR`, plus `HOOKS_BASE_DIR` and `LEGACY_BASE_DIR`
- `packages/server/codex-review.ts`: `DATA_DIR`, `DEBUG_LOG_PATH`, `SCHEMA_DIR`, `SCHEMA_FILE`
- `packages/server/tour/tour-review.ts`: `TOUR_SCHEMA_DIR`, `TOUR_SCHEMA_FILE`
- `packages/server/guide/guide-review.ts`: `GUIDE_SCHEMA_DIR`, `GUIDE_SCHEMA_FILE`

These constants are module-private, so no signatures change. `vendor.sh` copies `improvement-hooks.ts` into the Pi extension. Nothing outside tests sets `PLANNOTATOR_DATA_DIR` after startup, so production behavior is unchanged.

## Motivation

`storage.ts` captured the value the same way until https://github.com/backnotprop/plannotator/pull/1437. There it broke annotate's submission-record tests on Linux runners, and the initial diagnosis attributed the failure to path normalization in an unrelated feature.

`embedded.ts` loads `@plannotator/server` with `await import(...)`. That call ran during a test that had set `PLANNOTATOR_DATA_DIR` to its own temp directory, so `storage.ts` captured that directory and used it for the remainder of the process.

`packages/server/index.ts` imports `improvement-hooks.ts`, so the same dynamic import loads it under the same conditions. The other three modules are imported by `review.ts`, which that dynamic import does not reach. 38 test files set the variable at runtime, and Bun's test-file order is filesystem-dependent, so the captured value varies as test files are added.

## Approach

Replacing `DATA_DIR` alone is insufficient. `DEBUG_LOG_PATH`, `SCHEMA_FILE`, `TOUR_SCHEMA_FILE`, and `GUIDE_SCHEMA_FILE` are derived at module scope and stay fixed.

A regression test must import the module before setting the environment variable, because the reverse order passes with or without the fix. `packages/shared/storage.test.ts` is the model. `improvement-hooks.test.ts` currently spawns a subprocess with a modified `HOME` for the same reason, and can set the variable in-process after this change.

FNDEVVE:
## Problem

Four modules evaluate `getPlannotatorDataDir()` at module scope and freeze every path derived from it, so they answer to whatever `PLANNOTATOR_DATA_DIR` was when the module first loaded — the same import-time capture `storage.ts` had until #1437/#1473, with the same consequences: 38 test files set the variable at runtime and Bun's test-file order is filesystem-dependent, so the captured value varies as test files are added.

- `packages/shared/improvement-hooks.ts` — `DATA_DIR`, `HOOKS_BASE_DIR`, `LEGACY_BASE_DIR`
- `packages/server/codex-review.ts` — `DATA_DIR`, `DEBUG_LOG_PATH`, `SCHEMA_DIR`, `SCHEMA_FILE`
- `packages/server/tour/tour-review.ts` — `TOUR_SCHEMA_DIR`, `TOUR_SCHEMA_FILE`
- `packages/server/guide/guide-review.ts` — `GUIDE_SCHEMA_DIR`, `GUIDE_SCHEMA_FILE`

## Change

All of these now resolve per call. The `*SchemaMaterialized` booleans became per-path existence checks, so a data-dir switch after import materializes the schema in the new location instead of returning the old path forever. `CODEX_REVIEW_SCHEMA_PATH` (a re-export of the frozen string; no in-repo consumers) is replaced by `getCodexReviewSchemaPath()`.

## Verification

Bun 1.3.14, disposable HOME + temp `PLANNOTATOR_DATA_DIR`:

- Regression tests import each module **before** mutating the env var (the reverse order passes with or without the fix), modeled on `storage.test.ts`: `improvement-hooks.test.ts` is rewritten in-process (the subprocess/fake-HOME workaround is no longer needed) and `codex-command-builders.test.ts` covers all three schema modules, including re-materialization after a mid-test dir switch.
- Stash-the-fix control: with the four source edits stashed and the tests kept, 7 of the new assertions fail; restored, everything passes.
- `88 pass / 0 fail` across the 8 touched-area test files.
- Pi vendoring re-run (`apps/pi-extension/vendor.sh`) and smoke-imported with the env set after import — the generated copies follow the new data dir (they are gitignored build output, so there is nothing to commit).

Deliberately out of scope: `packages/server/browser.ts` (`IPC_REGISTRY`) keeps its module-scope capture — same pattern, separate surface.

Fixes #1477

### DIFF
--- a/packages/server/codex-command-builders.test.ts
+++ b/packages/server/codex-command-builders.test.ts
@@ -6,15 +6,15 @@
  */
 
 import { afterEach, beforeEach, describe, expect, test } from "bun:test";
-import { existsSync, readFileSync } from "node:fs";
+import { existsSync, readFileSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 
 // Imported statically and BEFORE any PLANNOTATOR_DATA_DIR is set: these modules
 // used to capture the data directory at import, so changing the env var later
 // left every derived path (and the materialized schema) in the old location.
-import { buildCodexCommand } from "./codex-review";
-import { buildGuideCodexCommand } from "./guide/guide-review";
-import { buildTourCodexCommand } from "./tour/tour-review";
+import { buildCodexCommand, CODEX_REVIEW_SCHEMA } from "./codex-review";
+import { buildGuideCodexCommand, GUIDE_SCHEMA_JSON } from "./guide/guide-review";
+import { buildTourCodexCommand, TOUR_SCHEMA_JSON } from "./tour/tour-review";
 
 import { createTestEnvironment } from "../../tests/helpers/environment";
 
@@ -82,5 +82,28 @@ describe("Codex command builders", () => {
 
     expect(schema).toBe(join(second, "codex-review-schema.json"));
     expect(existsSync(schema)).toBe(true);
+  });
+
+  test("overwrite a stale schema file left by an older binary", async () => {
+    // A schema written by an older version persists in the data dir forever
+    // (nothing prunes it). An existence check would keep serving those stale
+    // bytes; every process must refresh the file with its own schema once.
+    const dataDir = process.env.PLANNOTATOR_DATA_DIR!;
+    const stale = '{"stale":"written by an older binary"}';
+    const surfaces = [
+      [buildCodexCommand, "codex-review-schema.json", CODEX_REVIEW_SCHEMA],
+      [buildGuideCodexCommand, "guide-schema.json", GUIDE_SCHEMA_JSON],
+      [buildTourCodexCommand, "tour-schema.json", TOUR_SCHEMA_JSON],
+    ] as const;
+
+    for (const [build, fileName, currentSchema] of surfaces) {
+      const path = join(dataDir, fileName);
+      writeFileSync(path, stale);
+
+      const command = await build(OPTIONS);
+
+      expect(schemaPath(command)).toBe(path);
+      expect(readFileSync(path, "utf-8")).toBe(currentSchema);
+    }
   });
 });
--- a/packages/server/codex-command-builders.test.ts
+++ b/packages/server/codex-command-builders.test.ts
@@ -1,24 +1,86 @@
-import { describe, expect, test } from "bun:test";
+/**
+ * Schema materialization for the Codex output-schema flag, across all three
+ * review surfaces.
+ *
+ * Run: bun test packages/server/codex-command-builders.test.ts
+ */
+
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { existsSync, readFileSync } from "node:fs";
+import { join } from "node:path";
+
+// Imported statically and BEFORE any PLANNOTATOR_DATA_DIR is set: these modules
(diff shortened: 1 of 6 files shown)


backnotprop:
Deep review done. The sweep itself is solid: all four #1477 modules fully converted with no surviving derived-path constants, no consumer relied on freezing, vendoring clean, and per-module revert-verification confirms every module has a test that fails if it regresses (11/11 restored). Honest scoping too - browser.ts called out rather than smuggled.

One fix needed before merge: **the `schemaMaterialized` → `existsSync` substitution changes schema-refresh semantics**. The old flag meant "rewrite once per process", so every process start refreshed the schema files with the current binary's schema. `existsSync` writes only when absent - so a schema written by an older version persists forever (nothing prunes the data dir). Not hypothetical: #997 added a required `summary` field to the guide schema; replay that upgrade under this PR and Codex is `--output-schema`-forced into the old shape indefinitely, silently degrading guides. Suggested replacement preserving both properties (fresh per process AND data-dir switches):

```ts
const materializedSchemaPaths = new Set<string>();
async function ensureSchemaFile(): Promise<string> {
  const schemaPath = getCodexReviewSchemaPath();
  if (!materializedSchemaPaths.has(schemaPath)) {
    await mkdir(dirname(schemaPath), { recursive: true });
    await writeFile(schemaPath, CODEX_REVIEW_SCHEMA);
    materializedSchemaPaths.add(schemaPath);
  }
  return schemaPath;
}
```

Same shape in all three modules (using `dirname(schemaPath)` also removes the double `getPlannotatorDataDir()` call per materialization). Worth one added assertion: overwrite a schema file with stale bytes, call the builder, expect current bytes.

We'll file the follow-up for the one remaining frozen capture (`packages/server/browser.ts:13` IPC_REGISTRY) on our side. Merge follows as soon as the Set-based materialization lands.

backnotprop:
Pushed `75326e3b` with the Set-based materialization from the review finding: all three modules (`codex-review.ts`, `tour/tour-review.ts`, `guide/guide-review.ts`) now guard schema writes with a per-path written-this-process Set instead of `existsSync`, so every process refreshes the schema files with the current binary's schema (restoring the old once-per-process semantics) while a `PLANNOTATOR_DATA_DIR` change after import still materializes into the new location. `dirname(schemaPath)` replaces the second `getPlannotatorDataDir()` call per materialization.

Added the requested regression assertion in `codex-command-builders.test.ts`: overwrite each schema file with stale bytes, call the builder, expect the current schema bytes — verified it fails under the `existsSync` guard and passes with the Set. `bun test packages/server packages/shared` 1890 pass / 0 fail, vendoring clean, typecheck clean.

PR #1471, opened 2026-09-05
fix(ui): keep selection toolbar within compact touch viewport

katya4oyu:
## Summary

Fix horizontal overflow of the text-selection toolbar when selecting short text near a screen edge in compact touch layouts.

- Use the existing compact-touch marker and viewport/safe-area CSS variables with `translate: clamp(...)` to constrain horizontal placement. Keep the existing centered position when it fits.
- Preserve vertical positioning, display timing, animations, selection handling, and button layout.
- Leave desktop placement and the right-aligned code-block toolbar unchanged.

## Testing

- Added regression tests for the existing placement/animation declarations and for keeping right-aligned toolbars outside the correction.
- Passed 8 related DOM tests, the UI typecheck, and review/hook builds.
- Checked left- and right-edge overflow correction in Chromium and WebKit at a 390px touch viewport, and unchanged placement at 1200px and 390px desktop viewports.
- Verified display and interaction on a physical iPhone.

The existing happy-dom environment does not calculate CSS layout, so automated layout assertions are not included. Happy to add them using the project's preferred browser-testing setup if desired.

PR #1491, opened 2026-09-10
fix(review): keep the annotation toolbar inside the viewport when suggested code expands

### LINKED ISSUE #1424: No room to add suggested code to last lines on the file
### What happened?

The add comment button cannot be seen.

https://github.com/user-attachments/assets/72f29f9f-8834-4e2e-b755-441bc2c407b5

### Plannotator version

0.27.9

### OS

macOS

### Agent

Pi

### Where did it happen?

Plan review

FNDEVVE:
## Problem

Selecting lines near the bottom of a diff and expanding **Add suggested code** pushed the submit row off-screen: the toolbar's vertical clamp reserved a hardcoded `visibleBounds.bottom - 200` while `maxHeight` was set to the *entire* viewport height, so internal scrolling could never engage and the grown toolbar simply extended past the bottom edge. Both the anchored and the dragged branch had the flaw, and Ask AI mode renders inside the same container. #1424

## Change

The toolbar measures its real height (`useLayoutEffect` after every commit, so the suggested-code toggle, Ask AI history growth and manual textarea resizes all land) and a shared placement helper keeps the whole box — submit row included — inside `visibleBounds`:

- anchored toolbars flip above the anchor when they would not fit below;
- a toolbar taller than the viewport is pinned to the top with a bounded `maxHeight` (the only case where internal scroll engages);
- dragged toolbars keep the user's chosen position, clamped into bounds.

Horizontal clamping is untouched.

## Verification

Bun 1.3.14, `DOM_TESTS=1`, disposable HOME: `bun test packages/review-editor/components/AnnotationToolbar.placement.test.tsx` → **13 pass / 0 fail** (previously 4 pass / 4 skip). Pre-fix control with the component change stashed: 8 pass / 5 fail (flip-above in both Comment and Ask AI modes, oversized pin in both branches, dragged-expand) — the new tests fail on the old code. Coverage: flip-above when expanded at the bottom edge in both modes, no move when it fits below, top-edge clamp, oversized pin + height clamp, and a real pointer-drag → expand sequence.

Fixes #1424

### DIFF
--- a/packages/review-editor/components/AnnotationToolbar.placement.test.tsx
+++ b/packages/review-editor/components/AnnotationToolbar.placement.test.tsx
@@ -6,8 +6,20 @@ const hasDom = typeof document !== 'undefined';
 const toolbarModule = hasDom ? await import('./AnnotationToolbar') : null;
 const AnnotationToolbar = toolbarModule?.AnnotationToolbar as typeof import('./AnnotationToolbar')['AnnotationToolbar'];
 
+/** Laid-out height of the toolbar. happy-dom does no layout, so we stub the
+ * height the component measures: the collapsed composer vs. the same toolbar
+ * with the suggested-code section expanded. */
+const COLLAPSED_TOOLBAR_HEIGHT = 236;
+const EXPANDED_TOOLBAR_HEIGHT = 360;
+let toolbarHeight = EXPANDED_TOOLBAR_HEIGHT;
+
+const originalScrollHeight = hasDom
+  ? Object.getOwnPropertyDescriptor(HTMLElement.prototype, 'scrollHeight')
+  : undefined;
+
 let host: HTMLElement | null = null;
 let root: Root | null = null;
+let toolbarRef: React.RefObject<HTMLDivElement | null> | null = null;
 let originalMatchMedia: typeof window.matchMedia | undefined;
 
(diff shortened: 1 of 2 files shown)


backnotprop:
Reviewed in depth before merge, empirically: built main and this branch, drove a real review session to the bottom of a 300-line diff, expanded Add suggested code, and measured - on main the Add Comment row lands at bottom 655 in a 560px viewport (off-screen, matching the #1424 video); on this branch the toolbar flips above the anchor and everything is reachable. The no-change control at a roomy viewport is byte-identical to main, the #1471 compact-touch clamp from 0.27.13 is a different component family with zero overlap (its suite passes on your branch), and revert-verification confirms 5 of the 13 placement tests genuinely fail against main's component - including the real-pointer dragged path.

One deliberate secondary change we'll carry to release notes: a dragged toolbar is now fully clamped into the viewport, so it can no longer be parked partially off-screen. Two non-blocking notes: a native textarea resize-handle drag doesn't re-clamp until the next React commit (bounded by the textarea's own max-h; a ResizeObserver would close it if it ever bites), and TOOLBAR_ANCHOR_GAP silently mirrors the +10 anchor offset in useAnnotationToolbar - a cross-reference comment would keep them from drifting. Merging.

PR #1492, opened 2026-09-10
feat(editor): Mod+E shortcut to enter and leave Edit mode in place

### LINKED ISSUE #1479: [Feature Request] Keyboard shortcut to enter / leave Edit mode in place
First, thanks for Plannotator! ❤️ It has become part of my daily workflow. 
What follows is nothing more than a small comfort improvement for someone who lives in it every day.

## What I'd like

A keyboard shortcut to enter and leave **Edit mode** on a markdown document, in place, without scrolling the page.

Today the only way in and out is the `Edit` / `Done` control in the document card header:

- **Annotating**: the card header shows `Wide | Focus | Edit` at the top right ("Edit the document text directly").
- **Editing**: the same row becomes `Wide | Focus | Save • Cancel`.
- **After saving**: `Wide | Focus | Saved | Done` ("Commit your edits and return to annotating").

That row lives at the top of the document, so switching modes mid-document means scrolling all the way up, clicking, then scrolling back down and finding the paragraph again: once to edit, once more to go back to annotating. It breaks the flow of exactly the workflow Edit mode is best at, reading a long document and alternating between "annotate this / ask the agent about this" and "just fix this word myself".

`Mod+S` already covers saving in edit mode when the document is source-backed (`annotate` on a real local `.md` file, which is my case), so the save half of the loop is already keyboard-driven. It's only entering and leaving the mode that still needs the mouse and the scroll.

## Why this looks like a small gap rather than a new feature

`Focus` and `Edit` sit in the same control cluster on the document card, and `Focus` already has a binding: `Mod+.`, shipped in v0.27.0 (#1276), defined in `packages/ui/shortcuts/plan-review/documentView.shortcuts.ts`, in a `document-view` scope whose only member today is `toggleFocusMode`. The ask is essentially a sibling entry in that same scope. There is precedent for a mode toggle too: `Mod+Shift+A` toggles Interact/Annotate on HTML surfaces.

Also related but not the same: #1242 shipped keyboard switching for *annotation* modes, and those shortcuts are gated off while editing, so they don't cover this.

## Rough idea (very open to alternatives)

- One chord that toggles edit mode both ways, registered in the existing `document-view` scope. `Mod+E` looks free against `Mod+.`, `Mod+S`, `Mod+P`, `Mod+Enter` and `Mod+Shift+A`, but whatever fits the registry best is fine.
- **Scroll position must survive the toggle** in both directions. The Viewer → CodeMirror swap is a different DOM, so this probably needs saying explicitly: a chord that re-anchors the document to the top would not solve the problem it's meant to solve. Keeping the caret / selection across the swap would be a bonus, not a requirement.
- **On the exit semantics**, one question I can't answer from the outside: leaving edit mode isn't a single behavior. A clean document gets the plain `Done`, which commits the edits and exits. A source-backed document with unsaved changes gets the deliberate two-step `Cancel` → `Discard changes`. My assumption is that the chord should map to the `Done` path, and that it should *not* become a way to bypass that confirmation. What it should do while the two-step Cancel is armed is your call: no-op, or the same first press that arms the confirmation.

## Possible complement: keep the control reachable while scrolling

Secondary, and possibly a better fit than a shortcut for mouse users: make entering and leaving edit mode reachable anywhere in the document rather than only at the top, through a sticky or floating affordance, or a place in the persistent header chrome.

FNDEVVE:
## Problem

Edit mode was reachable only through the Edit/Done control at the top of the document card, so alternating between "annotate this" and "just fix this word" mid-document cost a scroll to the top and back — twice. `Mod+S` already covers saving in edit mode; entering and leaving the mode had no chord. #1479

## Change

- The `document-view` scope gains `toggleEditMode`, bound to `Mod+E` (unclaimed repo-wide; the sibling `toggleFocusMode` is `Mod+.`).
- **Enter** runs through the existing `handleEditToggle` behind the same `canEditMarkdown` gate the card's Edit control uses, so read-only/archive/HTML/live/plan-diff surfaces no-op.
- **Exit** works while CodeMirror has focus, via a window keydown listener mounted only while editing that routes to the card's Done path (`commitMarkdownEdits`). With unsaved source-backed changes (`cancelMode`) it no-ops — never a silent discard. The `isEditingMarkdown`/contentEditable refusal in `canHandleDocumentChromeShortcut` is untouched; the exit listener shares only the dialog/overlay half of the guard.
- **Scroll carry**: the offset is captured before the Viewer↔editor swap and re-applied after it (immediate + one rAF — CodeMirror settles its height a frame after mount, and a single write clamps to the top). New `packages/editor/editScroll.ts` makes explicit which surface actually scrolls: on desktop CodeMirror grows inside the document viewport, while a height-bounded compact shell scrolls internally.
- Settings → Shortcuts → View lists the new row (that list is hand-maintained, not registry-driven).

## Verification

- Bun 1.3.14, disposable HOME: `shortcuts.test.ts` **17/17**; `App.editModeShortcut.test.tsx` **4/4**; `editScroll.test.ts` **4/4**; combined regression run of 10 App/Settings files **46/46**. Each load-bearing piece (restore effect, exit listener, `cancelMode` guard) has a stash-the-fix red control.
- Real browser (Chrome for Testing over CDP against the hook dev server): viewer at scrollTop 3000 → `Mod+E` → editor mounted with the offset preserved; scrolled to 4321 and focus placed inside `.cm-content` → `Mod+E` → back in the viewer at the captured offset. The chord fires from inside CodeMirror; no keymap swallows it.
- guides-show gate in this tree: `build:viewer`, `sync:manifest`, `check:budgets`, `check:manifest` all pass, with no manifest drift.

Fixes #1479

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -188,6 +188,7 @@ jobs:
           packages/ui/components/ActionMenu.test.tsx
           packages/ui/components/PlanHeaderMenu.mobile.test.tsx
           packages/review-editor/components/ReviewHeaderMenu.mobile.test.tsx
+          packages/editor/App.editModeShortcut.test.tsx
 
   opencode-v2:
     name: OpenCode 2 installed package
--- a/packages/editor/App.editModeShortcut.test.tsx
+++ b/packages/editor/App.editModeShortcut.test.tsx
@@ -221,6 +221,30 @@ describe.if(hasDom)("Mod+E edit-mode toggle", () => {
     expect(document.querySelector(".cm-editor")).toBeNull();
   });
 
+  test("leaves the chord to a foreign text field, but still exits from CodeMirror", async () => {
+    await mount(ANNOTATE_PLAN);
+
+    await act(async () => { pressModE(document.body); });
+    const content = document.querySelector<HTMLElement>(".cm-editor .cm-content");
+    if (!content) throw new Error("CodeMirror content DOM did not render");
+
+    // Mod+E dispatched from a textarea outside the editor (Ask AI box, an
+    // annotation comment mid-edit-session) must stay the field's own chord:
+    // the session may not commit-and-exit underneath it.
+    const foreign = document.createElement("textarea");
+    document.body.appendChild(foreign);
+    try {
+      await act(async () => { pressModE(foreign); });
+      expect(document.querySelector(".cm-editor")).not.toBeNull();
+    } finally {
+      foreign.remove();
+    }
+
+    // From inside CodeMirror's contenteditable the exit still fires.
+    await act(async () => { pressModE(content); });
+    expect(document.querySelector(".cm-editor")).toBeNull();
+  });
+
   test("never discards an unsaved source-backed buffer silently", async () => {
     await mount({
       plan: "# Source document\n\nEditable body.\n",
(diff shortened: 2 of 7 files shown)


backnotprop:
Deep review done - the feature itself is well-built: correct scope placement (the docs page picks it up automatically), a genuinely clean collision check across all three surfaces (the only other Mod+E in the dependency tree is an opt-in atomic-editor extension this editor never mounts), the dirty-buffer refusal verified empirically with zero data loss, and revert-verified behavioral tests. Three fixes before merge:

1. **The new App tests never run in CI**: `packages/editor/App.editModeShortcut.test.tsx` is DOM-gated, and CI's DOM step is an explicit file enumeration this PR didn't extend - under default `bun test` all four silently skip. Add the file to the `DOM_TESTS=1 bun test --isolate` list in `.github/workflows/test.yml`.
2. **De-flake before adding it**: "never discards an unsaved source-backed buffer silently" fails ~3/8 runs under CI's exact invocation when preceded by another App suite - the single `setTimeout(0)` after mutating `content.textContent` doesn't always outlast CodeMirror's MutationObserver → React dirty-flip. Replace with a bounded poll for `saveLabel() === "Save"` (your own `mount()` already uses that pattern).
3. **Exit chord leaks into foreign text fields**: the dedicated exit listener drops the target check to serve CodeMirror's contenteditable, but that means Mod+E from the Ask AI box or an annotation textarea mid-edit-session commits and exits the editor (empirically confirmed), and on macOS steals native Cmd+E in those fields. Early-return when the target is an INPUT/TEXTAREA outside `.cm-editor`, and match the key case-insensitively (`event.key.toLowerCase()`) so Caps Lock doesn't break exit while enter still works.

Two non-blocking notes worth a comment in code: adding atomic-editor's `selectionToolbar()` to this editor would dead-key the exit chord (CM would claim Mod-e and preventDefault); and the inert-chord UX on a dirty buffer (no feedback on refusal - arming the visible Cancel → Discard? step instead would explain it).

Merge follows when the three land.

backnotprop:
Pushed the three review fixes in d038756a (maintainer edit on your branch):

1. **CI wiring** — `packages/editor/App.editModeShortcut.test.tsx` added to the `DOM_TESTS=1 bun test --isolate` enumeration in `.github/workflows/test.yml`.
2. **De-flake** — the dirty-buffer test no longer races CodeMirror's MutationObserver. Root cause found while de-flaking: under happy-dom, after CodeMirror's `stop()`/`start()` observer cycling the observer intermittently never re-delivers records at all (a fresh MutationObserver on the same node fires; CM's stays deaf for the whole session), so no poll length could fix the original DOM-mutation approach. The buffer is now dirtied through the editor's own `view.dispatch` (resolved via CM's DOM back-reference, both historical property names, loud failure on rename) with a bounded poll for the toolstrip's `Save` readout. 8/8 green under CI's exact invocation preceded by another App suite (was ~3/8 failing).
3. **Exit-chord scoping** — the dedicated Mod+E listener now early-returns for INPUT/TEXTAREA targets outside `.cm-editor` and matches the key case-insensitively; a code note warns that mounting atomic-editor's `selectionToolbar()` would claim Mod-e and dead-key the chord. New DOM test pins both halves (foreign textarea keeps the session; chord from inside CodeMirror exits), revert-verified — dropping the guard fails the test.

Verified: all PR DOM files + adjacent App suites green under `DOM_TESTS=1 bun test --isolate`, `bun test packages/ui/shortcuts.test.ts packages/editor` green, `bun run typecheck` clean.

PR #1493, opened 2026-09-10
fix(codex): fall back across rollout files for annotate-last and the Stop hook

### LINKED ISSUE #1367: Codex: `annotate-last` fails when one thread spans multiple rollout files
## Summary

Running `plannotator last` in Codex Desktop can fail with:

> No rendered assistant message found in session logs.

This happens when one Codex thread is represented by multiple rollout JSONL files. Plannotator may select an empty or aborted rollout even though another rollout for the same thread contains the previous rendered assistant message.

## Observed scenario

In the affected Codex thread, several rollout files contained the same thread ID:

- one rollout was aborted and contained no assistant `output_text`,
- another rollout contained the previous rendered assistant response,
- the active rollout contained the current command.

`plannotator last` selected the empty rollout and stopped without checking the other matching files.

Debug output, with identifiers redacted:

```text
[DEBUG] Codex detected, thread ID: <redacted>
[DEBUG] Rollout: ~/.codex/sessions/YYYY/MM/DD/rollout-...<thread-id>_<segment-id>.jsonl
No rendered assistant message found in session logs.
```

## Steps to reproduce
1. Open a task in Codex Desktop.
2. Continue or resume it until Codex creates multiple rollout files associated with the same thread ID.
3. Ensure one matching rollout is empty or aborted while another contains a completed assistant response.
4. Run plannotator last.
5. Plannotator may select the empty rollout and return the error above.

I do not yet have a deterministic sequence that always makes Codex split a thread into multiple rollout files, but this structure occurred in a real Codex Desktop task.

## Expected behavior

Plannotator should treat all rollout files belonging to the current Codex thread as one logical conversation and open the most recent completed assistant response before the active turn.

## Actual behavior

Plannotator selects one matching rollout file and fails immediately when that file does not contain a rendered assistant message, even though another rollout for the same thread does.

## Likely root cause

`findCodexRolloutByThreadId()` returns the first filename from `readdirSync()` that contains the thread ID:
`https://github.com/backnotprop/plannotator/blob/main/apps/hook/server/codex-session.ts`

The candidates are not ordered or inspected before one is selected. Therefore, when several rollout files match the same thread, filesystem iteration order determines which rollout is used.

### Plannotator version

plannotator 0.27.1

### OS

macOS

### Agent

Codex

### Where did it happen?

Annotate

FNDEVVE:
## Problem

One Codex thread can span several rollout JSONL files (`rollout-…-<thread-id>_<segment>.jsonl`), and `findCodexRolloutByThreadId` returned the **first** filename containing the thread id from an unsorted `readdirSync` — which can be an empty or aborted segment. `plannotator last` then exited with `No rendered assistant message found in session logs.` even though another rollout for the same thread contained the message. The Stop hook's plan detection had the same single-file assumption. #1367

## Change

- `findCodexRolloutsByThreadId` returns **all** rollouts matching the thread id, newest first (mtime, with the filename timestamp as tiebreak); the singular finder delegates to it.
- Both consumers iterate candidates newest-first and use the first rollout that actually yields the needed artifact — a rendered assistant message for annotate-last, plan content for the Stop hook.
- Single-rollout threads behave exactly as before, including the pre-existing not-found error path.

## Verification

Bun 1.3.14, disposable HOME: `apps/hook/server/codex-session.test.ts` gains 5 tests over synthetic `CODEX_HOME` fixture trees — same-day newer-empty/older-with-message, multi-day fallback, single-rollout unchanged (success and not-found), and the Stop-hook plan fallback. **64 pass / 0 fail** across `codex-session` + `strict-annotate-result` + `cli` test files. A throwaway repro script confirmed the first-candidate bug and the fix before deletion (sorting alone is insufficient: the newest segment may legitimately be the empty one — the fallback across candidates is the actual fix).

Fixes #1367

### DIFF
--- a/apps/hook/server/codex-session.test.ts
+++ b/apps/hook/server/codex-session.test.ts
@@ -7,10 +7,16 @@
  */
 
 import { describe, expect, test, afterEach } from "bun:test";
-import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from "node:fs";
+import { mkdtempSync, mkdirSync, writeFileSync, rmSync, utimesSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
-import { findCodexRolloutByThreadId, getLastCodexMessage, getLatestCodexPlan } from "./codex-session";
+import {
+  findCodexRolloutByThreadId,
+  findCodexRolloutsByThreadId,
+  getLastCodexMessage,
+  getLatestCodexPlan,
+  getRecentCodexMessages,
+} from "./codex-session";
 
 // --- Fixture Helpers ---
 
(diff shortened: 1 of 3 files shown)


backnotprop:
Deep review done. The annotate-last leg is correct and verified - same-thread-only fallback holds (full-UUID substring matching cannot cross threads), and `plannotator last` now recovers messages from older segments as intended.

The Stop-hook leg needs a fix before merge: **stale-plan resurrection**. `getLatestCodexPlan`'s turn gate protects plan-item events but NOT assistant `<proposed_plan>` messages, and `findTurnStartIndex` degrades to last-turn-in-file when the current turn_id is absent. So in a segmented thread whose older segment ends with a proposed plan - the normal shape after a plan is approved and the session resumes - every Stop event without `transcript_path` whose turn produced no plan falls back and deterministically reopens the already-decided plan review, blocking every turn end. The asymmetry: annotate-last asks a thread-level question so cross-segment fallback is right; the Stop hook asks a turn-level question, and the current turn cannot live in an older segment, so anything a fallback file yields there is stale by construction.

Fix options (any one): (a) Stop-hook leg takes only the first *existing* candidate, keeping your newest-first ordering - this alone fixes #1367's actual manifestation; (b) continue past a candidate only when `parseRolloutEntries()` is empty; (c) accept a fallback file's plan only when the event's turn_id appears in it. The test "Stop hook plan falls back to an older segment" currently pins the bug and should change with the fix - and it should pass the real `{turnId, stopHookActive}` options the production call site uses.

Two smaller items: the newest-first sort currently passes by filesystem readdir-order coincidence - add one fixture where mtime order contradicts filename order to genuinely pin it; and consider issue-realistic `_<segment>`-suffixed fixture filenames plus a 0-byte-file case. Also worth knowing: discovery now always walks the full sessions tree instead of early-returning - fine for most, measurable on years-old or networked `~/.codex/sessions`.

Happy to merge as soon as the Stop-hook fallback is narrowed.

backnotprop:
Pushed a maintainer-side commit (ac6bcd91) applying the fix from the [review above](https://github.com/backnotprop/plannotator/pull/1493#issuecomment-5626467752):

- **Stop-hook leg narrowed to the first existing candidate** (option a). The Stop hook asks a turn-level question and the current turn cannot live in an older segment, so it no longer falls back across segments — `getLatestCodexPlan` runs only on the newest existing rollout, which prevents the stale-plan resurrection (an older segment ending in an already-decided `<proposed_plan>` deterministically reopening the review on every turn end). The annotate-last leg keeps its cross-segment fallback unchanged — that one asks a thread-level question and is correct as submitted.
- **Test "Stop hook plan falls back to an older segment" replaced** with one pinning the new behavior (older segment's decided plan is NOT resurrected), plus a positive case (plan in the newest segment is still found). The mirror helper now passes the real `{ turnId, stopHookActive }` options the production call site uses. Revert-verified: restoring the old fallback makes the new test fail.
- **Newest-first sort genuinely pinned**: added a fixture where the newer-mtime segment has the lexicographically smaller name and sits in the earlier day directory, so directory-walk and readdir order both contradict mtime. Confirmed it fails with the sort neutered.
- Fixture filenames now use the issue-realistic `_<segment>` suffix, and a 0-byte-newest-segment case covers the aborted-segment fallback for annotate-last.

`bun test apps/hook/server/`: 248 pass / 0 fail; `bun run typecheck` clean. Merging-ready from my side.

PR #1494, opened 2026-09-10
fix(browser): honor POSIX PLANNOTATOR_BROWSER under WSL and warn on launch failure

### LINKED ISSUE #1472: PLANNOTATOR_BROWSER never invoked, and nothing says so
### What happened?

# Plannotator: `PLANNOTATOR_BROWSER` never invoked, and nothing says so.

**Tool:** [backnotprop/plannotator](https://github.com/backnotprop/plannotator) — reproduced on
0.27.8, 0.27.9 and **0.27.12** (current at time of writing) · Apache-2.0 · **Platform:** Linux
x86_64, WSL2 (Ubuntu), no display, no `xdg-open`

## Summary

When plannotator is launched **without a controlling terminal** — which is how an agent or a CI
process launches it — the executable named by `PLANNOTATOR_BROWSER` is never invoked, and nothing
reports that it was skipped. The server starts correctly and the ready file is written, so every
observable signal reports success while the reviewer is never actually shown the review.

A browser that cannot be opened is a reasonable thing to decline to do; declining without a word
makes it unresolvable from the outside. We spent a debugging session confirming that the extension
router existed, the IPC port was live, and the environment was correctly set — all true, all
irrelevant, because the invocation was never attempted.

**Narrowed on 2026-09-05, re-verifying against 0.27.12.** An earlier draft of this report said no
message was written at all. That is no longer accurate and the correction is upstream's credit:
0.27.12 prints `Plannotator session ready — open in your browser:` followed by the URL, so a human
watching a terminal is no longer stranded. What remains is narrower and still a defect: when
`PLANNOTATOR_BROWSER` is **explicitly set**, it is ignored without a word. The caller has stated
which browser to use, and the honest outcomes are to invoke it or to say why it was not invoked.
Printing a URL instead helps an interactive user and does nothing for the detached, agent-launched
case this report is about, because there is no terminal reading that line.

## Reproduction

Verified 2026-08-29 against 0.27.8 and **re-verified 2026-09-05 against 0.27.12**, unchanged.
`setsid` is the whole trick: it detaches from the controlling terminal, reproducing how any
agent-launched process runs.

```bash
mkdir repro && cd repro
printf '#!/usr/bin/env bash\necho "INVOKED args=$*" >> "$(dirname "$0")/browser.log"\n' > fake-browser
chmod +x fake-browser
echo '# repro' > target.md

PLANNOTATOR_PORT=4497 \
PLANNOTATOR_READY_FILE="$PWD/ready" \
PLANNOTATOR_DATA_DIR="$PWD/data" \
PLANNOTATOR_BROWSER="$PWD/fake-browser" \
setsid plannotator annotate target.md --gate --json --result-file "$PWD/result.json" \
  > pl.out 2>&1 < /dev/null &

sleep 5
ls ready          # present  — the server came up
cat browser.log   # No such file — fake-browser was never executed
cat pl.out        # 0.27.8:  "Resolved: .../target.md" and nothing else
                  # 0.27.12: "Resolved: ..." plus "Plannotator session ready — open in
                  #          your browser: http://localhost:4497" — still no mention that
                  #          PLANNOTATOR_BROWSER was set and ignored
```

**Expected:** `fake-browser` is invoked with the review URL, or plannotator states why it was not.

**Observed:** `ready` is written, `browser.log` never exists, and `pl.out` contains one line naming
the resolved target. No warning, no error, no mention of the browser. Exit status is unaffected.

## What we have not established

The `setsid` correlation is solid and reproducible. **We have not proven that the missing controlling
terminal is the mechanism** — we could not run the comparison, because the environment where this was
found has no TTY available to test the positive case against. It may be TTY detection, it may be an
`isatty` guard on an unrelated code path, it may be something else entirely that co-varies with
detachment. Treat the reproduction as the fact and the cause as a hypothesis.

## Why it matters more than usual for this tool

Plannotator's value is that an agent can put a document in front of a human and get structured
feedback back. That makes agent invocation a primary mode, not an edge case — and agent processes are
detached by construction. On a headless host such as WSL there is no browser for the human to fall
back to, so "here is a URL" is not a recovery path.

## Suggested fix, in order of preference

1. Invoke `PLANNOTATOR_BROWSER` whenever it is set, regardless of terminal state. If it is set
   explicitly, the caller has already decided.
2. Failing that, log one line when the browser is skipped, naming the reason. That alone would have
   turned our debugging session into a five-second read.
3. A `--open` / `--no-open` flag would make the behaviour explicit and testable, and would let a
   caller stop guessing.

### Plannotator version

plannotator 0.27.12

### OS

Linux

### Agent

Gemini CLI

### Where did it happen?

VS Code extension

FNDEVVE:
## Problem

With `PLANNOTATOR_BROWSER` explicitly set under WSL, the value was routed through `cmd.exe /c start` regardless of what it named, so a Linux path or script could never launch; the failure was swallowed by a bare `catch` and the ready output only printed the URL. A headless (agent-launched, no TTY) session then had every observable signal reporting success while the reviewer was never shown anything. The Pi runtime had the same branch and additionally reported `opened: true` on spawn errors. #1472

## Change

- Both runtimes classify the configured value first: Windows targets (`*.exe`, `C:\…`, `/mnt/<drive>/…`) keep the `cmd.exe` route; POSIX paths (`/…`, `./…`, `../…`) and bare names that resolve to an executable on the Linux PATH are executed directly — also under WSL.
- The Bun runtime spawns the **resolved absolute path**: Bun 1.3's `$` resolves bare command names against the PATH captured at process startup, so a bare name would otherwise be invisible to the shell. The end-to-end test pins this.
- When an explicitly configured browser fails, one stderr line names the executable and the error before the existing fallback (VS Code IPC / URL print) runs. The Pi runtime returns `opened: false` on spawn errors instead of claiming success.
- Non-WSL Linux and macOS paths are unchanged; the macOS branch recently fixed by #1429 is untouched.

## Verification

Bun 1.3.14, disposable HOME: `packages/server/browser.test.ts` + `apps/pi-extension/server/network.test.ts` → **49 pass / 0 fail**:

- direct spawn of a POSIX path and of a bare PATH name under mocked WSL, asserting the target script received the URL as its argument;
- `chrome.exe` still routed through a `cmd.exe` shim (end-to-end in the Pi runtime, whose `node:child_process` spawn honors a mutated PATH on every Bun version; the Bun-side resolver tests pin the same routing decision, since a `cmd.exe` shim cannot be spawned through Bun 1.3's `$` on a non-Windows host);
- a failing executable warns on stderr naming itself, and a missing one yields `opened: false` with the fallback intact.

On Bun 1.4.2 the only failures in these two files are the three pre-existing port-selection listener-count tests (baseline-verified before this change).

Fixes #1472

### DIFF
--- a/apps/pi-extension/server/network.test.ts
+++ b/apps/pi-extension/server/network.test.ts
@@ -1,5 +1,15 @@
 import { afterEach, describe, expect, test } from "bun:test";
+import {
+	chmodSync,
+	existsSync,
+	mkdtempSync,
+	readFileSync,
+	rmSync,
+	writeFileSync,
+} from "node:fs";
 import { createServer } from "node:http";
+import os, { tmpdir } from "node:os";
+import { delimiter, join } from "node:path";
 import { closeServer, occupyConsecutivePorts } from "../../../tests/helpers/ports.ts";
 import {
 	buildAdvertisedUrl,
@@ -7,6 +17,7 @@ import {
 	getServerPort,
 	getServerPorts,
 	isNoOpBrowserSentinel,
+	isPosixBrowserTarget,
 	isRemoteSession,
 	listenOnPort,
 	openBrowser,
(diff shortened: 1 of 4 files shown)


backnotprop:
Reviewed in depth before merge: mapped the full browser-launch matrix on both runtimes and confirmed exactly one cell changes execution (WSL + POSIX PLANNOTATOR_BROWSER - the #1472 silent no-op, now actually invoked with the resolved absolute path), the new failure warning can never break a session (warn then IPC fallback then the URL lifeline; no exit paths), WSL detection is untouched so plain-Linux false-positive risk is unchanged, and revert-verification at three granularities confirms the tests genuinely guard both halves. The Pi-side truthful `opened` flag is a nice bonus - headless Pi users now get the open-this-URL notification instead of silent false success. Two cosmetic notes for a rainy day: the /mnt/ prefix classifier is broader than its comment (a /^\/mnt\/[A-Za-z](\/|$)/ tighten would match it), and a non-daemonizing browser script blocks openBrowser until exit (pre-existing pattern). Merging.

PR #1495, opened 2026-09-10
fix(pi): render completed messages and errors

a4180p:
## Summary
- map text blocks from Pi RPC `message_end` events for completed assistant messages
- surface completed Pi request errors in Ask AI instead of reporting a successful empty response
- cover completed assistant, failed assistant, and non-assistant messages

Pi documents `message_end.message` as the authoritative completed assistant response. The completed-message fallback lets Ask AI render a response when a Pi provider does not emit usable `text_delta` events. In particular, a failed Pi request (such as an upstream 429) previously ended as a successful empty Plannotator response; the provider now sends its `errorMessage` to the UI. The UI retains its existing streamed-text preference, so normal delta streaming is not duplicated.

## Validation
- `bun run typecheck`
- `bun test packages/ai/ai.test.ts`

### DIFF
--- a/packages/ai/ai.test.ts
+++ b/packages/ai/ai.test.ts
@@ -1591,6 +1591,23 @@ describe("mapPiEvent", () => {
     expect(result).toEqual([]);
   });
 
+  test("completed assistant error maps to an error", () => {
+    const result = mapPiEvent({
+      type: "message_end",
+      message: {
+        role: "assistant",
+        content: [],
+        stopReason: "error",
+        errorMessage: "OpenAI API error (429): quota exceeded",
+      },
+    }, SESSION_ID);
+    expect(result).toEqual([{
+      type: "error",
+      error: "OpenAI API error (429): quota exceeded",
+      code: "pi_request_error",
+    }]);
+  });
+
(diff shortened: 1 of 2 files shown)


backnotprop:
Reviewed in depth before merge, including against the installed Pi package's actual RPC internals: message_end is authoritative and always follows deltas, but the existing first-text dedup guard in useAIChat (the same architecture the Claude provider uses) means normally streamed responses render byte-identically - no double-render. The no-delta path is real (agent-loop's !addedPartial branch), and the error branch targets the only place current Pi actually reports errors: the old message_update error mapping is dead code on current Pi, which is exactly why quota/auth errors rendered as blank successful answers. Verified the error path can't wedge the SSE drain (message_end → turn_end → agent_end still terminates normally) and the error field survives the subsequent result. Revert-verification: two of your three tests genuinely fail against main.

One follow-up we'll track on our side, not yours: in a no-delta session with tool use, the first-text guard keeps only the first assistant message's text - strictly better than today's empty response, and the same shape the Claude provider lives with, but a text-replaces-on-message-boundary refinement in useAIChat would improve both providers. Merging - excellent first contribution.

PR #1496, opened 2026-09-10
fix(pi): persist approved plan checklist progress

### LINKED ISSUE #1376: Execution progress ([DONE:n]) resets to 0/N after compaction or session fork — plan file checkboxes never written back
## Summary

During plan execution, `[DONE:n]` progress is tracked **only in memory**. The plan file's own checkboxes are never updated, and progress is rebuilt from sources that compaction destroys — so after a `/compact`, session fork, or resume, the per-turn injection reports `0/N steps complete` and re-lists every step as remaining, even though they were already executed.

## Root cause

In `apps/pi-extension/index.ts`:

1. `turn_end` handler calls `markCompletedSteps(text, checklistItems)`, which only mutates the in-memory array. Nothing ever writes the completion back to the plan file (`- [ ]` → `- [x]`).
2. `before_agent_start` **re-parses the plan file from disk every turn** (`checklistItems = parseChecklist(planContent)`), discarding the in-memory marks before the status snapshot is built — so the injected "Todo status" is stale even *without* compaction.
3. `resyncPhaseFromSession()` (session_start / session_tree) rebuilds progress by re-reading the plan file (all still `- [ ]`) plus scanning branch messages after the last `plannotator-execute` entry for `[DONE:n]`. Compaction removes those assistant messages from the active branch, so the rebuild finds nothing and progress resets to zero.

Net effect: any compaction (or fork onto a compacted branch) revives the full `[PLANNOTATOR - EXECUTING PLAN] Remaining steps: ...` injection mid-execution, and the agent is told to redo finished work.

## Repro

1. Enter plan mode, submit a plan containing a `- [ ]` checklist, approve it.
2. Execute some steps; the agent emits `[DONE:1]` etc. (progress widget shows partial completion).
3. Run `/compact` (or fork the session and then compact).
4. Next turn: injection claims `0/N steps complete` with all steps listed as remaining.

## Suggested fix

Persist completion into the plan file itself — the file is the only state that reliably survives compaction, forks, resumes, and restarts. Concretely, upgrade the n-th checkbox line (same ordinal mapping as `parseChecklist`) whenever marks are detected, and once more when resync rebuilds state from messages (self-healing for pre-fix sessions):

```ts
// persist-checklist.ts
import { readFileSync, writeFileSync } from "node:fs";

/** Upgrade `- [ ]` -> `- [x]` for the given 1-based checkbox ordinals.
 *  Upgrade-only, no-op when nothing changes; all failures silent. */
export function persistDoneStepsToPlanFile(
  fullPath: string,
  doneSteps: Iterable<number>,
): boolean {
  const done = doneSteps instanceof Set ? doneSteps : new Set(doneSteps);
  if (done.size === 0) return false;
  let content: string;
  try {
    content = readFileSync(fullPath, "utf-8");
  } catch {
    return false;
  }
  // Same line shape as parseChecklist(): /^[-*]\s*\[([ xX])\]\s+(.+)$/gm
  let ordinal = 0;
  let changed = false;
  const updated = content.replace(
    /^([-*]\s*)\[([ xX])\](\s+.+)$/gm,
    (line: string, bullet: string, mark: string, rest: string) => {
      ordinal += 1;
      if (mark === " " && done.has(ordinal)) {
        changed = true;
        return `${bullet}[x]${rest}`;
      }
      return line;
    },
  );
  if (!changed) return false;
  try {
    writeFileSync(fullPath, updated, "utf-8");
  } catch {
    return false;
  }
  return true;
}
```

Call sites in `index.ts`:

- `turn_end`, inside the existing `markCompletedSteps(...) > 0` branch:
  ```ts
  if (lastSubmittedPath) {
    persistDoneStepsToPlanFile(resolve(ctx.cwd, lastSubmittedPath), extractDoneSteps(text));
  }
  ```
- `resyncPhaseFromSession`, after the message-scan loop (writes back whatever was rebuilt):
  ```ts
  persistDoneStepsToPlanFile(
    fullPath,
    checklistItems.filter((i) => i.completed).map((i) => i.step),
  );
  ```

I've been running this locally as a patched copy of the extension: unit tests for the helper pass (13/13: ordinal mapping, idempotency, CRLF preservation, upgrade-only, missing-file silence), and an end-to-end run (approve → execute → `[DONE:1]`/`[DONE:2]` → full session reload) shows the file checkboxes auto-checked and the **Plan Complete!** state surviving reload with no regressed injection. Happy to open a PR with this change if you'd like.

## Environment

- `@plannotator/pi-extension` 0.27.4
- pi 0.84.2 (hosted via pi-web 0.8.9, multiple sessions per process)
- Windows 11, Chrome 151

a4180p:
## Summary
- persist `[DONE:n]` completion state into the approved plan Markdown file
- add `plannotator_mark_done`, which agents call after each completed step and before starting the next step
- restore older in-progress executions by writing recovered completion markers back to the plan
- add coverage for rendering, immediate per-step updates, execution persistence, and restored sessions

Fixes #1376.

## Verification
- `npx --yes bun@1.4.2 test apps/pi-extension/todo-provider-sync.test.ts apps/pi-extension/tool-scope.test.ts apps/pi-extension/phase-prompts.test.ts packages/shared/checklist.test.ts`


backnotprop:
Deep review done. The core design is right - persisting progress into the approved plan file's own checkboxes is exactly what #1376 proposed, the per-call path resolution is clean, the one-way todo-provider invariant holds, lifecycle keying is correct, and your new tests are behavioral and revert-verified. Two fixes needed before merge:

1. **`apps/pi-extension/phase-tools-runtime.test.ts` fails on your branch** (passes on main): "completing the plan releases the executing phase tools" pins the exact executing-phase tool set, and the new `plannotator_mark_done` union at index.ts:544-545 changes it. Your verification ran only the four files you added - please run the full `bun test apps/pi-extension` and update that test's expectations (its agent_end tail will then also usefully pin that mark_done is released).

2. **Contradictory first-turn instruction**: the plan-approved `doneMsg` (index.ts:1394-1397) still tells the model to include `[DONE:n]` per step, while the executing framing delivered on the same turn says to call `plannotator_mark_done` with markers as fallback. One line to align `doneMsg` with the new primary mechanism.

Non-blocking but worth taking while you're in there: `renderCompletedChecklist` can flip the wrong checkbox on a plan containing a whitespace-only checkbox line (the shared pattern's \s+ crosses newlines; first-occurrence `replace("[ ]", ...)` then hits the blank placeholder) - replacing via the match's own capture position and mirroring parseChecklist's empty-text skip hardens it. And a short comment noting the one-turn ordinal-desync window when an agent edits the plan mid-turn would save a future reader the analysis.

One user-visible consequence we'll carry to release notes rather than ask you to change: the plan file's git status now dirties as steps complete - inherent to the fix as designed. Merge follows as soon as items 1-2 land.

backnotprop:
Pushed the review fixes as 8aa7c37b (maintainer edit on the fork branch):

1. **phase-tools-runtime** now pins `plannotator_mark_done` in the executing-phase active tool union / `phaseAddedTools`, and its agent_end tail explicitly pins that mark_done is released. Full `bun test apps/pi-extension` passes (273/273).
2. **`doneMsg` aligned** with the executing framing: primary is `plannotator_mark_done` after each completed step, `[DONE:n]` markers as the fallback — one consistent instruction on the approval turn.
3. **`renderCompletedChecklist` hardened.** The marker is now replaced at the match's own capture position with parseChecklist's empty-text skip mirrored — and since the wrong-box flip is rooted in the shared pattern's `\s+` crossing newlines (a blank `- [ ]` line swallows the next real line as its "text", in parse and render alike), the pattern's whitespace is restricted to `[^\S\n]` so a match can never span lines. That also stops parseChecklist from producing garbage items like `"- [ ] Step two"` in the widget/todo mirror. Regression test added (blank checkbox placeholders around real steps; the real step's box flips, placeholders untouched) and revert-verified against the old implementation.
4. Short comments added on the one-turn ordinal-desync window when an agent edits the plan's checkboxes mid-turn (bounded by upgrade-only writes + per-turn re-parse).

Prompt-cache invariants re-checked on the result: no systemPrompt writes, active tool set changes only at phase entry/exit via `applyPhaseConfig`/release (mark_done execution never touches `setActiveTools`), and all per-turn deliveries remain conversation-suffix appends.

Verified: `bun test apps/pi-extension` 273 pass, `bun test packages/shared` 1057 pass, `bun run typecheck` clean.

PR #1497, opened 2026-09-10
fix(review): stabilize JJ line-of-work bases

graemefolk:
## Summary

- resolve automatic JJ line-of-work bases to immutable full commit IDs
- report ambiguous or unavailable fork points instead of choosing by timestamp or silently using trunk
- fall back honestly to Current change while preserving the requested Line of work preference
- expose detected fork points for explicit selection and keep Bun/Pi behavior aligned

## Why

A moving remote bookmark or ambiguous fork-point revset could make a JJ review include unrelated changes or label a Current-change diff as Line of work. Freezing the automatic base and representing resolution failure explicitly keeps every diff, snapshot, fingerprint, and file lookup on the same revision.

## Validation

- `bun test packages/shared/jj-core.test.ts packages/shared/vcs-core.test.ts packages/server/jj.test.ts packages/server/vcs.test.ts`
- `bun test apps/pi-extension/server.test.ts`

### DIFF
--- a/apps/pi-extension/server.test.ts
+++ b/apps/pi-extension/server.test.ts
@@ -1786,11 +1786,30 @@ describe("pi review server", () => {
       expect(last.diffType).toBe("jj-last");
       expect(last.rawPatch).toContain("last.txt");
 
+      const automaticLineResponse = await fetch(`${server.url}/api/diff/switch`, {
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({ diffType: "jj-line" }),
+      });
+      expect(automaticLineResponse.status).toBe(200);
+      const automaticLine = await automaticLineResponse.json() as {
+        diffType: string;
+        gitContext?: { diffFallback?: { requestedDiffType: string; effectiveDiffType: string } };
+      };
+      expect(automaticLine.diffType).toBe("jj-current");
+      expect(automaticLine.gitContext?.diffFallback).toEqual({
+        requestedDiffType: "jj-line",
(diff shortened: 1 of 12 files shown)


backnotprop:
Reviewed in depth, including empirically: we rebuilt your criss-cross and moving-bookmark scenarios against real jj 0.41 repos and confirmed both bugs and both halves of the fix - `exactly()` surfacing the ambiguity the old `latest()` timestamp tie-break silently swallowed, and the frozen commit id keeping the diff stable when a bookmark moves. The fallback-to-jj-current banner flow verified end-to-end on both server runtimes, and revert-verification confirmed the regression tests genuinely guard the fix. Quality work.

Since maintainer edits are enabled, we pushed one commit (6eccc2d5) with small review follow-ups rather than round-tripping comments:
- Scoped the switch handler's session-context adoption to gitbutler + jj. Your version adopted the recomputed context for every provider, which changed reload semantics for plain-git sessions (a worktree switch would repoint the WorktreePicker's launch-cwd row and the feedback branch label on the next reload). The recomputed context still rides the switch response transiently, as before.
- Truncated frozen 40-hex ids in the jj-line header label to the 12-char short form.
- Comment-marked the two deliberate test pins per the repo's testing rules.

Two observations left as-is, flagged for transparency: `jjLineBase` on GitContext currently has no reader (fine as future picker surface), and a jj too old for `fork_point` now gets an honest jj-current fallback where it previously got jj-line vs trunk() - we think honest is right. Merging once CI is green on the updated head.

PR #1498, opened 2026-09-10
feat(review): add provider-aware review defaults

graemefolk:
## Summary

- replace the Git-named settings tab with a provider-aware Review tab
- let registered VCS providers describe their own default review modes
- persist defaults under provider namespaces while retaining legacy Git behavior
- add JJ defaults without inventing Git-only staging, status, or remote-branch semantics
- keep explicit `--base` and `--diff-type` session overrides authoritative across Bun, OpenCode, and Pi

## Why

Review-wide settings and Git-only behavior were coupled in one global configuration. Provider-owned descriptors keep the UI small and generic while allowing Git and JJ to expose only the modes they actually support.

## Validation

- `bun test packages/shared/config.test.ts packages/shared/vcs-core.test.ts packages/shared/review-open-state.test.ts packages/shared/review-args.test.ts packages/server/vcs.test.ts packages/ui/config/reviewPanelViewLastUsed.test.ts`
- `bun test apps/pi-extension/review-args-parity.test.ts apps/pi-extension/server.test.ts`
- `bun run build:hook`

### DIFF
--- a/apps/hook/server/index.ts
+++ b/apps/hook/server/index.ts
@@ -1096,13 +1096,13 @@ if (args[0] === "sessions") {
         isPRMode: false,
         isWorkspace: false,
         providerId,
-        resolvedDefaultDiffType: resolveDefaultDiffType(config),
+        resolvedDefaultDiffType: resolveDefaultDiffType(config, providerId),
       });
       const diffResult = await prepareLocalReviewDiff({
         vcsType: reviewArgs.vcsType,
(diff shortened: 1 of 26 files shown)


graemefolk:
Closing in favor of an upcoming stacked pair. The VCS abstraction boundary could use some refactoring to make this cleaner, and then the JJ review-defaults feature on top.

PR #1513, opened 2026-09-11
fix(review): prevent submission footer overlap

leoreisdias:
## Problem

The GitHub/GitLab review submission dialog kept its sticky action footer inside the scrollable form body. At constrained heights, the footer overlaid the final form control and visibly clipped the "View on GitHub after submitting" checkbox.

## Fix

Move the action footer outside the scrollable body and make it a non-shrinking sibling. The form content can now scroll independently while the actions remain available without covering any fields.

Add a regression test that verifies the action footer is not contained by the dialog's scrollable form region.

## Review Focus

The meaningful layout change is limited to the boundary between the scrollable form body and the persistent action footer. Submission behavior, button states, and platform actions are unchanged.

Before:
file-c63dbd71d55ca53c3f9a43ceb8171eb8

After:
image

## Test Plan

- [x] `DOM_TESTS=1 bun test packages/review-editor/components/ReviewSubmissionDialog.ui.test.tsx` (6 pass)
- [x] `bun run --cwd apps/review build`
- [x] Rebuilt the local hook bundle and verified the GitHub approval dialog against a live PR: the checkbox is fully visible above the footer and the two regions do not overlap
- [x] Impeccable layout detector reports no findings for the changed files
- [ ] Full repository typecheck is currently unavailable from the root script because `tsc` is not on the root PATH after the latest upstream dependency layout; a direct app typecheck also reports unrelated baseline errors across existing files

### DIFF
--- a/packages/review-editor/components/ReviewSubmissionDialog.tsx
+++ b/packages/review-editor/components/ReviewSubmissionDialog.tsx
@@ -551,8 +551,10 @@ export function ReviewSubmissionDialog({
           View on {platformLabel} after submitting
         </label>
 
+        </div>
+
         {/* Actions */}
-        <div className="sticky bottom-0 -mx-4 -mb-4 flex justify-end gap-2 border-t border-border/50 bg-card px-4 pb-4 pt-3 sm:-mx-6 sm:-mb-6 sm:px-6 sm:pb-6">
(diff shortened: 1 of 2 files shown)


backnotprop:
Merged, thanks @leoreisdias. Verified the footer now sits outside the dialog's scroll region at short viewports (0px overlap at 1280x600). Ships in the next release.

PR #1514, opened 2026-09-12
fix(ci): build x64 releases for baseline CPUs

ybtam:
## Summary
- Compile the Linux x64 CLI and paste service with `bun-linux-x64-baseline`.
- Add a bounded QEMU Nehalem gate to the existing Linux release smoke job for CLI `--help`, exact `--version`, and paste-service startup.
- Preserve asset names, checksums, provenance, installers, the Bun pin, and other platform targets.

## Motivation
The official v0.27.14 Linux x64 CLI exits with SIGILL on an Ivy Bridge Xeon E3-1220 V2 and QEMU Nehalem. Explicit baseline builds start successfully. [Bun 1.3.14's executable documentation](https://github.com/oven-sh/bun/blob/bun-v1.3.14/docs/bundler/executables.mdx) recommends baseline for pre-2013 CPUs.

## Validation
- Rebuilt review/hook UI and both Linux x64 executables with pinned Bun 1.3.14.
- Native Ivy Bridge CLI help and exact `plannotator 0.27.14` version passed.
- Exact workflow smoke passed under QEMU 10.2.1 Nehalem; incompatible CLI/paste binaries and a version mismatch were rejected. A non-exiting real server was stopped at 60 seconds (exit 124).
- Installer/version tests: 126 passed, 8 PowerShell-dependent skips. Release-evidence tests: 4 passed.
- Actionlint 1.7.12 with ShellCheck 0.11.0, shell syntax checks, workflow scope assertions, and `git diff --check` passed.

## Limitations
Remote CI is pending. Paste coverage verifies startup after `Bun.serve`, without HTTP or persistence checks. Emulation covers bounded startup, and performance was not benchmarked.


backnotprop:
Thanks, kept your commit as-is (rebased onto main) and added two follow-ups on the branch.

The gate now runs a **control** first: the build job compiles the same CLI with the *default* x64 target (AVX2) into its own `baseline-gate-control` artifact (never `binaries`, so it can't be attested or released), and the Linux smoke leg requires it to die with exit 132 under the identical `qemu -cpu Nehalem` call. A green gate now proves the harness can actually tell baseline from non-baseline.

The `qemu-user` apt install is retried 3x with a short sleep so a mirror blip can't fail a release. The paste-service poll was sampling 30 s (60 x 0.5 s) against a 60 s subprocess timeout and could time out spuriously; it now samples every 2 s for 90 s under a 120 s per-run ceiling, reads the log before liveness, and every failure message names the binary and the CPU model.

Windows x64 (CLI + paste) moves to `bun-windows-x64-baseline` too, same CPU argument and upstream Bun did the same in oven-sh/bun#34782, but it is gated only by the existing smoke leg since there's no cheap emulated-old-CPU run for Windows on GitHub runners. macOS x64 stays default (every supported Mac is Haswell+).

Asset names are unchanged, so installers, evidence subjects, attestation subjects and docs need nothing. Added CI cost is one cross-compile plus an artifact round trip (~1 min) and a few seconds of emulated startup. actionlint + shellcheck clean; the added shell was dry-run locally against fake binaries. The real SIGILL-under-qemu assertion runs for the first time in CI, which is exactly what the control is for.

Issue #1524, opened 2026-09-13
Feature request: unified review navigator with independent layout and grouping


tekumara:
## Motivation

The current review navigator spreads related functionality across 3 overlapping surfaces: Tree, Git status, and Commits. I'd like to bring them into one cohesive, ergonomic surface: the unified review navigator.

This would make navigation less confusing and remove the need to switch modes to access complementary information. File layout and change grouping should be independent choices, rather than competing navigation modes.

## Proposed behaviour

Replace the separate navigation modes with one review navigator containing 2 independent controls:

| Control | Options | Behaviour |
| --- | --- | --- |
| File layout | Flat / Tree | Display files as a flat list or an expandable directory hierarchy |
| Grouping | All / By Git status | Display all changes together or separate them by Git status |

Changing one control should not change the other. All 4 combinations should be supported.

### Grouping by Git status

Show collapsible sections in this order:

- Staged — changes in the index
- Unstaged — working-tree changes not staged
- Committed — individual commits in the selected review range

The Committed section is the key distinction: show each commit as an expandable row, rather than flattening committed work into one file list. Each row shows its commit subject and aggregate additions/deletions.

Expanding a commit reveals the files it changed. Selecting one of those files should show its diff for that commit, not the combined branch diff.

```text
Staged
  files

Unstaged
  files

Committed
  commit subject
    files changed by this commit
  commit subject
    files changed by this commit
```

All file collections, including those inside commits, should respect the selected Flat or Tree layout.

### Showing all changes together

Show one combined set of changed files for the selected review scope, without status sections or individual commit rows. The Flat/Tree control should still work.

## File presentation and layout

- in Flat layout, show the filename first and the parent directory as secondary text
- in Tree layout, use expandable directories with compact, consistent indentation
- show each file's change status and additions/deletions in stable trailing columns
- show aggregate diff totals on status groups and commits, with directory totals where useful
- suppress redundant totals on single-child directories
- omit explicit file and commit counts

Keep the existing base-reference selector above the 2 compact segmented controls. Keep this toolbar fixed while the changes below it scroll. Allow the control groups to stack when the panel becomes too narrow.

Long commit subjects should wrap without overlapping the statistics or following rows. Align the disclosure control and statistics with the first text line.

## Scope

This request combines the navigation capabilities of Tree, Git status, and Commits. It does not require additional top-level tabs, a commit-message input, or changes to staging and commit workflows.

## Prototype

[...]

PR #1526, opened 2026-09-13
feat(review): unified review navigator with independent layout and grouping

backnotprop:
Implements #1524.

The review's left panel is now one surface instead of three competing modes.

## The model

Under the existing worktree / diff-type / base-reference pickers sit two **independent** segmented controls:

| Control | Options | What it decides |
| --- | --- | --- |
| **File layout** | `Flat` / `Tree` | how a list of files is drawn |
| **Grouping** | `All` / `By Git status` | which lists exist |

All four combinations render, and changing one never changes the other. The toolbar (base picker, both controls, the utility row) stays fixed while the list scrolls, and the two control groups stack once the resizable panel gets narrow.

**Layout.** `Flat` leads with the filename and carries the directory as quiet secondary text that absorbs truncation first; `Tree` nests expandable directories with per-directory totals, single-child folder chains merged, and totals suppressed on a folder whose only child is a file. Folders are expanded unless explicitly collapsed, so a refreshed diff that introduces a directory needs no re-seeding pass. Every file collection goes through one renderer — the combined All set, Staged, Unstaged, and the files inside a commit — so the layout applies uniformly.

**Grouping = All** lists the active diff's files, no sections. **Grouping = By Git status** renders **Staged** / **Unstaged** / **Committed**, each header carrying its aggregate `+/-`. Untracked files fold into Unstaged and keep their `U` letter. **Committed** is one expandable row per commit in the review range over the same first-parent walk and `In <base>` boundary the retired Commits rail used — HEAD badge on the subject line, author + avatar + relative time + short sha on a quiet second line, `Show more` paging at the end, long subjects wrapping without running into the stats. Expanding a commit lists the files **that commit** changed.

**Selection carries a scope**, not just a path: `{ scope: {kind:'working'} | {kind:'commit',sha}, path }`. Picking a file inside a commit row switches the diff to `commit:<sha>` (vs its first parent) and focuses that file; picking any Staged / Unstaged / All row restores the session's working diff. There is no mode to enter or leave — the old `preCommitDiffRef` "exit the Commits view" detour is gone, replaced by a `workingSet` snapshot App takes on every applied non-commit diff (files + sections + diff type + base). That snapshot is also what keeps Staged/Unstaged describing the working tree while a commit's diff is on screen.

One path with both committed and uncommitted work legitimately appears twice — under Unstaged with its working-tree delta, and inside each commit with that commit's delta. The counts differ by design, so the highlight and the viewed mark key on the **pair**: `viewedFiles` stays the working scope's (draft-persisted, unchanged) and commit rows get their own session-scoped `commitViewedFiles`.

## Server change (both runtimes)

- `listCommitHistory` now sums a `--numstat` pass riding the same `git log`, so every commit row carries its own `additions` / `deletions` without an extra walk.
- New `GET /api/commit-files?sha=<bare hex>` returns `{ sha, files: [{ path, oldPath?, status, additions, deletions, binary }] }`, fetched lazily when a row is expanded and cached for the session. It uses the same ref pair `runGitDiff` uses for `commit:<sha>` (first parent, or the empty tree for a root commit), so the list can never disagree with the patch a click on one of those rows renders. Same session gate as `/api/commits`.

Shared logic lives in `packages/shared/commit-history.ts` (vendored to Pi); the route is implemented in `packages/server/review.ts` **and** `apps/pi-extension/server/serverReview.ts`.

The pure layout/grouping helper lives in `packages/ui/utils/reviewNavigator.ts`, not `@plannotator/core`: `@plannotator/ui` pins the published core exactly and may only import subpaths that version exports (`packages/ui/scripts/package-smoke.ts` enforces it), and one helper does not justify a core publish. The existing `"./utils/*"` export already puts it on ui's public surface.

## The pinned guides.show viewer build changed — deploy on merge

`packages/core/guide-viewer-manifest.ts` is repinned (`viewer.KTNT-M2b.js` → `viewer.BZIcKvcg.js`, `viewer.B0Ro0qg9.css` → `viewer.D0niou63.css`, plus their two SRI hashes). **The new build must be uploaded to guides.show `/v1/` when this merges**, or exports pinned by this version will 404. Every grammar chunk hash is unchanged.

Why it moved: the viewer's `main.tsx` imports `@plannotator/review-editor/styles` and the shared `ThemeProvider`, and `configStore` statically imports the whole `SETTINGS` registry so it can seed defaults on every surface — so replacing the `reviewPanelView` entries with the navigator pair changes the content-hashed viewer JS, and the review stylesheet's Tailwind sources cover `packages/review-editor`, so the new components' utility classes change the viewer CSS. Unavoidable without splitting the registry.

No app-only surface leaked in: the bundle contains no `ReviewNavigator` / `NavigatorCollection` / commit-row code and no `/api/commit-files` call — only the two new cookie-key strings in the registry. `check:budgets` passes (js 396.4 KB gz of 400, css 51.1 KB gz of 64) and `check:manifest` is in sync.

## Settings migration

`reviewPanelView` (`tree | sections`) and its cookie-only `reviewPanelViewLastUsed` memo encoded one three-way mode that set layout and grouping together. They are replaced by `reviewNavigatorLayout` (`flat | tree`) and `reviewNavigatorGrouping` (`all | status`), both cookie-only.

Both **migrate the retired cookies on read and never write them back** — last-used first, then the persisted default, mapped `tree` → Tree + All and `sections` → Flat + By Git status (`commits` was session-only and has no mapping). A migrating read returns a value, so the registry's default-seeding write never fires and resolution stays pure and identical every load until the reviewer touches a control — the same shape as the `tokenHoverCards` → `tokenHoverTrigger` migration. `hasPersistedNavigatorChoice()` counts a legacy cookie as a real choice, so first-run seeding never resets an upgrading reviewer.

`openStatePinned` semantics are unchanged: `--base` / `--diff-type` still skip the first-run dialog without consuming its cookie and still persist nothing (nothing the navigator writes is a diff default, so pinning is unaffected).

## What happened to the coupling

The old rule was "Sections only renders since-base, so a classic diff default snaps the view to Tree". With grouping orthogonal:

- **`setReviewDefaultDiffType` no longer touches the panel.** No conflicted pair can exist, so `shouldRepairPanelPair` and the App-level self-heal are removed.
- **Clicking `By Git status` still asks for the composite diff**, session-only: if a classic diff is active it switches the live diff to `since-base` (exactly what the retired Git status segment did) and persists no diff default. The layout is untouched.
- **A diff with no git-status partition degrades to `All` in the render**, with the reason stated above the list, and the *selection* survives — returning to a partitionable diff restores the sections without re-picking.
- **Sessions with no git index** (PR reviews, multi-repo workspaces, jj / GitButler / Perforce) disable the `By Git status` segment with a per-mode reason in its tooltip rather than showing three empty section headers. Committed is offered only where `/api/commits` is (plain local git).

## Removed

`PanelViewToggle`, `SectionsPanel`, `CommitsPanel`, `FileTree`, `FileTreeNode`, `resolvePanelView`, `useCommitsView` (plus its HEAD auto-select and centre veil, which belonged to the retired mode), `shouldRepairPanelPair` and the self-heal effect, the `reviewPanelView` / `reviewPanelViewLastUsed` settings, and the explicit counts the issue asked to drop (`N commits`, `N more files`, the viewport-adaptive Committed truncation). `SearchFileGroup` moved out of `FileTree` into `SearchResults.tsx` unchanged. The viewed `N/M` counter survives in the fixed toolbar, outside the list.

## Known gaps

- The `PanelControlsRow` utility row (viewed counter, staged count, icons) wraps awkwardly at the 200px minimum panel width. Pre-existing and shared; not touched here.
- The Committed header totals only the commits loaded so far, so a branch with more than one page of branch-local commits under-reports until `Show more` (chosen over over-reporting the base's whole history).
- The flat layout's directory text can truncate to nothing on a very narrow panel; the filename always survives and the row `title` carries the full path.
- `/api/commit-files` results are cached per session and never invalidated — correct, since a commit's diff is immutable, but a force-push that reuses a sha within one session would serve the old list until reload.

## Test plan

- `bun run typecheck` (includes `apps/pi-extension/vendor.sh`) — clean.
- `bun test packages/shared packages/core packages/server tests apps/pi-extension` — 2404 pass, 1 skip, 0 fail.
- `DOM_TESTS=1 bun test packages/review-editor packages/ui packages/editor packages/guide-viewer` — 2189 pass, 5 fail; all five are the pre-existing order-dependent `DocBadges` / `artifactDocument` failures that pass in isolation on `main`.
- `bun run --cwd apps/review build && bun run build:hook && bun run build:opencode && bun run build:marketing` — all succeed.
- New tests: the settings migration (`packages/core/review-navigator.test.ts`, `packages/ui/config/reviewNavigatorSetting.test.ts`), the shared commit parsers and `listCommitFiles` over real temp repos (`packages/shared/commit-files.test.ts`), the endpoint in **both** runtimes (`packages/server/commit-files-endpoint.test.ts`), the pure navigator model (`packages/review-editor/utils/navigatorModel.test.ts`), and the four layout×grouping combinations, commit-scoped selection, the Committed total, and the non-git gating in the DOM (`packages/review-editor/components/ReviewNavigator.test.tsx`). Updated: `reviewSetup.test.ts`. Removed: `resolvePanelView.test.ts`, `reviewPanelViewLastUsed.test.ts`, `FileTreeNode.controls.test.tsx`, `FileTree.search-placement.test.tsx`.
- Manual: captured against a throwaway repo (2 commits ahead of a base, 2 staged, 3 unstaged, 1 untracked) in all four combinations, with an expanded commit, and at both a wide and the 200px-minimum panel width. A manual checklist for the same flows is in `tests/UI-TESTING.md`.

### Captures

Seven screenshots of that session were taken (full window + panel crop each) and are
available on request; GitHub's API cannot upload images, so they are not inline here:

1. **Flat + All** - one row per file, filename first with its directory after it, no sections.
2. **Tree + All** - the same files nested under `fakesnow` / `transforms` / `tests` with directory totals.
3. **Tree + By Git status** - `Staged +8 -2`, `Unstaged +28 -1`, `Committed +21 -2`, each section's files as a tree, the two commits as rows and the `In base-main` divider below them.
4. **Flat + By Git status** - the same sections with flat rows; the layout choice survived the grouping switch and vice versa.
5. **Expanded commit** - `refactor: simplify file format tests` opened into `fakesnow/transforms/file_format.py +4` and `tests/test_file_format.py +5 -2`, which are that commit's numbers, not the branch's.
6. **Wide panel** - the two control groups sharing one row.
7. **200px panel** - the control groups stacked, filenames ellipsized inside their column, and the second commit's subject wrapping with `+12` still on its first line.


backnotprop:
Closing after evaluating it in the browser next to the current panel. The proposal ends up as more control surface for the same information, and the one genuinely new piece (commits expandable inside the Git status view with per-commit file diffs) is small enough to land on its own if there's demand. Keeping the current Tree / Git status / Commits panel.

PR #1548, opened 2026-09-16
fix(ui): parse nested frontmatter maps and arrays of maps

### LINKED ISSUE #1485: Annotate: frontmatter card shows wrong values when the YAML has nested keys
### What happened?

I run `plannotator annotate` on [OKF](https://github.com/GoogleCloudPlatform/open-knowledge-format)-based documents before committing them. Their frontmatter holds nested provenance metadata: who generated the document, who verified it, and which sources back it. The frontmatter card at the top of the annotate view flattens that structure, so child keys from different parents end up next to each other under the wrong names.

Example document:

```yaml
---
type: Fact
title: Markdown code fences
tags:
  - markdown
  - commonmark
status: stable
generated:
  by: claude-code/claude-fable-5-1
  at: 2026-09-02T12:05:02+09:00
verified:
  - by: human:jaeyeong.yang
    at: 2026-09-02T12:11:00+09:00
sources:
  - id: commonmark-fences
    resource: https://spec.commonmark.org/0.31.2/#fenced-code-blocks
  - id: rumdl
    resource: https://github.com/rvben/rumdl
---
```

What the card shows:

Image

I expected nested mappings to stay together: `generated` as a group with its own `by` and `at`, and each `sources` or `verified` item as one entry with its fields. If nested rendering is out of scope, showing only the top-level keys and leaving the nested part as raw text would still be better than merging child keys upward.

The card is what I read to decide whether a document is ready: is `verified` filled in, by whom, when, and do the sources cover the claims. Right now it answers those questions wrongly while looking complete, so I end up scrolling to the raw YAML anyway. Any frontmatter with a sub-map such as `author:` or `params:` gets the same treatment.

Related: #43 (frontmatter card), #1101 (block scalars). Neither mentions nested keys.

### Plannotator version

plannotator 0.27.12

### OS

macOS

### Agent

Claude Code

### Where did it happen?

Annotate

FNDEVVE:
`extractFrontmatter` trimmed every line, so indented child keys (`generated:`/`by:`/`at:`, `verified:`/`sources:` arrays of maps) hoisted to the document root, where same-named keys clobbered each other and the card rendered wrong-but-complete-looking values.

The parser is now indentation-aware: nested maps stay grouped, arrays of maps group per item, and flat scalars, string arrays, block scalars (`|`/`>`), and CRLF all behave as before. `FrontmatterCard` renders nested structures recursively (flat frontmatter renders exactly as today).

Testing: `bun test packages/ui/utils/parser.test.ts` (197 pass, incl. 6 new nested-structure tests + regressions), full `bun run typecheck` clean.

Closes #1485.


backnotprop:
Merged, thanks @FNDEVVE. Two small parser additions went in on top of your branch before merge: quoted keys (`"title": "Doc"`) parse as a mapping instead of vanishing from the card, and a nested map as an array item's first key no longer swallows its siblings. Fixes #1485. Ships in the next release.

PR #1549, opened 2026-09-16
feat(core): carry elementContext through html-anchor host helpers

### LINKED ISSUE #1521: Carry elementContext through the core html-anchor host helpers (follow-up to #1517)
#1517 added `elementContext` (and per-target `context`) to raw-HTML / live-app pinpoint annotations in `@plannotator/ui`, captured by the bridge and validated at the parent trust boundary. It deliberately left `@plannotator/core/html-anchor` untouched, so a host that persists through `buildPersistedHtmlAnchor` and reads through `projectHostThreads` drops the field on save and never gets it back on projection. Workspaces is such a host and has asked for this; their read-only review of #1517 is the source of the requirements below.

## Requirements (from the Workspaces host review)

1. `parseHtmlElementContext` lives in `@plannotator/core/html-anchor` next to `parseHtmlElementAnchor`, and `@plannotator/ui` re-uses it rather than keeping a hand-mirrored validator (the rule since core 0.25.0: no mirror validators).
2. `PersistedHtmlAnchor` carries `elementContext?` and each `HtmlAnnotationTarget` carries `context?`. **Key order:** append `elementContext` AFTER `htmlAdditionalTargets` so rows without it serialize byte-identically; the host's codec test pins the wire fingerprint.
3. `buildPersistedHtmlAnchor` sheds contexts BEFORE targets under `maxBytes`. The host's server cap stays 16 KiB, and one primary at 2 KiB plus 16 extras at 1 KiB already exceeds it; raising `DEFAULT_HTML_ANCHOR_MAX_BYTES` is not an engineering default (the 48 KiB figure in the design write-up was not shipped and core stays at 16 KiB).
4. `projectHostThreads` returns `elementContext` / per-target `context` on the projection so the host's panel (and the panel's per-row Copy, which calls `exportAnnotationEntry`) sees it. Paint never reads it; the repaint path already posts only anchors to the bridge.
5. An export entry point that takes the projected/host shape and can print the identity lines WITHOUT the fenced outline (an option on `exportAnnotationEntry` / `elementContextExportBlock`), because in a model turn the 600-char outline is the expensive part per annotation.

## Also noted by the host

- The panel's Copy is now gated on `elementContext` only (#1520), so anchor-only host rows do not change on a package bump.
- The context never exposes more than the document bytes the viewer can already open; href scrubbing matters to hosts where a share link is the credential.

Full write-up on the host side: `/Users/ramos/workspaces/research/integration/PLANNOTATOR-HTML-ELEMENT-CONTEXT-COMPARISON-2026-09-12.md` (local, not in this repo).

FNDEVVE:
Implements the five follow-up requirements from the Workspaces host review so hosts persisting through the core helpers no longer lose element context:

1. `parseHtmlElementContext` moves into `@plannotator/core/html-anchor` next to `parseHtmlElementAnchor`; `useHtmlAnnotation` imports and re-exports it and the hand-mirrored validator is deleted.
2. `PersistedHtmlAnchor` gains `elementContext?` (serialized after `htmlAdditionalTargets`, so rows without it are byte-identical) and `HtmlAnnotationTarget` gains `context?`.
3. `buildPersistedHtmlAnchor` sheds per-target contexts, then the primary context, before dropping targets under `maxBytes`; the 16 KiB cap is unchanged.
4. `projectHostThreads` projects both context fields through to host panels and per-row copy.
5. New `includeOutline` export option (default true, output unchanged) prints the identity lines without the fenced outline for model turns.

Testing: `bun test packages/core/html-anchor.test.ts packages/ui/utils/parser.test.ts` (214 pass, incl. wire-fingerprint, shed-order, projection, and outline-less export tests), full `bun run typecheck` clean.

Closes #1521.


backnotprop:
Thanks @FNDEVVE, this closes #1521 cleanly. The validator move is byte-faithful (no cap, allowlist, or fail-closed path changed), and it is what Workspaces needs to stop carrying a copy.

I pushed three small commits onto your branch rather than round-tripping: a single definition of `MAX_PAGE_URL_LENGTH` (the ui copy now imports core's), and the HANDOFF / README / module-doc updates so hosts read the new contract instead of the old "gap is open" note. Nothing in your code changed.

Publish order for this one is core 0.25.3 first, then ui 0.40.0 with the pin bumped, since ui now imports a symbol that published core 0.25.2 lacks. That lands in the 0.40.0 bundle PR going up shortly. Merging now.

PR #1554, opened 2026-09-16
feat(review): add support for patch/diff files

soundvibe:
# Review static unified diffs without a repository (--patch-file)                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                  
 Adds plannotator review --patch-file <path | -> — a way to open the code review UI against a caller-supplied unified diff, with no Git repo, no worktree, and no VCS detection. Useful for reviewing patches from emails, pastes, remote agents, CI artifacts, or anywhere a diff exists without  a checkout.                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                  
 ```bash                                                                                                                                                                                                                                                                                          
   plannotator review --patch-file reading.diff                                                                                                                                                                                                                                                   
   curl -s https://example.com/change.diff | plannotator review --patch-file -                                                                                                                                                                                                                    
 ```                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                  
 ### What it does                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                  
 - Serves the patch verbatim as the review content (diffType: "static-patch", snapshot-bound), with the patch path (or stdin patch) as the header label                                                                                                                                           
 - No local VCS affordances: no git context, no refresh (there is no live tree to recompute against), no diff-type switching —the static patch is the session                                                                                                                                    
 - Ask AI context carries the patch content, so questions about the changes work as usual                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                  
 ### Supported surfaces                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                  
 - Direct CLI (plannotator review, incl. --patch-file - for stdin)                                                                                                                                                                                                                                
 - OpenCode bridge (opencode-review; stdin variant refused with a clear error since the bridge's stdin carries the input JSON)                                                                                                                                                                    
 - Pi extension (/plannotator-review --patch-file, plus patchFile on the host-facing code-review event payload)                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                  
 ### Validation (in parseReviewArgs, shared by all runtimes)                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                  
 --patch-file wins over VCS detection entirely, so conflicting selectors are usage errors, not silently ignored flags:                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                  
 - with a PR/MR URL, --base, --diff-type, --git/--gitbutler, or an explicit --local → startup error naming the conflict                                                                                                                                                                           
 - missing or duplicate flag value → startup error                                                                                                                                                                                                                                                
 - empty/whitespace-only patch → startup error (both runtimes)                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                  
 ### Tests                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                  
 - Parser: parse / missing / duplicate / each conflict combination                                                                                                                                                                                                                                
 - CLI help surface: --patch-file in top-level and review usage                                                                                                                                                                                                                                   
 - Ask AI: patch used as context in static-patch mode                                                                                                                                                                                                                                             
 - Verified end-to-end manually: server serves the patch with a :static-patch snapshot id and correct label


soundvibe:
@backnotprop could you TAL?

backnotprop:
Thanks @soundvibe, and welcome. This is a good first PR: the design is right, the server degrades cleanly without a repo, nothing writes the patch to disk, and semantic diff works on the patch alone.

Rather than round-trip a review, I pushed the remaining pieces onto your branch:

- The browser now knows it is in patch mode. The server sends `sourceKind: "patch"` on every diff payload (absent reads as the VCS default, so existing sessions are unchanged) and serves `diffType`, which it had withheld; that was why the client fell back to its uncommitted default and showed Git Add buttons and "No uncommitted changes". Staging, base and diff-type pickers, the Git status and Commits panels, the fetch-base banner, hunk-context expansion, Edit Mode, the freshness probe, and the open-in / code-nav affordances are all off in patch mode, the header shows the patch filename, and a bad or empty patch says so.
- Open-in resolved patch paths against whatever repo the server was started from and could open an unrelated local file with the same name. Those endpoints now answer 400 in patch mode on both runtimes, as do git-add and file-content.
- `--no-local` now conflicts with `--patch-file` like `--local` does; Pi rejects `--patch-file -` with a clear message instead of reading a file named `-`; the dead `prUrl` guard is gone; a dual-runtime server test covers the advert and the 400s; the code-review doc page and the skill reference mention the flag.

Deliberately not touched: OpenCode's embedded path, which still ignores the flag. That only affects the new flag, not normal usage, so it can be a follow-up.

soundvibe:
> Thanks @soundvibe, and welcome. This is a good first PR: the design is right, the server degrades cleanly without a repo, nothing writes the patch to disk, and semantic diff works on the patch alone.
> 
> Rather than round-trip a review, I pushed the remaining pieces onto your branch:
> 
> * The browser now knows it is in patch mode. The server sends `sourceKind: "patch"` on every diff payload (absent reads as the VCS default, so existing sessions are unchanged) and serves `diffType`, which it had withheld; that was why the client fell back to its uncommitted default and showed Git Add buttons and "No uncommitted changes". Staging, base and diff-type pickers, the Git status and Commits panels, the fetch-base banner, hunk-context expansion, Edit Mode, the freshness probe, and the open-in / code-nav affordances are all off in patch mode, the header shows the patch filename, and a bad or empty patch says so.
> * Open-in resolved patch paths against whatever repo the server was started from and could open an unrelated local file with the same name. Those endpoints now answer 400 in patch mode on both runtimes, as do git-add and file-content.
> * `--no-local` now conflicts with `--patch-file` like `--local` does; Pi rejects `--patch-file -` with a clear message instead of reading a file named `-`; the dead `prUrl` guard is gone; a dual-runtime server test covers the advert and the 400s; the code-review doc page and the skill reference mention the flag.
> 
> Deliberately not touched: OpenCode's embedded path, which still ignores the flag. That only affects the new flag, not normal usage, so it can be a follow-up.

Thanks for such fast response and merge!

PR #1569, opened 2026-09-18
Switch OpenCode V2 model with approved agent

rcdailey:
OpenCode stores the session agent and model independently. Approving a plan currently selects Build while leaving Plan's model active when the agents have different model configurations.

This preserves model metadata from V2 agent discovery and applies `switchAgent` followed by `switchModel`, matching OpenCode's own clients. Hosts that cannot apply both selections avoid a partial switch, and failures continue to leave plan approval intact.


backnotprop:
Merged, thanks. Verified against the V2 wire shape (AgentV2Info.model is the ModelRef, so the id/variant read is right) and the CLI bridge path is untouched. One optional follow-up if you feel like it: when the host lacks switchModel but the agent has a configured model, the PR skips the agent switch entirely, while a throwing switchModel keeps the agent and warns. Treating both the same (switch the agent, warn about the model) would remove that inconsistency and keep older hosts switching agents as they did before. Will go out in 0.27.17.

rcdailey:
Thanks for the quick merge; I wanted to provide the screenshot below to show the fix working in action during local testing.

image

PR #1570, opened 2026-09-19
Preserve OpenCode agent switching on older hosts

rcdailey:
The coordinated model switch added in #1569 made `switchModel` support a prerequisite for switching agents. Older V2 hosts expose `switchAgent` without `switchModel`, so plan approval no longer changed the active agent there.

This keeps OpenCode's sequential behavior: switch the agent first, then apply its configured model when the host supports it. Missing or failing model switching warns without undoing the successful agent selection, while a failed agent switch still prevents a model attempt.


backnotprop:
Merged, thanks @rcdailey, that was quick. Older hosts keep switching agents and the model switch stays best-effort. Ships in the next release with #1569.

PR #1581, opened 2026-09-20
feat(pi): TUI plan-review renderer via Herdr (renderer config + PLANNOTATOR_RENDERER)

### LINKED ISSUE #1580: feat(pi): let plannotator_submit_plan review plans in plannotator-tui (terminal) instead of the browser
### What

The Pi extension's `plannotator_submit_plan` always starts the browser review server. For people working inside [Herdr](https://github.com/herdrdev/herdr) — especially on Windows — reviewing the plan in the existing **plannotator-tui** terminal UI would fit the workflow better. Plannotator already ships that TUI (and Herdr Annotate uses it for document review), but the Pi extension can't use it for plan review: the submit tool is hardwired to the web server, with no renderer option.

### Proposal

Make the plan-review renderer selectable in the Pi extension's existing config layers:

```jsonc
// .pi/plannotator.json (project) or ~/.pi/agent/plannotator.json (global)
{ "renderer": "tui" }   // "browser" (default) | "tui" | null (clear inherited)
```

with `PLANNOTATOR_RENDERER` as a per-session env override. Precedence matches `executionMode`: env > project > global > browser; unknown values warn and fall through to the inherited value.

### How it would work

When `tui` is active, `plannotator_submit_plan` spawns `plannotator-tui <plan.md>` in a Herdr pane (split beside Pi's own pane) with `PLANNOTATOR_DATA_DIR` pointed at a throwaway session dir, waits for exit, and maps the outcome back onto the exact same decision contract the browser flow returns (`{ approved, feedback }`):

- quit with no annotations → **approved**
- annotations present → **denied**, feedback = the TUI's own numbered export (send/archive text if the reviewer pressed `E`, rebuilt from `annotations.json` otherwise)
- decisions are appended to the shared feedback archive, so browser and TUI history stay unified

Since the TUI has no approve/deny gate, the approve mapping is implicit (`q` with no notes = approve). The alternative — adding a gate mode to plannotator-tui itself — is a much bigger change (second repo, Rust); happy to discuss.

### Platform notes

Works on Windows today (pi 0.79.x, Herdr 0.9.1, plannotator-tui 0.8.0). Exit detection uses a shell exit-status sentinel, since `herdr pane process-info` doesn't list the TUI process on Windows. This path drives the `herdr` CLI directly, so it isn't affected by the mac/linux gating of the Herdr Annotate doc-review pane.

I have this implemented and tested end-to-end on Windows (annotate → deny with feedback → revise → approve → execute; archive records written). Happy to open a PR if you're open to the feature.

AndrewJacop:
Closes #1580

### What

Adds an alternative plan-review renderer to the Pi extension: `plannotator_submit_plan` can hand the plan to **plannotator-tui** in a Herdr pane instead of opening the browser annotation UI. The browser flow is untouched and remains the default; the TUI path returns the same `PlanReviewDecision` the browser flow does, so approvals, denials-with-feedback, checklist parsing, and phase transitions all work unchanged.

### Config (layered like `executionMode`: env > project > global > default)

| Layer | Value |
|---|---|
| `PLANNOTATOR_RENDERER` env | `tui` \| `browser` — per-session override |
| `.pi/plannotator.json` (project, trust-gated) | `"renderer": "tui"` |
| `~/.pi/agent/plannotator.json` (global) | `"renderer": "tui"` |

`null` clears an inherited value; unknown values warn and fall through to the inherited value.

### Mechanics

- Binary resolved from `PLANNOTATOR_TUI_BIN` or `PATH`; clear startup error when missing
- Pane opened via `herdr pane split` beside the current pane, `PLANNOTATOR_DATA_DIR` pointed at a throwaway dir so annotation readback is deterministic and the user's real annotation history is untouched
- Exit detected via a shell exit-status sentinel (`…DONE_<rand>_True/0`): on Windows `herdr pane process-info` doesn't list the TUI process, and a reviewer who quits fast can beat alt-screen polling — a sentinel without success status means the TUI failed to run, which errors instead of ever mapping a broken launch onto an approval
- Outcome: no annotations → approved; annotations → denied + feedback (prefers the TUI's send-archive text, falls back to rebuilding the numbered format from `annotations.json`); decisions appended to the shared feedback archive with `surface: "plan"`

### Docs

README config reference updated (`renderer` row + top-level shape).

### Testing

- New unit tests: renderer config precedence (project over global, `null` clears, unknown warns + falls through), env-override chain, feedback formatter (line derivation, kinds, byte-parity with the TUI's export)
- `bun test` (config, plannotator-tui, browser, phase-tools-runtime, startup): 46 pass
- `tsc --noEmit` clean for the extension
- Manual E2E on Windows 11 / herdr 0.9.1 / plannotator-tui 0.8.0: annotate → quit → agent received denial with the annotation text → revised → resubmit → fast quit → approved → plan executed; feedback archive shows the `denied`/`approved` plan records


backnotprop:
Thanks for this, and for the Windows testing.

I think what you want already exists as [plannotator/herdr-annotate](https://github.com/plannotator/herdr-annotate) plus its [`plannotator-tui` skill](https://github.com/plannotator/herdr-annotate/blob/main/skills/plannotator-tui/SKILL.md): the agent opens the plan in a Herdr pane with `plannotator-tui herdr open plan.md`, you review it, and your comments come back as its next message. There are two short demos in that README.

It's macOS/Linux only today, which I'm guessing is why you went around it. Windows support is in flight in [herdr-annotate#49](https://github.com/plannotator/herdr-annotate/pull/49) and we'll ship the Windows fixes today. Your Windows findings would be really useful there.

Because of that I'd rather not add Herdr-specific code to the Pi extension, so I'm leaning toward not merging this one, unless you think I'm missing something. If the piece you care about is the tie-in to Pi's plan mode (a clean review approving and executing the plan), tell me, because that's the part the skill doesn't give you.

backnotprop:
Windows support shipped: plannotator/herdr-annotate#49 is on main.

```
herdr plugin install plannotator/herdr-annotate/windows-full
```

Needs Herdr 0.9.0+. Bind `prefix+o` to `annotate.open` per the README. It's labeled preview (verified on Windows 11 x86_64; ARM64 untested), so results on plannotator/herdr-annotate#43 would help, including `herdr plugin list` and any `pane.exit` lines from herdr-server.log if something fails.

If you still want the Pi adapter: open the pane with `herdr plugin pane open --plugin annotate --entrypoint doc --env PLANNOTATOR_TUI_FILE=<abs path> --env PLANNOTATOR_TUI_DELIVER_TO=<pane>` instead of typing into the shell, and read feedback with `plannotator-tui --export <file>` after the pane exits. And the `pane process-info` gap you found has no Herdr issue yet; a repro from you on herdrdev/herdr would be the most useful thing.

AndrewJacop:
Thanks for the quick response.
I noticed that windows was in flight but I managed to run it using the skill normally
the issue was that the plugin forced the local webserver, and i had to stop it and direct tit to the skill usage instead
It's really great that windows support now shipped. I'm going to test your apporach by the end of the day. and will try to reproduce the error after the update and get back to you with what i find

AndrewJacop:
Tested windows-full like I said I would, and it works. Opened a plan through the plugin pane API, annotated a couple of lines, hit E, and the feedback showed up as my next message in the pi session. 

The problem I keep hitting is that this flow and the pi plugin don't coexist. With the plugin installed, plan review always opens the web server: plannotator_submit_plan calls openPlanReviewBrowser with no  setting or env var to change it. So the realistic options are:
 - Uninstall the pi plugin and rely on the skill plus the CLI. That works fine, but plan mode is gone. Nothing maps a finished review to "approved, go execute", the feedback is just another message.
 - Keep the pi plugin and live with the browser. 
 - Run plannotator-tui vanilla outside pi and wire things up myself, updating it manually when releases land.

None of these are awful but each one gives up something I actually want. It basically requires some tinkering around if I'm running background/terminla-only-headless servers.
 
I completly agree on seprating the herdr code out of the plugin and makes total sense. I tihnk the thing that would fix it for me is a generic hook in the pi extension: a config option like `reviewCommand` that receives the plan file and whose result gets read back as feedback. No Herdr code in your extension, no web server requirement either. herdr-annotate could satisfy it, standalone plannotator-tui could, or default back to the local web server. anything that follows the convention could.

backnotprop:
Great, thanks for confirming Windows Full works end to end. A generic `reviewCommand` hook is the shape I'd take too: the Pi extension hands off the plan file, reads the result back, no Herdr or web-server knowledge inside it. The one thing it needs that doesn't exist yet is a decision contract, since plannotator-tui today has no approve/deny signal (any quit exits 0). If the command's result carried approved / denied-with-feedback / dismissed, plan mode's execute step would have something real to key on. It's the owner's call; I'll put it in front of him with your write-up. Leaving this open until then.

